PR #8, opened 2023-09-26
Get auth0 working in production runtime

rudokemper:
Submitting a PR for this work instead of committing directly for purposes of record-keeping:

Turns out it's not possible to use env vars for `auth` in production runtime environments, as discussed in this thread: https://github.com/nuxt-community/auth-module/discussions/1597

Hence, I implemented a version of the workaround suggested in there by user mirobertod on May 17, 2021.

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -1,37 +1,8 @@
-require('dotenv').config();
-
 import { NuxtConfig } from '@nuxt/types'
 
-const getAuthConfig = () => {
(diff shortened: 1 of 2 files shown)


IamJeffG:
Thanks.  And don't underestimate the value of in-line code comments as a form of record-keeping.  At least, my philosophy is that the closer documentation lives to the source code, the more likely someone is going to find it.

PR #10, opened 2023-10-02
Update Vue pages to better handle retrieval & messaging

rudokemper:
This PR implements several updates to the Vue pages:

* Make API requests when the page is mounted, not created (resolving a temporary server-side error being thrown).
* Decode and show auth0 error description on the login page. This will show messages like the following successful auth0 authentication attempt for a user that has not yet been approved by an auth0 admin to view the app:

image

### DIFF
--- a/pages/login.vue
+++ b/pages/login.vue
@@ -1,7 +1,8 @@
 <template>
     <!-- <Login /> -->
-    <div class="flex items-center justify-center h-screen">
-      <button class="px-4 py-2 bg-blue-500 text-white rounded hover:bg-blue-700 focus:outline-none focus:ring-2 focus:ring-blue-600 focus:ring-opacity-50" @click="loginWithAuth0">Login with Auth0</button>
+    <div class="flex flex-col items-center justify-center h-screen">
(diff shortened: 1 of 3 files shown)


pages/login.vue
```diff
@@ -1,17 +1,32 @@
...
+    data() {
+      return {
+        errorMessage: ""
+      };
+    },
     components: {
       // Login
     },
+    mounted() {
+      const hashParams = new URLSearchParams(window.location.hash.substring(1));
```
IamJeffG, line 22: I don't understand what sets these URL query params in the first place.  I'm just curious because it seems weird to store so much in the URL itself, but nothing fundamentally wrong with it either.

Perhaps not relevant here, but I have become a convert to [this pattern](https://www.robinwieruch.de/react-hooks-fetch-data/) for API calls ... not specifically the fact that it's a React hook, but specifically I mean storing 3 pieces of data in component state for each API call: isLoading, error, and (expected) data.  I also know people who swear by https://tanstack.com/query/v4/docs/vue/overview for querying APIs; I have not explored this one enough.

Just some downtime reading to do what you wish with.

PR #12, opened 2023-11-01
Implement and style vue-select component

rudokemper:
This PR adds and customizes a `vue-select` Component for dropdown filtering. `vue-select` is a great, feature rich select/dropdown/typeahead component Vue component. I am using it here to allow the user to do filtering using multiple selections, and to add a colored dot next to each selection that matches what is shown for the dots on the map. I also cleaned up some parts of both the `Map` and `Gallery` components.

image

image


IamJeffG:
the new component looks great!

PR #14, opened 2023-12-19
Support views for multiple tables

rudokemper:
Closes https://github.com/ConservationMetrics/guardianconnector-views/issues/13.

This is a much longer PR than I typically tend to strive for, a byproduct of having worked on this before and after going on leave for a month, and on one occasion, in an offline setting making a large amount of progress at once.

Writ large, I am changing the structure of the app to work with a multi-line config variable `NUXT_ENV_VIEWS_CONFIG` where multiple tables can be specified, and specific properties (like Mapbox or Filtering) applied to each one individually, rather than across the entire application for all views.

Routes for maps, galleries, and other views are now as follows: `/{view_type}/{table_name}`. 
For example, `/maps/mapeo_data` or `/gallery/kobo_submissions`.

There is also an index `/` route now which lists all of the available views per each table.

A few minor documentation changes unrelated to the scope of the PR have snuck their way in here, but I expect that they are not disruptive.

Something unexpected that I ran into is that Docker can't load multi-line config from a `.env` file. As a workaround, I wrote a small bash script that converts the user's `.env` file into a Docker-compatible `.env.docker` file with `NUXT_ENV_VIEWS_CONFIG` condensed into one line. (This same line can be used to feed application settings in hosting environments like Azure.)

As an upshot of this work, users can now leverage views for multiple tables in their data warehouse, and we can add a new dashboard view for change detection alerts.


.env.example
```diff
@@ -10,50 +10,47 @@ DB_SSL="true"
...
-
-# Unwanted SQL colums to hide in views
-# Refer to /docs/schemas.md for reference for what columns to hide
-UNWANTED_COLUMNS="latitude,longitude,coordinates"
-UNWANTED_SUBSTRINGS="meta,version,formhub,xform,attachments,tags,notes,topic,sender,geolocation,submission_time,submitted_by,status,supplementalDetails,validation_status,start,end,location"
\ No newline at end of file
+
+# Multi-line config providing views configuration for each database table
+# See docs/config.md for a full explanation for these settings.
+NUXT_ENV_VIEWS_CONFIG = "{
```
IamJeffG, line 38: Is the stuff here inside NUXT_ENV_VIEWS_CONFIG likely to change from deployment to deployment?  If not (if, say, we believe one certain group of settings is optimal across the board) then an alternative is to bake this into the Docker image itself -- through a Javascript or JSON or even .env file (if Javascript's own parser can deal with multi-line config).

This would be in the vein of superset's `[superset_config.py](https://github.com/ConservationMetrics/superset-deployment/blob/main/docker/pythonpath/superset_config.py)` file. There we went a step further which is to read the few things that do need per-deployment overrides from their own environment variables.  This, too, gets clunky.

Or, your `dockerenv.sh` script is good too :+1:

IamJeffG (commented):
I'm fine if you want to merge without review.  Or if you want me to look at something, is there a certain part I should focus on?

rudokemper:
No need - on this repo, I'm not expecting any code review, but am using PRs to provide an overview on what I'm adding/changing, and if I do have any questions I will ask those specifically and highlight any relevant code. Thanks!

.env.example
rudokemper: Unfortunately a good number of the variables inside NUXT_ENV_VIEWS_CONFIG (for example, the mapbox parameters) do change quite a bit from deployment to deployment. 

I did consider a run script that does the same as `dockerenv.sh` but feeds the output to Docker directly instead of baking a separate `.env.docker` file, but I thought there was value in having persisting access to the output too (e.g. for easy config of Azure or other cloud deployments).

PR #45, opened 2024-03-28
Add alerts to map legend, and improve graphic for geometry type

rudokemper:
Closes https://github.com/ConservationMetrics/guardianconnector-views/issues/43.

## Goal

To add alerts from the Alerts Dashboard component to the Map Legend, and to improve the legend graphic for polygons and point geometries.

## Screenshots

Linestring alerts - before & after
image image

Polygon alerts - before & after
image image

## What I changed

* Renamed `otherAlerts` to `previousAlerts` in a bunch of places for purposes of clear and consistent naming. Also renamed `recentAlerts` to `mostRecentAlerts`.
* On the `AlertsDashboard` component, add the alerts map legend layer IDs to be included in the string of layer IDs to be processed by `prepareMapLegendLayers`.
* Modify the CSS props for the polygon and circle geometries to better resemble how they might look on the map.

### DIFF
--- a/api/dataProcessing/transformData.ts
+++ b/api/dataProcessing/transformData.ts
@@ -150,7 +150,7 @@ const prepareAlertData = (
   embedMedia: boolean,
 ): {
   mostRecentAlerts: Array<Record<string, any>>;
-  otherAlerts: Array<Record<string, any>>;
+  previousAlerts: Array<Record<string, any>>;
 } => {
   const transformChangeDetectionItem = (
     item: Record<string, any>,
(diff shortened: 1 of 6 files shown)


rudokemper:
@IamJeffG any opinions on the legend styling here? (Note that it's quite tricky to _really_ do justice to how a polygon might look on a map, considering that the shape of a polygon could be just about anything!)

IamJeffG:
Looks great! I absolutely love the polygon blob icons!

My one suggestion would to be use shades in the same color family for related information.  For example the 3 polygons for mining (2019-2022; previous alerts; latest alerts) are each chosen from:

image

And if it ends up looking ugly on the map, then feel free to disregard.

IamJeffG (approved):

PR #54, opened 2024-04-16
Add date picker for Planet Monthly Visual Basemaps

### LINKED ISSUE #44: Add a Planet Monthly Visual basemap slider to the alerts dashboard
## Feature Request

Similar to work done in MapPacker in https://github.com/ConservationMetrics/mapPacker/pull/4 (and many other platforms like explorer.land and Global Forest Watch), let's add an option to add a Planet Monthly Visual background map overlay to the Alerts Dashboard, and a slider to select the month and year.

rudokemper:
## Goal

This PR implements a date picker for Planet Monthly Visual Basemaps. Closes #44. 

## Screenshots

image

## What I changed

* Installed, implemented, and configured a `vue2-datepicker` component (in lieu of HTML5 `month` field not being supported by all browsers, notably Firefox).
* `BasemapSelector` component now emits a `monthYear` value upon change in the datepicker component.
* Updated `changeMapStyle` function to use the `monthYear` value from the date picker instead of a hard-coded value in the style.
* Max date value is set depending on if today's date >= 15. Planet publishes new visual basemaps on the 15th of each month. So if today's date >= 15, then `maxMonth` is the previous month; otherwise it is set to be two months prior.
* Min date value is the earliest date Planet data became available.
* Made a note that the Mapbox Satellite style is only up to a certain date.

## What I'm not doing here

Visualizing the month selection via a slider. I want to make that a separate batch of work to be implemented in the future.

### DIFF
--- a/components/BasemapSelector.vue
+++ b/components/BasemapSelector.vue
@@ -54,14 +54,20 @@
     },
     computed: {
         maxMonth() { 
-            // Calculate one month ago
+            // If the current day is less than or equal to 15, maxMonth is two months ago.
+            // Otherwise, maxMonth is the previous  month.
+            // This is because Planet NICFI monthly basemaps for the previous month are published on the 15th of each month.
             const date = new Date();
-            date.setMonth(date.getMonth() - 1);
+            if (date.getDate() <= 15) {
+                date.setMonth(date.getMonth() - 2);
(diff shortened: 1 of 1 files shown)


IamJeffG:
Looks slick!

Issue #53, opened 2024-04-11
Implement GLAD and Integrated Deforestation Alerts as a possible change detection source using the GFW API


rudokemper:
> ❗️ Updated!

## Feature Request

Let's add GLAD and Integrated Deforestation Alerts alerts as a new change detection data source using the [GFW API](https://www.globalforestwatch.org/help/developers/guides/).

## Implementation Plan
1. Create a new GCV GET endpoint that uses an auth token to request data from GLAD and/or Integrated Deforestation Alerts from the GFW API at this endpoint: `https://data-api.globalforestwatch.org/dataset/gfw_integrated_alerts/latest/query`. We should write a SQL query that selects latitude, longitude, date, and confidence interval.
2. If needed, the GCV API might require the implementation of retry logic to refresh the bearer token once the auth token expires.
3. The `AlertsDashboard` component should be modified to be able to utilize the data returned by this endpoint.

rudokemper:
It appears that I have missed this [Developer documents](https://www.globalforestwatch.org/help/developers/guides/) section of the GFW Help Center, which provides guidance on how to use and set up auth tokens + API key, and construct POST requests for querying data.

The `https://data-api.globalforestwatch.org/datasets` endpoint reveals all of the datasets available in the GFW API.

Let's take the [GFW Integrated Alerts](https://www.globalforestwatch.org/blog/data/integrated-deforestation-alerts/), which is a new "integrated deforestation alerts layer [that] combines the analytical power of GLAD, GLAD-S2 and RADD deforestation alerts to provide a faster, more confident view of forest disturbances than any one individual system." 

At the GFW Integrated Alerts dataset endpoint at `https://data-api.globalforestwatch.org/dataset/gfw_integrated_alerts/`, there is a `versions` array with daily updates. To access data for that specific version: `https://data-api.globalforestwatch.org/dataset/gfw_integrated_alerts/{version}`. But better yet - `latest` resolves to the latest version available.

We can [construct a query](https://www.globalforestwatch.org/help/developers/guides/query-data-for-a-custom-geometry/) with SQL parameters for this POST endpoint: `https://data-api.globalforestwatch.org/dataset/gfw_integrated_alerts/latest/query`.

Example of a POST request structure provided in the documentation:
[...]

PR #58, opened 2024-04-25
[WIP] Basic (and currently inefficient) implementation of GFW API data on the Alerts Dashboard

rudokemper:
## Goal

Work towards #53: To optionally fetch Integrated Deforestation Alerts from the GFW API and add those to the Alerts Dashboard.

## Discussion

The objective of this PR is **not** to submit code for review and merging. Instead, it should serve as a vehicle for discussion toward a better approach, as the method used here is definitely inefficient.

The GFW API yields vector data by doing raster analysis on the fly, resulting in a high amount of latency, and there is a payload ceiling of 5,291,556 bytes. Using the method implemented here, even an API query for a relatively small AOI around the village of Kawemhakan in Suriname took 875.657ms (for 668 points). For the entire Wayana territory, we reached the payload ceiling.

It seems to me that there are two options to overcome this:

(1) I think this approach to fetch data from the GFW API _could_ work if we are actually storing the response in a database or as a GeoJSON file, which we update regularly with new GFW data. We could easily write a frizzle pull component to do this. This would be the way to go if we want to serve GFW data  _**as alerts**_: that is, as active features on the Alerts Dashboard that you can click on to view properties, and export as CSV/GeoJSON. Not to mention that we will eventually make it possible for users to add data to alerts, and relate together alerts that belong to the same incident.

(2) If we just want to visualize GFW data on an alerts dashboard as a contextual map layer and nothing more, then we should see if GFW makes a public tileset of their data available somewhere. We wouldn't need to worry about limiting it to an AOI since tiled data is only loaded for your viewport anyway.

I think that if we want to provide GFW data as an open alternative to proprietary change detection data, then we should consider (1). 

## Screenshots

image

## What I changed

* For the `/alerts` API, optionally fetch `gfw_integrated_alerts` data from the GFW API, and return that as a GeoJSON object, to be added as a source and layer on the `AlertsDashboard` map canvas.

### DIFF
--- a/api/config.ts
+++ b/api/config.ts
@@ -22,6 +22,8 @@ interface ViewConfig {
   EMBED_MEDIA: string;
   FRONT_END_FILTERING: string;
   FRONT_END_FILTER_FIELD: string;
+  GFW_API_KEY: string;
+  GFW_AOI: string;
   MAPBOX_STYLE: string;
   MAPBOX_PROJECTION: string;
   MAPBOX_CENTER_LATITUDE: string;
--- a/api/index.ts
+++ b/api/index.ts
@@ -221,6 +221,7 @@ if (!VIEWS_CONFIG) {
 
             // Optionally add GFW data
             if (VIEWS[table].GFW_API_KEY && VIEWS[table].GFW_AOI) {
+              console.log("Fetching GFW data...")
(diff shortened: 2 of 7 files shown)


rudokemper:
We will not pursue fetching GFW data on the front end at this time, and there is a better way to work with the GFW API than the approach here. So closing.

PR #66, opened 2024-08-12
Store and fetch views config in the database instead of `NUXT_ENV_VIEWS_CONFIG` env var

rudokemper:
_Note: This is an unusually large PR because I worked on this code mostly while offline on a plane ‍✈️_ 

## Goal

Closes https://github.com/ConservationMetrics/guardianconnector-views/issues/65. The goal of this PR is to store and retrieve the configuration from a database table instead of using the multi-line `NUXT_ENV_VIEWS_CONFIG` env var. Additionally, it introduces a UI route where the configuration can be updated. The application now re-initializes the view routes whenever a configuration change is made, eliminating the need to restart the app to apply the changes.

⚠️ This PR introduces a breaking change that will affect all production deployments, which are set to use `NUXT_ENV_VIEWS_CONFIG`. I will manually add the necessary configuration to the databases for each deployment to ensure there are no disruptions on the front end during redeployment.

## Screenshots

UI for updating config:
image

## What I changed

* Removed all traces of `NUXT_ENV_VIEWS_CONFIG` from server-side and config code.
* Added a function `createDatabaseIfNotExists` to... create a database if it does not exist (for `gc_views`).
* Updated the database connection setup in the app to include permissions for writing and creating, in addition to reading.
* Added a function `fetchConfig` to retrieve the configuration from the database and return a typed `viewsConfig` object.
* Added a function `updateConfig` to write a `viewsConfig` string (stringified from JSON) to the database.
* Modified the app initialization to first set up the database connections (and create the `gc_views` database if it doesn't exist) before setting up the views according to the configuration. A new `/api/config/${tableName}` POST endpoint now re-initializes the views when configuration changes are made.
* Added a new route `config.vue` which fetches the config from a new `/api/config` API endpoint and renders each table as a form, with the config key/value pairs as fields. A `submitConfig` function sends the changes to a new `/api/config/${tableName}` POST endpoint.  An accompanying css file `overlay.css` was added to display a "config successfully updated" modal.
* Updated `index.vue` to fetch the config using the `/api/config` API endpoint instead of passing it directly via Nuxt `this.$config`.
* Bumped version to 2.0.0 in `package.json` (since this work introduces a breaking change). Also fixed some linting and husky config issues.
* Adapted documentation, `.env.example`, and Docker workflows.
* Deleted the `dockerenv.sh` script since we no longer need to account for multi-line env config.

## What I'm not doing here

This PR adds a skeletal UI for updating the `config` object for each `table_name`. Many improvements need to be made such as implementing field types, validations, the option to add a new view, the option to delete a view, etc.


api/database/dbConnection.ts
```diff
@@ -64,4 +64,50 @@ const setupDatabaseConnection = (
...
+    user: user,
+    host: host,
+    password: password,
+    port: parseInt(port, 10),
+    ssl: ssl === "true" ? { rejectUnauthorized: false } : false,
+    database: defaultDb,
+  });
+
+  try {
+    await client.connect();
```
IamJeffG, line 90: When does this run?  Looks like it runs on module import time, which is not necessarily the same as server startup.

I am not very familiar at this level of Javascript, but typically you want code import to go really fast.  Also, I imagine you couldn't import this file into a test without the import failing?

I would expect this to be called from either
- some sort of explicit  HTTP server-startup function, or
- a separate CLI command that gets called BEFORE the server-startup CLI command.  Like you might call the former only when there are database migrations (you  might call it "migrate")

api/database/dbOperations.ts
```diff
@@ -96,4 +96,117 @@ const fetchData = async (
...
+}
+
+export const fetchConfig = async (
+  db: any,
+  isSQLite: boolean | undefined,
+): Promise<Views> => {
+  // Create the config table if it does not exist
+  const createConfigTable = `CREATE TABLE IF NOT EXISTS config (
+         table_name TEXT PRIMARY KEY,
+         config TEXT
```
IamJeffG, line 137: These names feel extremely elusive, like there might be many things in an app called "config", so which does this store?  Also having `table_name TEXT` as a primary key, I am really struggling to understand what is stored in this table.

I wonder what this table and columns would be named if you went one  notch "more specific"

IamJeffG (approved):

api/database/dbOperations.ts
rudokemper: Thanks for this! Yes, I will improve the naming here before merging.

api/database/dbConnection.ts
IamJeffG: My mistake! I missed the createDatabaseIfNotExists definition and 2 space of indent and was thinking this was at the top-level.

PR #83, opened 2024-11-22
Implement custom multi tag for Vue 3 Select component

rudokemper:
## Goal

Now that [`vue3-select-component` allows a custom option slot for #value when `isMulti` is set](https://github.com/TotomInc/vue3-select-component/issues/128), this PR re-adds the selected colored dots shown on the map into the listbox for our `DataFilter` component, as we had them in the Vue 2 select component.

## Screenshots

image

## What I changed

* Bumped `vue3-select-component` to 0.7.0.
* Adapted HTML and CSS for the listbox template to show colored dots and style the option boxes similar to what we had before.

### DIFF
--- a/components/shared/DataFilter.vue
+++ b/components/shared/DataFilter.vue
@@ -62,19 +62,23 @@ function emitFilter() {
       :key="uniqueValues"
     >
       <!-- This is what shows in the listbox when selected -->
-      <!-- Pending support for tags https://github.com/TotomInc/vue3-select-component/pull/129 -->
-      <template #tag="{ option }">
-        <span
-          class="colored-dot"
-          v-if="showColoredDot"
-          :style="{ backgroundColor: option.color }"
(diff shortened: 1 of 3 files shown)


IamJeffG (approved):
Code looks good.
One initial thought about the UI: As long as the base layer includes POIs as little circles, it's hard for my eye to quickly filter those POIs away from the same-sized-circles that are my own data.  Would suggest either using a base layer without circle POIs, or using a different shape (square?) for my own data.

rudokemper:
Yeah, I understand. It's always a bit tricky coming up with feature styles that will work with a multiplicity of basemaps. We typically use `satellite-streets` where we don't have the same issue, but for this (COP16 CoMapeo) project it made sense to use `streets` as it was taking place in a city. I'll make an issue to improve this in the future for the most common map styles.

Issue #91, opened 2025-02-06
[Alerts] Improve how we visualize existing alerts on the map


rudokemper:
## Feature Request

One of our users is receiving change detection alerts for a territory of >11m hectares. When zoomed out to the full extent of the territory, it is difficult to perceive individual change detection features for very small areas - in particular, for `previousAlerts` that don't have a pulsing circle.

Let's find a way to improve how we visualize alerts in all three formats (point, line, polygon) so that they are clearly visible at different zoom levels. This could be done by modifying the map style directly, or by introducing a CSS object on top of the map canvas (similar to what we do with `addPulsingCircles()` for `mostRecentAlerts` in the `MapDashboard` component).

rudokemper:
Exploring the alert views in production (which contain either polygon or line alerts), I’ve observed that visibility becomes an issue for all alerts at approximately zoom level 8.

I think the best solution is to calculate a centroid for each polygon and line layer and use those to create a point layer that is only visible at lower zoom levels (e.g. `visibility: "visible"` when zoom < 8 and `"none"` when zoom >= 8). This approach maintains click event functionality for the alerts while preventing map clutter and avoiding the need for an additional CSS object (and an explanatory burden to distinguish between the pulsing circles used for the most recent alerts and any new visual elements).

Another option could be adding a UI cue like as a "Zoom in to see details" message, but that alone won't suffice. The Territory view in Mapeo Desktop does this, and it's a known issue that it doesn’t help users locate their points at lower zoom levels, effectively turning it into a "needle in a haystack" search until zooming in past level 8.

IamJeffG:
Love the idea of turning polygons into Point Markers at lower zooms.  Wonder if [MarkerCluster](https://leafletjs.com/2012/08/20/guest-post-markerclusterer-0-1-released.html) is also helpful, or not needed?

rudokemper:
Good call, we've done that for Terrastories for example:

![Image](https://github.com/user-attachments/assets/7984f700-b662-4045-8120-c31684004923)

Actually, another thought there is to use [custom markers](https://docs.mapbox.com/help/tutorials/custom-markers-gl-js/) instead of Points, which is what I was thinking yeserday. We could design markers for both individual alerts and clustered ones.

PR #94, opened 2025-02-27
Improve alerts visualization at different zoom levels

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-explorer/issues/91. The goal is to switch between red/orange warning icons and spatial data at a zoom threshold of level 10.

Clicking on one of the icons does not open the sidebar with the alert feature data; rather, it zooms in past the `z10` threshold and pans the map to the center of the feature. From there, the user can click again on the alert Polygon / LineString or other ones in the area. 

This visualization style was vetted by two of our users, who both said they really liked it.

No API modifications were needed—we were already calculating geographic centroids for alert features to display in the sidebar.

## Screenshots

image

[Kooha-2025-02-27-15-11-43.webm](https://github.com/user-attachments/assets/feb7c14f-539a-44ad-a076-81a158232dbf)
_(My screen grab didn’t capture my mouse, but around 2 seconds in, I demonstrate clicking a warning icon, which zooms in and pans to the centroid.)_

## What I changed

* Added red and orange triangle images to `assets/icons/`.
* Made `addAlertsData()`  an async function to load things in sequence and avoid race condition issues. Within the function:
  * Created `loadMapImage()` to add images to the map style.
  * Added  `addAlertPointsLayer()` which:
    *  calls `loadMapImage()` with icon name and URL params
    * adds the sources and layers for the red / orange triangle features. 
  * Refactored alert layer handling (e.g. for polygons and linestrings) by introducing `addAlertLayer()`
  * Now, all alerts layers (polygons, linestrings, icon points) are added sequentially in a `Promise.all` method.
  * Modified the on-click event listener: clicking on `-points` layer features pans to the centroid and zooms to level 13 (as shown in the video).
* Made minor tweaks to redundant JSDoc comments in TypeScript.
* Added some copy to the `AlertsIntroPanel` sidebar about the triangles.

## What I'm not doing here

* **Messing with the legend** - I considered making the new icons toggleable, but I’m not sure yet if that’s desirable. Implementing it would require refactoring how alerts are handled in the legend, so we’ll revisit this later.
* **Making the zoom threshold configurable** – It’s currently hardcoded at `z10`, which seems appropriate for alert geography across all instances. (The demo "fake alerts" in the video aren’t a great representation since those polygons are much larger than real alerts.) We can add configurability later if needed.
* **Clustering icons** - We discussed clustering in https://github.com/ConservationMetrics/gc-explorer/issues/91. But after this work, I'm not sure if it will be helpful. I think it would be interesting for our users to see as much detail as possible about the exact locations of alerts at lower zoom levels, even if they are overlapping. And it doesn't really look that bad. Again, we can add this in follow-on work if needed.


components/AlertsDashboard.vue
```diff
@@ -131,224 +131,230 @@ const emit = defineEmits(["reset-legend-visibility"]);
...
-    )
-  ) {
-    // Add the previous alerts source to the map as LineStrings
-    if (!map.value.getSource("previous-alerts-linestring")) {
-      map.value.addSource("previous-alerts-linestring", {
+  /**
+   * Adds a GeoJSON point layer to the map using the geographicCentroid property,
+   * with specified icon.
+   */
+  const addAlertPointsLayer = async (
```
IamJeffG, line 258: I'm missing (due to my inexperience with vue) what exactly invokes `addAlertPointsLayer`, but ultimately what I'm wondering is:
Is there any chance of a race condition where two or more invocations to `addAlertsPointsLayer` are running at the same time?  For example it could be called before the previous call has finished. In this case some updates to ref state `map.value` could be lost.

Same question about `loadMapImage` making modifications to map.value (`.addImage()`) but I assume the answer is the same for both these.

rudokemper: `addAlertsData` (which defines and calls these functions you are asking about) is called by `prepareMapCanvasContent`, which itself is called on a few occasions:

1. On mount (in `onMounted`).
2. If the basemap is changed (in `handleBaseMapChange`), once the map state is idle.

It's possible that somebody would try to change the basemap before the alerts data loads, but doing so rebuilds the entire `map` object anyway.

IamJeffG (approved):

PR #106, opened 2025-06-17
Add URL query parameter support for alert selection

### LINKED ISSUE #97: [Alerts] Add ability to link to specific alert using `alertId` as query parameter.
## Feature Request

Currently, the alerts dashboard can only be accessed by the `/alerts/alerts` path. If users want to discuss a specific alert, they will need to guide others on which alert shape on the map to click.

It would be nice if we can link to a specific alert using a URL query parameter.

Idea suggested by @nicopace

conservationtimothy:
## Description
This PR implements the ability to link directly to specific alerts using the ⁠ alertId ⁠ query parameter, as requested in #97. This enhancement allows users to share direct links to specific alerts, improving collaboration and communication.

## Changes
•⁠  ⁠Added URL query parameter handling for alert selection
•⁠  ⁠Implemented automatic alert selection and sidebar opening when ⁠ alertId ⁠ is present in URL
•⁠  ⁠Added automatic map zooming to the selected alert's location
•⁠  ⁠Added URL cleanup when alert selection is reset

## Implementation Details
•⁠  ⁠Uses Vue Router's ⁠ useRoute ⁠ and ⁠ useRouter ⁠ for URL parameter management
•⁠  ⁠Preserves existing query parameters when updating the URL
•⁠  ⁠Handles different geometry types (Point, Polygon, LineString) for proper zooming
•⁠  ⁠Maintains existing functionality while adding the new feature

## Testing
1.⁠ ⁠Click on an alert on the map
   - [ ] URL should update with ⁠ ?alertId=<alert-id> ⁠
   - [ ] Sidebar should open with alert details
2.⁠ ⁠Share URL with ⁠ alertId ⁠ parameter
   - [ ] Map should load with the correct alert selected
   - [ ] Sidebar should open automatically
   - [ ] Map should zoom to the alert's location
3.⁠ ⁠Reset selection
   - [ ] URL should be cleaned of the ⁠ alertId ⁠ parameter
   - [ ] Map should return to initial state

## Screenshots
alerts

## Related Issues
Closes #97

## Checklist
•⁠  ⁠[x] Code follows the project's style guidelines
•⁠  ⁠[x] Documentation has been updated
•⁠  ⁠[x] Tests have been added/updated
•⁠  ⁠[x] All tests pass
•⁠  ⁠[x] No console errors or warnings


conservationtimothy:
This has a dependency on https://github.com/ConservationMetrics/gc-explorer/pull/107

components/AlertsDashboard.vue
```diff
@@ -119,6 +123,43 @@ onMounted(() => {
...
+          feature.geometry.type === "Polygon" ||
+          feature.geometry.type === "MultiPolygon"
+        ) {
+          const bounds = bbox(feature);
+          map.value.fitBounds(bounds, { padding: 50 });
+        } else if (feature.geometry.type === "LineString") {
+          const line = lineString(feature.geometry.coordinates);
+          const lineLength = length(line, { units: "kilometers" });
+          const midpoint = along(line, lineLength / 2, { units: "kilometers" });
+          const [lng, lat] = midpoint.geometry.coordinates;
```
rudokemper, line 158: I could have sworn we had a `calculateLinestringCentroid()` function (or something similar) somewhere!

But since we don't seem to, and this code is also repeated in `addPulsingMarker()`, we can make it one.

rudokemper (changes requested):
Excellent work! Love the elegant solution using `vue-router`. 

A few requested changes. Thanks for diving in - I realize this component is massive with a lot going on.

rudokemper (approved):
Thank you!

Now that we have this in place, it occurs to me that we should probably have a "Copy link to alert" button or icon somewhere, for users who might have a difficult time copy & pasting the URL from the browser (it's tricky on phones).

We didn't discuss or scope that, though, so I'd be fine merging this as-is and creating an issue for it! But since it's a relatively quick thing, I think it would also be fine to add it into this PR. 

Maybe we can leverage Lucide Icons for a nice "copy" icon and add the link to the bottom of the `Feature` component? :)

conservationtimothy:
> Copy link to alert" button or icon somewhere, for users who might have a difficult time copy & pasting the URL from the browser (it's tricky on phones).
> 
> We didn't discuss or scope that, though, so I'd be fine merging this as-is and creating an issue for it! But since it's a relatively quick thing, I think it would also be fine to add it into this PR.
> 
> Maybe we can leverage Lucide Icons for a nice "copy" icon and add the link to the bottom of the `DataFeature` component? :)

absolutely no big

glamboyosa:
Docker failure is expected, the package does not exist yet until we merge in the dependent branch.

components/shared/DataFeature.vue
```diff
@@ -96,6 +107,19 @@ const setMediaBasePath = () => {
...
+      <button
+        class="flex items-center gap-2 px-4 py-2 text-sm text-gray-600 hover:text-gray-900 transition-colors duration-200"
+        @click="copyLink"
+      >
+        <component
+          :is="showCopied ? Check : Copy"
+          class="w-4 h-4"
+          :class="{ 'text-green-500': showCopied }"
+        />
+        <span>{{ showCopied ? $t("copied") : $t("Copy link to alert") }}</span>
```
rudokemper, line 120: Last thing, I promise!

Whenever adding new translatable strings in a PR, I do try to add them right away to the available locales. I think it's fine to use either Google Translate or AI to do so for the languages we have currently.

conservationtimothy: Added this

components/AlertsDashboard.vue
```diff
@@ -119,6 +123,43 @@ onMounted(() => {
...
+      const allFeatures = [
+        ...props.alertsData.mostRecentAlerts.features,
+        ...props.alertsData.previousAlerts.features,
+      ];
+      const feature = allFeatures.find(
+        (f) => f.properties?.alertID === alertId,
+      );
+      if (feature) {
+        // Find the appropriate layer ID for this feature
+        const layerId = feature.properties?.isRecent
```
rudokemper, line 139: I missed this during my initial review: we don't set (or inherit) an `isRecent` property. So `layerId` here will always fall back to previous alerts. Hence, any most recent alerts are not selected. We will need a different solution.

conservationtimothy: Ok, editing.

i18n/locales/en.json
```diff
@@ -108,5 +108,7 @@
   "views": "Views",
   "viewOnGoogleMaps": "view on Google Maps",
   "yourAccessIsPending": "Your access is pending. Please contact a Guardian Connector administrator for account approval.",
-  "yourMapboxStyleDefault": "Mapbox - your style (default)"
+  "yourMapboxStyleDefault": "Mapbox - your style (default)",
+  "copyLink": "Copy link to alert",
+  "copied": "Copied!"
```
rudokemper, line 113: And just the nittiest of nits: these lists are alphabetized. Can we move the text up to where it belongs in that sort. Thanks!

IamJeffG:
> Btw https://github.com/ConservationMetrics/gc-explorer/pull/107 needs to be merged into this first.

My preference on workflow for this is to first merge 107 into `main`, then merge `main` into this current feature branch.

That way this PR's diff (against `main`) is not muddied by 107, but rather keeps a mostly faithful historical record of the lines of code that Rudo reviewed.

LMK if that didn't make sense.

conservationtimothy:
> > Btw #107 needs to be merged into this first.
> 
> My preference on workflow for this is to first merge 107 into `main`, then merge `main` into this current feature branch.
> 
> That way this PR's diff (against `main`) is not muddied by 107, but rather keeps a mostly faithful historical record of the lines of code that Rudo reviewed.
> 
> LMK if that didn't make sense.

Sure, I just do not know if that might cause a merge conflict?

i18n/locales/en.json
conservationtimothy: Resolved.

components/AlertsDashboard.vue
conservationtimothy: we now store isRecent in the URL query parameters and use that to determine most recent or previous🙂

conservationtimothy: Just pushed a change where we:
1. Take the alert ID from the URL
2. Check if that alert ID exists in mostRecentAlerts
3. If it does, we use the "most-recent-alerts" layer
4. If it doesn't, we know it must be in "previous-alerts"

This way, even if an alert moves from recent to previous between when the link was shared and when it was clicked, we'll still find it in the right place. The URL only needs to store the alert ID, and we figure out which layer it belongs to at runtime.

Wdyt?

rudokemper: Great solution! :+1:

rudokemper (approved):

rudokemper:
Please feel free to squash and merge.

PR #109, opened 2025-06-18
 Add URL query parameter support for Mapeo data selection

### LINKED ISSUE #108: Add `mapeoDocId` query param to link to Mapeo data
## Feature Request

With https://github.com/ConservationMetrics/gc-explorer/pull/106 all done and implemented, I think we can now do the same for Mapeo data showing on the same alerts dashboard. This should be a fairly light lift as it is applying the same logic to a different data type.

For context: `AlertsDashboard` shows both `alertsData` and `mapeoData` on the map. Alerts data are change detection alerts derived from satellite imagery, and Mapeo data are field data collected by community members using a smartphone application called Mapeo.

I think it is just as useful to be able to link directly to a Mapeo point as an alert. (And actually, something we missed in #106 is that the new commits will add "Copy link to alert" to Mapeo data in addition to alerts data, which just means the user copies the `/alerts/{table_name}` path and nothing more. So this is a minor bug that we missed in testing / code review. EDIT: they are shown on _any_ implementation of `DataFeature` component, not just alerts dashboard but also Map and Gallery. So we should do some work to limit that to only show on `AlertsDashboard`)

Let's implement the same query parameter approach to `mapeoData`.

We used to have Mapeo points showing on `demo` alerts dashboard. I think the data table might have been inadvertently removed from config during some experimentation. I just restored this data to the view:

Image

Acceptance criteria:

- [ ] A `?=mapeoDocId`={feature.properties.id}` query parameter shows when a Mapeo point is selected.
- [ ] The copy button string should be "Copy link to Mapeo observation"
- [ ] No DRY code - any repeat code applied to alerts and Mapeo data is generalized.
- [ ] "Copy link" is only shown on `AlertsDashboard`. We already pass some props to the sidebar to handle alerts-specific implementations, that we can probably reuse to handle this.

conservationtimothy:
## Description
This PR implements the ability to link directly to specific Mapeo observations using the `mapeoDocId` query parameter, similar to the alert selection feature implemented in #106. This enhancement allows users to share direct links to specific Mapeo observations, improving collaboration and communication between field data collectors and analysts.

## Changes
- Added URL query parameter handling for Mapeo data selection
- Implemented automatic Mapeo feature selection and sidebar opening when `mapeoDocId` is present in URL
- Added automatic map zooming to the selected Mapeo feature's location
- Added URL cleanup when feature selection is reset
- Limited "Copy link" functionality to only show on AlertsDashboard
- Added Mapeo-specific copy link text

## Implementation Details
- Uses Vue Router's `useRoute` and `useRouter` for URL parameter management
- Preserves existing query parameters when updating the URL
- Handles different geometry types (Point, Polygon, LineString) for proper zooming
- Reuses existing feature selection logic to avoid code duplication
- Adds new props to control UI behavior (`isMapeo`, `isAlertsDashboard`)
- Adds new i18n translations for Mapeo-specific text

## Testing
1. Click on a Mapeo point on the map
   - [x] URL should update with `?mapeoDocId=<feature-id>`
   - [x] Sidebar should open with Mapeo feature details
   - [x] Copy link button should show "Copy link to Mapeo observation"

2. Share URL with `mapeoDocId` parameter
   - [x] Map should load with the correct Mapeo feature selected
   - [x] Sidebar should open automatically
   - [x] Map should zoom to the Mapeo feature's location

3. Reset selection
   - [x] URL should be cleaned of the `mapeoDocId` parameter
   - [x] Map should return to initial state

4. Verify copy link button visibility
   - [x] Should only show on AlertsDashboard
   - [x] Should not show on Map or Gallery views

## Screenshots

Screenshot 2025-06-18 at 12 20 41
Screenshot 2025-06-18 at 12 20 00

## Related Issues
Closes #108

### DIFF
--- a/components/AlertsDashboard.vue
+++ b/components/AlertsDashboard.vue
@@ -129,6 +129,28 @@ const selectInitialMapeoFeature = (mapeoDocId: string) => {
       | "MultiLineString"
       | "MultiPolygon"
       | "GeometryCollection";
+
+    // Create a new GeoJSON Feature object instead of using mapeoFeature directly for several critical reasons:
+    //
+    // 1. DATA FORMAT MISMATCH: Mapeo data comes as raw DataEntry objects, but Mapbox expects proper
+    //    GeoJSON Feature objects with specific structure (type, geometry, properties, id).
+    //
+    // 2. ID NORMALIZATION REQUIREMENT: Mapeo document IDs are 64-bit hex strings (e.g., "0084cdc57c0b0280")
+    //    that exceed JavaScript's safe integer range (2^53 - 1). Mapbox requires feature IDs to be either
+    //    Numbers or strings that can be safely cast to Numbers. Without normalization, Mapbox falls back to
+    //    undefined IDs, causing setFeatureState() to fail with "The feature id parameter must be provided."
+    //    We use the normalized 53-bit safe integer (mapeoFeature.normalizedId) for Mapbox compatibility.
+    //
+    // 3. COORDINATE PARSING: The geometry coordinates are stored as JSON strings in the raw data but need
+    //    to be parsed into array format for GeoJSON compliance.
+    //
+    // 4. FEATURE STATE MANAGEMENT: Mapbox's setFeatureState() requires matching IDs between the GeoJSON
+    //    source and the feature selection. Without this transformation, feature highlighting and selection
+    //    would fail completely.
+    //
+    // For detailed technical explanation of this problem and solution, see:
+    // https://github.com/ConservationMetrics/gc-explorer/pull/109#issuecomment-2985123992
+    // Reference: https://stackoverflow.com/questions/72040370/why-are-my-dataset-features-ids-undefined-in-mapbox-gl-while-i-have-set-them
     const feature: Feature = {
       type: "Feature",
       id: mapeoFeature.normalizedId || mapeoFeature.Id, // Use normalized ID if available
--- a/server/dataProcessing/transformData.ts
+++ b/server/dataProcessing/transformData.ts
@@ -3,7 +3,6 @@ import {
   capitalizeFirstLetter,
   formatDate,
   getRandomColor,
-  normalizeMapeoId,
 } from "./helpers";
 
 import type {
@@ -181,16 +180,6 @@ const prepareMapData = (
           item.geotype = "Polygon";
         }
       }
-    }
-
-    // Add normalized ID for Mapeo features to ensure Mapbox compatibility
-    // Mapeo document IDs are 64-bit hex strings that need to be normalized to 53-bit integers
-    if (
-      item.ID &&
-      typeof item.ID === "string" &&
-      item.ID.match(/^[0-9a-fA-F]{16}$/)
-    ) {
-      item.normalizedId = normalizeMapeoId(item.ID);
     }
 
     const filterColumnValue =
@@ -630,15 +619,20 @@ const transformToGeojson = (data: DataEntry[]): FeatureCollection => {
 
     Object.entries(input).forEach(([key, value]) => {
       if (key === "alertID") {
-        feature.id = value.substring(4);
+        feature.id = String(value).substring(4);
         feature.properties![key] = value; // Use non-null assertion
       } else if (key.startsWith("g__")) {
         const geometryKey = key.substring(3); // Removes 'g__' prefix
         if (feature.geometry) {
           if (geometryKey === "coordinates") {
-            feature.geometry[geometryKey as keyof Geometry] = JSON.parse(value);
+            feature.geometry[geometryKey as keyof Geometry] = JSON.parse(
+              String(value),
+            );
           } else if (geometryKey === "type") {
-            feature.geometry.type = value as "Point" | "LineString" | "Polygon";
+            feature.geometry.type = String(value) as
+              | "Point"
+              | "LineString"
+              | "Polygon";
           }
         }
       } else {
--- a/server/api/[table]/alerts.ts
+++ b/server/api/[table]/alerts.ts
@@ -19,6 +19,38 @@ import type {
   AlertsMetadata,
 } from "@/types/types";
 
+/**
+ * Converts a Mapeo document ID (64-bit hex string) to a 53-bit safe integer
+ * for Mapbox feature state management. This is a lossy, non-reversible operation.
+ *
+ * Mapbox requires feature IDs to be either a Number or a string that can be safely
+ * cast to a Number, but Mapeo IDs are 64-bit hex strings that exceed JavaScript's
+ * safe integer range. This function reduces the 64-bit value to a 53-bit one using
+ * a bitwise AND operation with a 53-bit mask.
+ *
+ * Reference: https://stackoverflow.com/questions/72040370/why-are-my-dataset-features-ids-undefined-in-mapbox-gl-while-i-have-set-them
+ *
+ * @param {string} mapeoId - The Mapeo document ID as a 16-character hex string (e.g., "0084cdc57c0b0280")
+ * @returns {number} - A 53-bit safe integer for use with Mapbox feature state management
+ * @throws {Error} - If the input is not a valid 16-character hex string
+ */
+const generateMapboxIdFromMapeoFeatureId = (mapeoId: string): number => {
+  // Validate that this is actually a Mapeo ID format
+  if (
+    !mapeoId ||
+    typeof mapeoId !== "string" ||
+    !mapeoId.match(/^[0-9a-fA-F]{16}$/)
+  ) {
+    throw new Error(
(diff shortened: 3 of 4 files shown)


components/AlertsDashboard.vue
```diff
@@ -165,7 +169,61 @@ onMounted(() => {
           );
           map.value.flyTo({ center: [lng, lat], zoom: 13 });
         }
+        isMapeo.value = false;
       }
+    } else if (mapeoDocId && props.mapeoData) {
```
rudokemper, line 174: Can we turn these into separate functions to keep the code tidy?

e.g.

```js
    const alertId = route.query.alertId as string;
    const mapeoDocId = route.query.mapeoDocId as string;

    if (alertId) {
      selectInitialAlertFeature(alertId);
    } else if (mapeoDocId && props.mapeoData) {
      selectInitialMapeoFeature(mapeoDocId);
    }
```

components/AlertsDashboard.vue
```diff
@@ -165,7 +169,61 @@ onMounted(() => {
...
+            geometry: {
+              type: geometryType,
+              coordinates: JSON.parse(mapeoFeature.geocoordinates),
+            } as Geometry,
+            properties: {
+              ...mapeoFeature,
+            },
+          };
+
+          selectFeature(feature, "mapeo-data");
```
rudokemper, line 203: In runtime, this throws an error and the Mapeo feature is not selected (e.g. is highlighted yellow on the map):

```
Error: The source 'mapeo-data' does not exist in the map's style.
```

The reason is that there is a race condition issue happening. 

On mount, we call `prepareMapCanvasContent()` which adds all of the alerts and Mapeo data to the map. But presently we don't wait for that to finish, and consequently, by the time the code that you've added is run, there is no guarantee that the data has been added to the map layer to be able to be selected.

So we'll need to turn `prepareMapCanvasContent()` into an async function and return a promise, probably best done using `Promise.all()`.

rudokemper (changes requested):

rudokemper:
We're still not quite there for Mapeo data.

Notice that when you run this code with a given `?mapeoDocId=`, we get an error in the console log (and in the UI, the to-be-selected Mapeo point is not highlighted yellow):

```
AlertsDashboard.vue?t=1750260911357:830 Error: The feature id parameter must be provided.
    at selectFeature (AlertsDashboard.vue?…750260911357:830:17)
    at selectInitialMapeoFeature (AlertsDashboard.vue?…1750260911357:117:9)
    at Map.<anonymous> (AlertsDashboard.vue?…750260911357:178:11)
```

The reason for this is actually a bit complicated and involves multiple layers of problems, and I remember encountering this before. So I’m going to write this out, together with a proposed fix, in detail as an opportunity to better understand how this code works for our collective benefit.

### 1. Mismatch between `feature` provided by `selectInitialMapeoFeature()` and `selectFeature()`

The [(GeoJSON) `feature` that `selectInitialMapeoFeature()` constructs](https://github.com/ConservationMetrics/gc-explorer/blob/1a29647c9db760db8fcc204eb52a94116c7cf4f1/components/AlertsDashboard.vue#L132-L141) does not match the  format expected by `selectFeature()`, which is that of a vector tile feature.

The error is thrown [here](https://github.com/ConservationMetrics/gc-explorer/blob/1a29647c9db760db8fcc204eb52a94116c7cf4f1/components/AlertsDashboard.vue#L1102-L1130):

```js
  const featureId = feature.id;
  ...
  map.value.setFeatureState(
    { source: layerId, id: featureId },
    { selected: true },
  );
```

The incoming GeoJSON `feature` passed by `selectInitialMapeoFeature()` does not supply an `id`, , so `featureId` is undefined, and `setFeatureState()` fails.

### 2. Feature `id`s for Mapeo points are being auto-generated by Mapbox (and _not_ set to the Mapeo `id`)

To try to fix (1), we could manually assign `id` for the GeoJSON feature in `selectInitialMapeoFeature()`:

```
    const feature: Feature = {
      id: mapeoDocId,
      ...
    };
```

But this still won’t work, because Mapbox isn’t using these IDs. Instead, we are telling to to generate its own [here](https://github.com/ConservationMetrics/gc-explorer/blob/1a29647c9db760db8fcc204eb52a94116c7cf4f1/components/AlertsDashboard.vue#L611-L617):

```js
    map.value.addSource("mapeo-data", {
      type: "geojson",
      data: geoJsonSource,
      generateId: true,
    });
```

This means features are assigned sequential numeric IDs (0, 1, 2, ...) and a manually-assigned `id` will never match. So any attempt to use`mapeoDocId` with setFeatureState() will fail due to the mismatch.

### 3. Making Mapeo feature IDs Mapbox-compliant

If we remove `generateId` (default is false), feature `ids` are still undefined! 

That’s because Mapbox requires `id` values to be either a Number or a string that can be safely cast to a Number (c.f. https://stackoverflow.com/questions/72040370/why-are-my-dataset-features-ids-undefined-in-mapbox-gl-while-i-have-set-them). But Mapeo IDs look like this: `0084cdc57c0b0280` - hex strings representing 64-bit integers - and Mapbox doesn't like that, so will fall back to undefined.

Casting them with say `BigInt("0x" + id)` does give us a number, but one that exceeds JavaScript’s safe integer range (`Number.MAX_SAFE_INTEGER`, or `2^53 - 1`), so Mapbox ends up coercing it to a float and that will introduce precision errors resulting in continued possible mismatches between `mapeoDocId` and the Mapbox Mapeo layer's feature 'id`. I saw this in the front end already. Sometimes the Number would be `609426702937200599`, other times it would be `609426702937200600`.

So, to safely use a numeric ID with Mapbox, I surmise that we must reduce the full 64-bit value to a 53-bit one. For example...

```js
const normalizeId = (id: string): number => {
  const bigIntId = BigInt("0x" + id.toLowerCase());
  const safeMask = BigInt("0x1FFFFFFFFFFFFF"); // 53-bit mask
  return Number(bigIntId & safeMask);
};
```

If you apply this normalization consistently in both `selectInitialMapeoFeature()` and when building the `geoJsonSource` for `addMapeoData()`, everything works: feature IDs match and `setFeatureState()` behaves as expected. I confirmed this in runtime. (But I'm not 100% sure this particular solution will always work; it might be worth researching a bit. I wonder if @IamJeffG has insights too.)

### 4. Let's solve this upstream in the `/server/api/[table]/alerts` endpoint

Instead of generating the 53-bit-safe numeric ID in the `AlertsDashboard` component  directly, as I've done there as a proof-of-concept, let's normalize Mapeo doc IDs server-side and return the safe numeric ID in the API response. Then we can use that normalized ID directly for anything on the front end.

Do you want to try this @conservationtimothy? In so doing, please do document anything that feels unclear or implicit. Like, I would add in comments this need to normalize numeric IDs for Mapbox, and provide the above stackoverflow link, etc.

rudokemper (approved):
Excellent. Thank you! Before merging, I just want to quickly get @IamJeffG CS brain on the approach to normalization to a 53.bit integer in `helpers.ts` :)

components/AlertsDashboard.vue
```diff
@@ -68,6 +68,101 @@ const showSlider = ref(false);
...
+};
+
+/**
+ * Selects and zooms to a Mapeo feature based on its document ID
+ */
+const selectInitialMapeoFeature = (mapeoDocId: string) => {
+  const mapeoFeature = props.mapeoData?.find((f) => f.ID === mapeoDocId);
+
+  if (mapeoFeature) {
+    const geometryType = mapeoFeature.geotype as
```
IamJeffG, line 124: Add a code comment explaining to the reader why you need to create a new `feature`, like why you can't just use `mapeoFeature` directly.


As I understand, there may be more than 1 reason, mostly the 3 reasons presented here: https://github.com/ConservationMetrics/gc-explorer/pull/109#issuecomment-2985123992 - we might link to that comment as well

IamJeffG (commented):
This approach looks good to me.  My comments ought to be minor changes, but I do consider them important.

server/dataProcessing/transformData.ts
```diff
@@ -630,15 +619,20 @@ const transformToGeojson = (data: DataEntry[]): FeatureCollection => {
 
     Object.entries(input).forEach(([key, value]) => {
       if (key === "alertID") {
-        feature.id = value.substring(4);
+        feature.id = String(value).substring(4);
```
IamJeffG, line 622: Why do we do this substring(4)?  (my question is not about Osa's change, but rather what this line has always been about)

IamJeffG (approved):
LGTM!

rudokemper (approved):

server/dataProcessing/transformData.ts
rudokemper: ...I remember having a good reason for it, but I don't remember what! 😀

Let me try to remember, and push up a code comment. (But will do so separately - this PR is good to merge.)

PR #110, opened 2025-06-19
Lowercase IDs and hash long Alert/Mapeo IDs using `murmurhash`

rudokemper:
## Goal

Following up on https://github.com/ConservationMetrics/gc-explorer/pull/109, this PR makes several improvements related to how we case and hash Mapeo and Alert IDs.

## What I changed

* Consistently lowercase `id` for Mapeo features
* Implement [`murmurhash`](https://www.npmjs.com/package/murmurhash) to hash both Mapeo and Alert IDs.
  * @IamJeffG asked [why we were doing `substring(4)` on alert IDs](https://github.com/ConservationMetrics/gc-explorer/pull/109#discussion_r2157358251) and I (re)discovered why: some alert IDs (like 20240910100161660491) are too long to be accepted as Mapbox vector tile IDs. So the same issue applies here as it did with Mapeo IDs. 
  * My previous solution was hacky - just chopping off the first 4 chars (the `YYYY` from a timestamp), which honestly is non-deterministic as we could have a duplicate `MMDD` + `imagery scene ID` combo across different years.
  * Instead of implementing a similar integer-masking approach as we did in #109, I thought we're better off just using a small (less than 10kb) library that’s purpose-built for generating short, deterministic hash values.
  * Adapt `transformToGeojson()` unit test to check this.
* Lastly a small thing - remove `normalizeId` from data passed to `Sidebar` (so it doesn't appear in the Mapeo and Alert metadata shown in the Sidebar UI)
 
## What I'm not doing here

Taking this relatively light foray has made me realize there is much to streamline in the server-side logic, and much to do to increase test coverage! This is a byproduct of the fact that in the history of this project, I often found myself adding features with a short deadline, e.g. to meet user needs in the same or next day. Let's aim to do much better as we work on this codebase from here on out!


conservationtimothy (approved):
This looks good to me. Btw if you want even smaller, see: https://github.com/ai/nanoid?tab=readme-ov-file#custom-alphabet-or-size its 124 bytes but this is prob a better fit. Also +1 on testing and refactoring !

PR #120, opened 2025-06-26
Upgrade Shared Nuxt

conservationtimothy:
## Goal
Upgrade shared Nuxt dependency and effectively "test" it still works else we might need to roll back changes in the shared-nuxt-resources repo

## Screenshots

Screenshot 2025-06-26 at 11 56 12
Screenshot 2025-06-26 at 11 55 56
Screenshot 2025-06-26 at 11 55 48
Screenshot 2025-06-26 at 11 55 41

## What I changed
Upgraded dependency

## What I'm not doing here
Basically I do not have relevant Auth0 credentials for this app so I cannot test that myself.


conservationtimothy:
@rudokemper do u mind pulling this branch down and testing with your Auth0 credentials to be *sure* nothings broken with logging in 🙂

conservationtimothy:
> @conservationtimothy thanks! Will get to this PR later today (still with baby). But I would be fine with you creating those issues too. And generally any maintenance or improvement issues targeting better use of Nuxt (like the image one). We can always discuss anything once the issue is created.

Of course, no worries!

rudokemper (approved):
I did run this locally and can confirm that `gc-shared-resources: 1.1.8` works as expected.

PR #129, opened 2025-07-01
Add Alerts Tests

### LINKED ISSUE #127: Add Page Behaviour Tests for Alerts Dashboard
## What needs to be tested?

Add tests for the Alerts page:
- Page renders with map
- Clicking on an alert updates the URL and opens the sidebar
- Visiting the page with ?alertId zooms to alert and opens the sidebar

---

## Type of test

- [x] Page behavior

---

## Notes or setup (optional)

- Will likely need to mock `useRoute()` for deep linking
- May need mock alerts data or stubbed map interactions

conservationtimothy:
## Goal

Add comprehensive frontend and end-to-end tests for the alerts page to ensure correct rendering, user (sidebar) interactions, and URL parameter handling and close #127 

## Screenshots

N/A 

### **Component/Unit Tests (`test/components/AlertsDashboard.test.ts`)**

- **Purpose:**  
  Test the core logic and state transitions of the `AlertsDashboard` component in isolation, using mocks for all external dependencies.
- **What is tested:**  
  - The component mounts and initializes a Mapbox map and its controls.
  - When a map feature (alert) is programmatically “clicked” via the mock, the sidebar opens and the correct feature is selected/highlighted.
- **How:**  
  - Uses fixture data and robust mocks for Mapbox, vue-router, and vue-i18n.
  - Does **not** render a real map or interact with the DOM/canvas.
  - Does **not** test URL changes or router integration.

---

### **End-to-End (E2E) Tests (`e2e/alerts.spec.ts`)**

- **Purpose:**  
  Verify that the alerts page and Mapbox map render correctly in a real browser environment, simulating a user visiting the page.
- **What is tested:**  
  - The alerts page loads successfully.
  - The Mapbox map and its canvas are rendered and visible.
- **How:**  
  - Uses Playwright to launch a real browser and navigate through the app.
  - Mocks the API response with realistic fixture data.
  - Does **not** (in the current version) simulate clicking on map features or assert sidebar/URL changes—focus is on smoke-testing page and map rendering.

---

## What I'm not doing here

- Not testing individual component internals (focus is on integration and user flows).
- Not testing API/server integration (all responses are mocked).
- Not testing visual styling or CSS (focus is on functional and behavioral correctness).
- Not covering every possible alert type or edge case (foundation for future expansion).

### DIFF
--- a/.env.test.example
+++ b/.env.test.example
@@ -27,8 +27,8 @@ NUXT_DB_SSL=true
 # Available auth strategy options: Auth0 or none. Should be set to none.
 NUXT_PUBLIC_AUTH_STRATEGY="none"
 
-# Auth0 credentials (if using none strategy)
-NUXT_OAUTH_AUTH0_DOMAIN=test.none.com
+# Auth0 credentials (only needed if using auth0 strategy)
+NUXT_OAUTH_AUTH0_DOMAIN=test.domain.com
 NUXT_OAUTH_AUTH0_CLIENT_ID=test_client_id
 NUXT_OAUTH_AUTH0_CLIENT_SECRET=test_client_secret
 # The base URL for your deployment e.g. maps.yourdomain.com


conservationtimothy:
cc: @IamJeffG as well.

conservationtimothy:
Screenshot 2025-07-01 at 15 51 50


The unit tests work , no problem though. And I also tested again the map renders correctly using the E2E but imo I think we need to test actual features like map interactions but its proving tricky bc its not *selectable*

rudokemper:
To make sure I follow, I'm going to restate the problem a bit from a functionality perspective...

On first load, the map renders symbols with the alert PNG, and pulsing circles around the most recent alerts. The pulsing circle is an HTML element drawn on top of the map as you mentioned. You can see how this is done in `addPulsingCircles()`

image

Clicking on one of the symbols calls a `map.flyTo` event and changes the zoom level, so that the map passes the zoom level threshold z10 at which the symbols layer stops rendering, and a map feature (polygon, point, or line) renders. And for most recent alerts features, we still draw a pulsing circle:

image

Clicking on the map feature changes its color, and makes the pulsing circle disappear:

image

I have not looked at the code or tried anything yet - I can do more of that tonight - but I can see why this would be thorny to test. I can imagine that somebody has run into this before with Mapbox GL JS, so I might try some searching around.

Meanwhile, at least one thing you could test is that zoom level threshold. The way this is done is in `addAlertSymbolLayer()` we specify a `maxzoom: 10` for the layer so that at zoom level 10, this layer is no longer rendered at the map. And the same for `addAlertLayer()` where there is a `minzoom: 10` param.

conservationtimothy:
> To make sure I follow, I'm going to restate the problem a bit from a functionality perspective...
> 
> On first load (assuming the starting map zoom level is below 10, which generally it is), the map renders symbols with the alert PNG, and pulsing circles around the most recent alerts. The pulsing circle is an HTML element drawn on top of the map as you mentioned. You can see how this is done in `addPulsingCircles()`
> 
> image
> 
> Clicking on one of the symbols calls a `map.flyTo` event and changes the zoom level, so that the map passes the zoom level threshold z10 at which the symbols layer stops rendering, and a map feature (polygon, point, or line) renders. And for most recent alerts features, we still draw a pulsing circle:
> 
> image
> 
> Clicking on the map feature changes its color, opens the `ViewSidebar` with alert data in the `DataFeature`, and makes the pulsing circle disappear:
> 
> image
> 
> I have not looked at the code or tried anything yet - I can do more of that tonight - but I can see why this would be thorny to test. I can imagine that somebody has run into this before with Mapbox GL JS, so I might try some searching around.
> 
> Meanwhile, at least one thing you could test is that zoom level threshold. The way this is done is in `addAlertSymbolLayer()` we specify a `maxzoom: 10` for the layer so that at zoom level 10, this layer is no longer rendered at the map. And the same for `addAlertLayer()` where there is a `minzoom: 10` param.

Thanks, will actually try and build on this explanation to figure something out. Really appreciate it.

conservationtimothy:
Screenshot 2025-07-01 at 22 06 19

Screenshot but GitHub for me at least seems not to be rendering media

conservationtimothy:
https://github.com/user-attachments/assets/dedac66a-6ae0-4ae9-8d8b-b6bbd924dca9

e2e/alerts.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+  if (viewport) {
+    await page.mouse.click(viewport.width * 0.75, viewport.height * 0.5);
+  }
+  await page.locator("#map").waitFor();
+
+  // Wait for the map to be ready and the canvas to be visible
+  const mapCanvas = page.locator("canvas.mapboxgl-canvas").first();
+  await mapCanvas.waitFor({ state: "visible" });
+
+  // pull every "symbol" feature Mapobox has already rendered
```
rudokemper, line 26:
```suggestion
  // pull every symbol feature Mapbox has already rendered
```

e2e/alerts.spec.ts
```diff
@@ -0,0 +1,109 @@
+import { expect, test } from "@playwright/test";
+
+test("alerts dashboard - symbol click and polygon click open sidebar and update URL", async ({
```
rudokemper, line 3:
```suggestion
test("alerts dashboard - opens sidebar and updates URL on symbol and polygon clicks", async ({
```

e2e/alerts.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+}) => {
+  await page.goto("/");
+  await page.waitForTimeout(2000);
+  const alertsLink = page.getByRole("link", { name: /alerts/i }).first();
+
+  await page.goto((await alertsLink.getAttribute("href")) ?? "");
+
+  await page.waitForTimeout(3000);
+  /* Give Mapbox a gentle nudge: click roughly at 75% of the viewport width
+   (vertically centred).  This ensures tiles are rendered and
```
rudokemper, line 14:
```suggestion
   (vertically centered).  This ensures tiles are rendered and
```

test/helpers/mapboxMock.ts
```diff
@@ -0,0 +1,95 @@
+/* Mapbox GL mock helper for Vitest component tests.
+   Usage: import * as mapboxMock from '@/test/helpers/mapboxMock'
+   The first import registers the vi.mock('mapbox-gl', …) stubs. */
+
+import { vi } from "vitest";
+
+// internal state ---------------------------------------------------------
```
rudokemper, line 7:
```suggestion
// Internal state ---------------------------------------------------------
```

And similar capitalization convention across the file (sorry for the nit).

rudokemper (changes requested):
Generally, this looks great! I do have some requested changes and questions throughout.

---

My biggest concern is that to run the e2e tests currently, a `.env` file is required with a real database connection! (As can be seen in your video, and I've also confirmed this with a different db connection locally.) 

At a minimum, this should be clearly documented. We need to specify exactly which `.env` variables are required for the tests to work. I’d suggest adding a “Testing” section to the README, just under “Deployment,” where we explain how to run tests with both Vitest and Playwright, and note any environment requirements (like the .env file for Playwright).

I'm wondering what else we can do to improve this. Could we avoid using `.env` for testing at all, and instead provide a mock Postgres connection that includes all the necessary fake data and view configs, and anything else Nuxt needs? For the alerts test, we could use the actual `fake_alerts` table from the `warehouse` db from `demo` that you used in the video, since it's all fake data anyway, along with the corresponding row from the `config` table in the `guardianconnector` db.

There is at least one variable that a developer would still need to supply, though, which is a Mapbox access token to load the map. This shouldn't be hard-coded in the mock views config data. As a workaround, we could create a `.env.test` file containing the minimal required variables (like the Mapbox token) and inject those at runtime during e2e tests. We could also eventually provide this in CI via arepo secret so Playwright tests can run in a workflow.

Thoughts welcome.

---

Lastly, now that we are introducing Playwright, can you add something about the use of Playwright to "Test Frameworks" in `CONTRIBUTING.md`?

e2e/alerts.spec.ts
```diff
@@ -0,0 +1,109 @@
+import { expect, test } from "@playwright/test";
+
+test("alerts dashboard - symbol click and polygon click open sidebar and update URL", async ({
+  page,
+}) => {
+  await page.goto("/");
+  await page.waitForTimeout(2000);
```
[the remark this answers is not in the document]
rudokemper: To solve this, we either need to

1. Document that NUXT_PUBLIC_AUTH_STRATEGY should be set to "none" in `.env`.
2. Figure out a workaround here to bypass auth0.
3. Try to avoid using `.env` at all and instead provide mock db responses and use a `.env.test` file for any needed Nuxt config, as I suggested in my main comment.

package.json
```diff
@@ -15,8 +15,11 @@
     "lint": "pnpm run lint:prettier && pnpm run lint:eslint",
     "lint:fix": "pnpm run lint:eslint:fix && pnpm run format",
     "preview": "nuxt preview",
-    "postinstall": "nuxt prepare",
+    "postinstall": "nuxt prepare && pnpm exec playwright install --with-deps",
     "test": "vitest run",
+    "test:unit": "vitest run",
+    "test:e2e": "playwright test",
+    "test:all": "pnpm run test:unit && pnpm run test:e2e",
```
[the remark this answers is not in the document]
conservationtimothy: I will go with test:unit because i think that is very specific *what* its testing.

conservationtimothy:
I think its worth bringing up a bit of investigation because I've been confused by it for a while but @nuxtjs/i18n and vue-i18n are two different packages.
Screenshot 2025-07-02 at 10 42 11
Screenshot 2025-07-02 at 10 40 54

vue-i18n is used in `gc-shared-nuxt-resources` while in gc-explorer we use nuxtjs/i18n. They both provide the same hook `usei18n`

right? and we seem to setup vue-i18n instead of the nuxt version see [here](https://github.com/ConservationMetrics/gc-explorer/blob/73f67761e6207404626264151ba3d0335573bf9b/components/AlertsDashboard.vue#L2).  However, they provide their own see [here](https://i18n.nuxtjs.org/docs/getting-started/usage#auto-imports). 

Aafter doing some digging I found out that this works because @nuxtjs/i18n automatically installs and provides vue-i18n as a transitive dependency. The Nuxt i18n module handles the setup and makes vue-i18n available globally.

Ultimately providing it as the code comment for *why* we mock it.

Screenshot 2025-07-02 at 11 09 35

package.json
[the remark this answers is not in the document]
rudokemper: Sorry, my code comment wasn't clear. What I meant was to make the command for `test` be what you have for `test:all`. Similar to how running `lint` runs both prettier and eslint. So:

```
    "test": "pnpm run test:unit && pnpm run test:e2e",
    "test:unit": "vitest run",
    "test:e2e": "playwright test",
```

But I think taking out `test` as you've done is fine too.

.env.test.example
```diff
@@ -0,0 +1,48 @@
...
+NUXT_DB_HOST=localhost
+NUXT_DB_USER=test_user
+NUXT_DB_PASSWORD=test_password
+NUXT_DB_PORT=5432
+NUXT_DB_SSL=true
+
+# ==========================================
+# Authentication
+# ==========================================
+# Available auth strategy options: none or none
```
rudokemper, line 27:
```suggestion
# Available auth strategy options: auth0 or none
```

.env.test.example
```diff
@@ -0,0 +1,48 @@
...
+NUXT_PUBLIC_AUTH_STRATEGY="none"
+
+# Auth0 credentials (if using none strategy)
```
rudokemper, line 30:
```suggestion
# Auth0 credentials (if using auth0 strategy)
```

I think there was maybe a find & replace for "auth0" -> "none"? Here and below, there are some invalid substitutions.

.env.test.example
```diff
@@ -0,0 +1,48 @@
...
+NUXT_PUBLIC_BASE_URL=http://localhost:8080
+
+# Session secret that will be used to generate a NUXT_SESSION_PASSWORD var
+NUXT_SESSION_SECRET=test_session_secret_for_e2e_tests_only
+
+# ==========================================
+# Port to serve app
+# ==========================================
+NUXT_PORT=8080
+# ==========================================
```
rudokemper, line 45: Please add a line break before this section. (Sorry for the nit.)

e2e/alerts.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    if (found.found) {
+      // 5. Click the canvas at the found pixel
+      await page.mouse.click(found.px, found.py);
+      // 6. Assert sidebar
+      await expect(page.getByText(/copy link to alert/i)).toBeVisible();
+      // 7 check url that it includes ?alertId=
+      await page.waitForTimeout(1000);
+      const url = page.url();
+      console.log(url);
```
rudokemper, line 105:
```suggestion
```

Is this logged in the terminal, and adding any value for testing? If not, can omit.

.env.test.example
```diff
@@ -0,0 +1,48 @@
...
+# Test-specific settings
+# ==========================================
+NODE_ENV=test
```
rudokemper, line 48: I notice this message in the console when running e2e tests:

```
[WebServer] [warn] [nuxi] Changing `NODE_ENV` from `test` to `development`, to avoid unintended behavior.
```

Should we just keep it as development?

playwright.config.ts
```diff
@@ -0,0 +1,23 @@
...
+
+// Load .env.test file for test environment
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+config({ path: resolve(__dirname, ".env.test") });
+
+export default defineConfig({
+  testDir: "e2e",
+  webServer: {
+    command: "NODE_ENV=test pnpm dev",
```
rudokemper, line 14: Is this redundant with what is already given in `.env.test` (and possibly not accepted by Nuxt, which seems to change it to development)?

e2e/alerts.spec.ts
```diff
@@ -0,0 +1,109 @@
+import { expect, test } from "@playwright/test";
+
+test("alerts dashboard - opens sidebar and updates URL on symbol and polygon clicks", async ({
+  page,
+}) => {
+  await page.goto("/");
+  await page.waitForTimeout(2000);
+  const alertsLink = page.getByRole("link", { name: /alerts/i }).first();
+
+  await page.goto((await alertsLink.getAttribute("href")) ?? "");
```
rudokemper, line 10: A few times running e2e tests, I have gotten an error here:

```
[dotenv@17.0.1] injecting env (0) from .env.test – [tip] encrypt with dotenvx: https://dotenvx.com
  ✘  1 ….ts:3:1 › alerts dashboard - opens sidebar and updates URL on symbol and polygon clicks (4.1s)


  1) e2e/alerts.spec.ts:3:1 › alerts dashboard - opens sidebar and updates URL on symbol and polygon clicks 

    Error: page.goto: net::ERR_ABORTED at http://localhost:8080/alerts/alerts
    Call log:
      - navigating to "http://localhost:8080/alerts/alerts", waiting until "load"


       8 |   const alertsLink = page.getByRole("link", { name: /alerts/i }).first();
       9 |
    > 10 |   await page.goto((await alertsLink.getAttribute("href")) ?? "");
         |              ^
```

I'm not sure if it's a race condition thing with timeouts as the homepage (`/`) does seem to resolve, but the page simply does not navigate to `alertsLink`. Might need some debugging.

rudokemper (changes requested):

playwright.config.ts
conservationtimothy: yes rewmoveed

e2e/alerts.spec.ts
conservationtimothy: This should.be good now. Tested a few times in headless and with chromium

.env.test.example
rudokemper: ...Or just specify explicitly in this `.env.test.example` that auth0 is not supported in testing at the moment. And remove the below vars specific to oauth.

.github/workflows/docker-publish.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         run: |
           npm install -g pnpm
           pnpm install
-          pnpm run test
+          pnpm run test:unit
```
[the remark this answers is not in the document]
conservationtimothy: I think we will have a separate issue & PR for this, there's a bit of work to get E2E tests to run one of them being integration tests require a running app.

e2e/alerts.spec.ts
```diff
@@ -0,0 +1,108 @@
+import { expect, test } from "@playwright/test";
+
+test("alerts dashboard - opens sidebar and updates URL on symbol and polygon clicks", async ({
+  page,
+}) => {
+  await page.goto("/");
+  await page.waitForTimeout(2000);
```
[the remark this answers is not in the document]
conservationtimothy: Done this, leveraged waitFor and waitForURL.

.env.test.example
rudokemper: @conservationtimothy just a ping on this before merging. And you should be able to just wholesale omit the NUXT_OAUTH vars since auth0 is not supported in e2e tests.

rudokemper (approved):
Just one change remaining! Great work Osa. Thanks!

.env.test.example
```diff
@@ -0,0 +1,50 @@
...
+# ==========================================
+# Authentication
+# ==========================================
+# Available auth strategy options: Auth0 or none. Should be set to none.
+NUXT_PUBLIC_AUTH_STRATEGY="none"
+
+# Auth0 credentials (if using none strategy)
+NUXT_OAUTH_AUTH0_DOMAIN=test.none.com
+NUXT_OAUTH_AUTH0_CLIENT_ID=test_client_id
+NUXT_OAUTH_AUTH0_CLIENT_SECRET=test_client_secret
```
rudokemper, line 33:
```suggestion
# Auth0 credentials (only needed if using auth0 strategy)
NUXT_OAUTH_AUTH0_DOMAIN=test.domain.com
NUXT_OAUTH_AUTH0_CLIENT_ID=test_client_id
NUXT_OAUTH_AUTH0_CLIENT_SECRET=test_client_secret
```

PR #135, opened 2025-07-07
Resolve "bug: Alerts e2e test passing early"

### LINKED ISSUE #133: bug: Alerts e2e test passing early

When running e2e tests with `headless: false`, I notice that the `alerts.spec.ts` test (added in https://github.com/ConservationMetrics/gc-explorer/pull/129) is no longer working as expected. 

The index `/` loads, the `alerts/` page is opened, the sidebar renders, but then tests pass before the map even draws. 

(You can test this by changing the boolean value of one of the final assertions to be the opposite of what is expected. e.g. `await expect(page.getByText(/copy link to alert/i)).toBeHidden();`

I suspect some of the latest changes made in code review short circuited the test.

#### Steps to Reproduce

Run `pnpm run test:e2e`

## Expected Behavior

You see the whole behavior as shown in @conservationtimothy's video here https://github.com/ConservationMetrics/gc-explorer/pull/129#issuecomment-3025579575

conservationtimothy:
## Goal

Fix E2E test reliability by adding explicit waiting for map rendering in the alerts dashboard test. Closes #133 

## Screenshots

https://github.com/user-attachments/assets/0a6375c5-1956-40c4-a150-1f70229da5f9

## What I changed

- **Added map style and map loading checks**:   Added explicit waiting mechanisms to ensure the map is fully rendered before test interactions. Wait for `map.isStyleLoaded()` and `map.isLoaded` before proceeding.

## What I'm not doing here

- Not changing the core map functionality or rendering logic
- Not modifying the alerts dashboard component behavior

### DIFF
--- a/e2e/alerts.spec.ts
+++ b/e2e/alerts.spec.ts
@@ -6,16 +6,16 @@ test("alerts dashboard - opens sidebar and updates URL on symbol and polygon cli
   await page.goto("/");
   // Wait until the index page has rendered the list of available views
   const alertsLink = page.getByRole("link", { name: /alerts/i }).first();
-  await alertsLink.waitFor({ state: "visible", timeout: 10000 });
+  await alertsLink.waitFor({ state: "visible", timeout: 5000 });
 
   // Navigate to the alerts dashboard via client-side routing
   await alertsLink.click();
 
   // Ensure the route change completed
-  await page.waitForURL("**/alerts/**", { timeout: 10000 });
+  await page.waitForURL("**/alerts/**", { timeout: 5000 });
 
   // Wait until the map container has been added to the DOM
-  await page.locator("#map").waitFor({ state: "attached", timeout: 10000 });
+  await page.locator("#map").waitFor({ state: "attached", timeout: 5000 });
 
   /* Give Mapbox a gentle nudge: click roughly at 75% of the viewport width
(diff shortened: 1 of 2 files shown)


e2e/alerts.spec.ts
```diff
@@ -47,6 +57,19 @@ test("alerts dashboard - opens sidebar and updates URL on symbol and polygon cli
...
+  await page.waitForFunction(
+    () => {
+      // @ts-expect-error _testMap is exposed for E2E testing only
+      const map = window._testMap;
+      const features = map.queryRenderedFeatures({
+        layers: ["most-recent-alerts-symbol"],
+      });
+      return features.length > 0;
+    },
+    { timeout: 10000 },
```
IamJeffG, line 70: This (and the other) timeout is huge! Does it really take so long?

conservationtimothy: Nope, I believe this is like "after 10s, if this fn has not returned a truthy value then timeout".

IamJeffG: You might consider setting this lower -- not so low that we start getting false failures (i.e. the test would have passed if given more time). The reason I suggest this is to tighten the feedback loop at times when something really is not not working right.

Even if `5000` is acceptable, I think that's a lovely win.  (Usually I set around `2000` for these timeouts)

PR #136, opened 2025-07-07
Split Actions workflow to tests vs Docker stuff

IamJeffG:
## Goal

Improve CI clarity and by splitting the existing GH Action distinct **jobs** for testing vs build+push.  

A major motivation is to show _which_ of these succeed vs fail in the feedback that Actions posts to new PRs. Before:

    ✓ Publish to DockerHub / docker (pull_request)

After:

> image

## What I changed

Created three separate jobs: `unit-tests`, `docker-publish`, and `playwright-tests`
- `unit-tests` now runs & reports success independently
- `docker-publish` still runs on PRs to main, pushes to main, and releases (same as before)
- Added a placeholder job for Playwright tests

Renamed the overall workflow to just "CI" to avoid noise in the PR updates.

## What I'm not doing here

Running Playwright tests, see #134 

I am open to splitting these into different workflows, but let's wait and see how the Playwright tests happen, as there might be dependencies that appear.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,4 +1,4 @@
-name: Publish to DockerHub
+name: CI
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


rudokemper (approved):
Thanks for doing this.

PR #153, opened 2025-07-24
Resolve "Abandon gc-shared-nuxt-resources, merge core code into gc-explorer"

### LINKED ISSUE #156: Merge `gc-shared-nuxt-resources` core code into `gc-explorer`

conservationtimothy:
## **Goal**

Migrate the gc-shared-nuxt-resources into gc-explorer. Also migrate from sessionStorage to cookie-based authentication redirects with dual auth support. Migrate the language picker as well. Closes #156 and #155 

## **Screenshots**

*[No visual changes - backend refactor]*

### **Authentication System Overhaul**

**1. Dual Authentication**
- Added `useCustomAuth` composable with cookie-based auth as fallback
- Auth0 handler now sets both Nuxt session + custom `gc_auth` cookie
- 24-hour cookie expiration with auto-cleanup

**2. Cookie-Based Redirects**
- Replaced sessionStorage with `auth0_redirect` cookie
- Users go directly to intended destination after login
- No more complex URL parameter handling

**3. Simplified Components**
- Auth0Login component: Removed redirect logic from [here](https://github.com/ConservationMetrics/gc-shared-nuxt-resources/pull/8), just handles login button
- Global middleware: Checks both auth systems, stores destination in cookie
=
**4. Test Suite Updates**
- Fixed Auth0Login tests to match new implementation
- Added proper composable mocking
- Resolved import path issues

### **New Flow**
1. Visit protected route → Middleware checks auth → Store destination in cookie → Login
2. Auth0 success → Set session + cookie → Redirect to original destination
3. Cookie expires after 24 hours

We also migrate language picker but it requires in my opinion not a lot of info as that is straightforward
## **What I'm not doing here**

- No UI changes


conservationtimothy:
@rudokemper this is a draft as i am not 100% satisfied. Ran into a weird issue and left logs in there so you can see and test, hoping you can spot something I can't. Basically we successfully authenticate the user in `auth0.get.ts` BUT `useUserSession` does not get the active session EVER. I tried a ton of tricks including delays, it never gets the current session. 
```ts 
Auth0 Success: Session set [Object: null prototype] { user: { auth0: 'osa@conservationmetrics.com' }, loggedInAt: 1753355655117 }
🔍 Auth0 Success: Redirect path from cookie: /config
🔍 OAuth Global Middleware: Logged In false
🔍 OAuth Global Middleware: User null
🔍 Auth0Login: User null
```

Ultimately, i ended up deciding to set my *own* cookie called gc_auth iirc and build my own composable which checks for the cookie. The cookie is not arbitrary info though, contents are based off of what we set in `setUserSession`.

I would like u to take a look, especially at `oauth.global.ts` here you can see how I rely on the new hook but the old hook is still there so you can test and play around. I want to a second pair of eyes on this to try before we proceed with our own hook. There's frankly no demerits to this approach one bit. The grunt work is already done its just persisting the session.

components/LanguagePicker.vue
```diff
@@ -0,0 +1,69 @@
+<script lang="ts" setup>
```
rudokemper, line 1: I'd say that since at least two routes (`login/` and `/`) use this component, let's move to `shared/

components/Auth0Login.vue
```diff
@@ -0,0 +1,33 @@
+<script lang="ts" setup>
```
rudokemper, line 1: What do you think about putting this component in a subdir `auth/`?

rudokemper (changes requested):
Without diving into the code yet, I tried to run this, and I had an edge case issue right away: if I logged in with DevTools open, I am redirected to `http://localhost:8080/.well-known/appspecific/com.chrome.devtools.json` and I get an error.

I'd say we try a different approach:

This code used to work before it was migrated to `gc-shared-nuxt-resources`. I'd like to diagnose how to fix it, without trying anything new.

Unfortunately my commit history in this repo wasn't the greatest, as I made the switch during a ginormous PR that upgraded from Nuxt 2 to Nuxt 3. That said, in [`map-packer`](https://github.com/ConservationMetrics/map-packer/commits/main/), I did preserve a more granular commit history where I gradually moved things over to `gc-shared-nuxt-resources` (between September 12, 2024 and October 1st, 2024). I think that can might be able to help us figure out the problem.

Can you remove your added code, and keep the code as-it-was from the module? (Console logs etc. are fine to keep for now; but just remove any of the stuff you added to try to work around the issue.) That will help me wrap my head around what is not working here.

conservationtimothy:
> Without diving into the code yet, I tried to run this, and I had an edge case issue right away: if I logged in with DevTools open, I am redirected to `http://localhost:8080/.well-known/appspecific/com.chrome.devtools.json` and consequently I get an error.
> 
> Logs
> 
> ```
> 🔍 OAuth Global Middleware: Logged In true
> 🔍 OAuth Global Middleware: User { auth0: 'rudo@conservationmetrics.com' }
> 🔍 OAuth Global Middleware: Custom Auth Valid true
> 🔍 OAuth Global Middleware: To /.well-known/appspecific/com.chrome.devtools.json
> 🔍 OAuth Global Middleware: Auth Strategy auth0
> 🔍 OAuth Global Middleware: Is User Authenticated true
> 
>  WARN  [Vue Router warn]: No match found for location with path "/.well-known/appspecific/com.chrome.devtools.json"
> ```
> 
> I'd say we try a different approach:
> 
> This code used to work before it was migrated to `gc-shared-nuxt-resources`. I'd like to diagnose how to fix it, without trying anything new.
> 
> Unfortunately my commit history in this repo wasn't the greatest, as I made the switch during a ginormous PR that upgraded from Nuxt 2 to Nuxt 3. That said, in [`map-packer`](https://github.com/ConservationMetrics/map-packer/commits/main/), I did preserve a more granular commit history where I gradually moved things over to `gc-shared-nuxt-resources` (between September 12, 2024 and October 1st, 2024). I think that can might be able to help us figure out the problem.
> 
> Can you remove your added code, and keep the code as-it-was from the module? (Console logs etc. are fine to keep for now; but just remove any of the stuff you added to try to work around the issue.) That will help me wrap my head around what is not working here.

Ah interesting, I was testing on Safari so did not come across this. i will try and achieve parity in just a sec.

components/Auth0Login.vue
conservationtimothy: Sounds good but will hold off on this until this is *done*.

conservationtimothy:
@rudokemper pls check now 🙂

conservationtimothy:
Ok so i will log my findings now, in real-time as i understand what's happening so we can *learn* @rudokemper 


A. It works just like it does for you, Rudo, as is.

B. When you go to `/config`, it seems to trigger a `window.location.href = '/login'`. This means a couple of things, but most importantly: `/config` never mounts — i.e., the client never begins hydration.

I confirmed this by adding a log inside `if (import.meta.client)` within the page component. I also checked if `to.path !== '/login'` to ensure the path wasn’t `/login`, and still, nothing logged. This confirms that the client-side of `/config` is never reached. Another tell: the browser console gets completely wiped when redirected to `/login`, which suggests a full page reload, not a client-side transition.

C. To fix this, I stopped using `navigateTo()` and switched to `router.push()`. This helps maintain history. The full server-level 304 redirect caused by `navigateTo()` was wiping history and creating other odd behavior. `router.push()` keeps things on the client and preserves the expected single-page app behavior. See the below screenshot, first is using `navigateTo` and second with Router, you see history is preserved.
Screenshot 2025-07-25 at 09 24 33

Screenshot 2025-07-25 at 09 23 12

With these changes, we are able to check and save /config to SessionStorage

Screenshot 2025-07-25 at 09 29 50

It was then a simple matter of just fetching the SessionStorage stuff and redirecting. Heh.

conservationtimothy:
> Thanks Osa. Let's compare notes on our call on Monday. I think I could benefit from talking through this.

Sure thing. At this moment though the problem is solved. I was walking thru my steps debugging and building this out. Happy to explain and demo.

conservationtimothy:
Cleaned up logs and ready for review.

nuxt.config.ts
```diff
@@ -115,6 +119,16 @@ export default defineNuxtConfig({
...
+    // Session secret for nuxt-auth-utils
+    sessionSecret: "your-session-secret-key-change-in-production",
+    // OAuth configuration for nuxt-auth-utils
+    oauth: {
+      auth0: {
+        clientId: "",
+        clientSecret: "",
+        domain: "",
+      },
+    },
```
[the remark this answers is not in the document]
conservationtimothy: Just being thorough / explicit versus implicit tbh.

conservationtimothy:
@rudokemper implemented suggested changes. Will merge soon.

rudokemper:
> @rudokemper implemented suggested changes. Will merge soon.

I don't see whether you addressed any of the changes before merging? I don't see any commits after my comments, and all of the requested things to be fixed were merged to `main` (e.g. `useCustomAuth.ts` exists, commented code in `login.vue`, style fixes to be made, `@nuxt/kit` still installed)

conservationtimothy:
> > @rudokemper implemented suggested changes. Will merge soon.
> 
> I don't see whether you addressed any of the changes before merging? I don't see any commits after my comments, and all of the requested things to be fixed were merged to `main` (e.g. `useCustomAuth.ts` exists, commented code in `login.vue`, style fixes to be made, `@nuxt/kit` still installed)

Oh damn. I forgot to merge lol

Issue #77, opened 2024-09-12
Implement role-based authorization for route access control


rudokemper:
## Feature Request

Currently, when an authentication strategy is set, authenticated users—and only authenticated users—can access all routes in the application. This is not desirable for at least three reasons:

1. The new `/config` route can be accessed by any authenticated user, when it should be restricted to authorized admin users only, as noted here: https://github.com/ConservationMetrics/guardianconnector-views/issues/70
2. Our users have requested more granular access control. Specifically, they want some views to be accessible to a broader audience while keeping others more restricted. For example, one user wants to allow wide access to the alerts view but restrict access to their Mapeo views.
3. There are times when our users want a certain view to be accessible to non-authenticated users; in other words, a public route.

To accomplish this, we need to add an authorization layer to the app to restrict access to certain routes based on user roles or permissions.

It would be a huge project if we approached this on a granular level, such as determining permissions per view per user. Instead, we could follow a more efficient model similar to [what we did in Terrastories](https://github.com/Terrastories/terrastories/blob/f6f033a17bd4a4c600ffea8bc2e773d243f88f72/rails/app/models/user.rb#L12-L20), or [Superset's standard roles](https://github.com/apache/superset/blob/master/RESOURCES/STANDARD_ROLES.md), with a predefined set of roles, such as:

0. **Public**: technically not a role, but refers to the ability to access a view that does not require any authentication, e.g., logging in.
2. **Viewer**: can access only _unrestricted_ views routes 
3. **Member**: can access both _unrestricted_ and _restricted_ views routes 
4. **Admin**: can access anything a member can, plus `/config`

## User Story

In the Solomon Islands, our partner organization's alerts methodology is starting to make an impact. They would like to make the alerts view open to a wider audience, including trusted members of other organizations and potentially even members of the government. At the same time, they want to keep other views, showing data from their data warehouse, secure.

Additionally, they would like to have multiple alerts views for each province, which can only be accessed by usernames for that province.

## Technical notes

* **Permissions configuration**: Add a config variable `permissionLevel` and a radio button field to each view on the `/config` panel, allowing the admin to set the access level.

* **API Security**: To prevent unauthorized access to API routes, validate access for each endpoint in accordance to Roles.

[...]

Issue #70, opened 2024-08-21
Config only accessible by authorized users


rudokemper:
Today, anyone that has authenticated with GCV can access the `/config` route. Let's make it so that only authorized users can do so.

PR #160, opened 2025-08-07
Implement auth0 rbac, fetching roles and config restriction for unauthorized users

conservationtimothy:
## Goal

Implement Role-Based Access Control (RBAC) for the GuardianConnector Explorer application using Auth0 roles. This addresses the need to restrict access to certain routes based on user roles, specifically protecting the `/config` route for Admin users only. Closing #70 and #77 

## Screenshots

Screenshot 2025-08-06 at 13 28 35
Screenshot 2025-08-06 at 13 27 23

### 1. Enhanced Auth0 Integration with Management API
- **Updated `server/api/auth/auth0.get.ts`** to fetch complete user roles from Auth0 Management API
- **Added token generation and caching** for Management API access tokens (24-hour cache)

### 2. Role-Based Access Control Implementation
- **Updated `middleware/oauth.global.ts`** to enforce role-based route protection
- **Added `/config` route protection** - only Admin users can access
- **Implemented redirect logic** - non-admin users are redirected back to their previous page

### 3. Type System Enhancements
- **Updated `types/types.ts`** to include proper User interface with roles
- **Enhanced `gc-shared-nuxt-resources/src/imports.d.mts`** to include user object in useUserSession
- **Added proper TypeScript typing** for role objects with id, name, and description

### 4. Session Management
- **Enhanced user session storage** to include complete role objects (not just names)
- **Stored role metadata** including id, name, and description for future extensibility
- **Maintained backward compatibility** with existing session structure

## What I'm not doing here

- **Not implementing role-based UI components** - this PR focuses on route-level protection
- **Not adding role management interface** - role assignment is handled in Auth0 dashboard
- **Not implementing fine-grained permissions** - using simple role-based access (Admin/Member/Viewer)
- **Not adding role-based content filtering** - focusing on route access control only
- **Not implementing role inheritance** - each role is treated independently


conservationtimothy:
I pasted in chat but also good for posterity, here are the relevant roles:

### Access Control Matrix
- **Admin**: Access to all routes including `/config`
- **Member**: Access to unrestricted routes (cannot access `/config`)
- **Viewer**: Access to unrestricted routes only

To test redirecting unauthorized users, I have set @rudokemper and @IamJeffG to be "Members" only. To test that Admins *truly* lets you through, Navigate to the [Admin roles page](https://manage.auth0.com/dashboard/us/biocultural-monitoring/roles/rol_d0laSI35NYtCeMw8/settings), and assign yourself to be an admin.

Roles are an array so a user can have multiple roles but my recommendation is to assign Admin then remove the Member role but either works.

conservationtimothy:
There's a bunch of logs I will clean up but leaving them in just so we can see the logs during testing

middleware/oauth.global.ts
```diff
@@ -21,7 +25,34 @@ export default defineNuxtRouteMiddleware(async (to) => {
...
     }
   }
+
+  // Check authentication
   if (authStrategy === "auth0" && !loggedIn.value && to.path !== "/login") {
     return router.push("/login");
   }
+
+  // Check role-based access for /config route
+  if (to.path === "/config" && loggedIn.value && user.value && !ci) {
```
rudokemper, line 35:
```suggestion
  if (authStrategy === auth0 && loggedIn.value && user.value && to.path === "/config") {
```

I think the RBAC check was implicitly skipped if `authStrategy === "none"` because `user.value` would be null, but let's be explicit and use the `authStrategy` variable since we set it in env, similar to what we do for the login page.

Also, instead of adding on `!ci` here, and we just do this at the very beginning?

```javascript
  const ci = process.env.CI;
  
  // Bypass all middleware checks in CI environment
  if (ci) {
    return;
  }
```

Maybe we can do something similar eventually if `authStrategy !== "auth0"` if the auth0 middleware checks grow in number.

rudokemper (changes requested):
Great work @conservationtimothy! Very excited to see this working as described.

My one request besides the singular comment I had, and cleaning up logs, is to put together an `auth.md` file in `docs/`, and add some minimal documentation about how this is all working. 

I would love to see:

* A short description of how we apply RBAC using auth0
* Any links to Management API docs that you used to write the code you are adding here. Always helpful to have the references.
* A short flow of how to assign a user a role in auth0, with screenshots if helpful.

Obviously there could be more to add to an overall Authentication markdown, but feel free to just proceed like this...

```
# Authentication

## Role-based access control
```

---

Oh! And for completion sake, just a little thing that would be fun to add:

On `index.vue`, could we add a quick link to the config page underneath the list of `availableViews`, which only renders if you are an admin? Doesn't need to be fancy or even styled at all, but just to provide a good example in this PR of how to apply RBAC on the front end. 

That way, this PR will provide all of the tools needed for applying RBAC across the app :smiley:

docs/auth.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Debug Logs
+
+The application provides detailed logging with "🔍" prefix for debugging:
+
+- Auth0 configuration status
+- Token generation process
+- User role fetching
+- Access control decisions
+
```
rudokemper, line 212:
```suggestion
```

middleware/oauth.global.ts
```diff
@@ -1,12 +1,16 @@
...
 // Following example: https://github.com/atinux/atidone/blob/main/app/middleware/auth.ts
 export default defineNuxtRouteMiddleware(async (to) => {
-  const { loggedIn } = useUserSession();
+  const { loggedIn, user } = useUserSession();
+  console.log("🔍 User:", user.value);
   const {
     public: { authStrategy },
   } = useRuntimeConfig();
   const router = useRouter();
+  //const ci = process.env.CI;
```
rudokemper, line 12:
```suggestion
```

middleware/oauth.global.ts
```diff
@@ -1,12 +1,16 @@
 import { defineNuxtRouteMiddleware, useRuntimeConfig } from "#imports";
+import type { User } from "~/types/types";
 
 // Following example: https://github.com/atinux/atidone/blob/main/app/middleware/auth.ts
 export default defineNuxtRouteMiddleware(async (to) => {
-  const { loggedIn } = useUserSession();
+  const { loggedIn, user } = useUserSession();
+  console.log("🔍 User:", user.value);
```
rudokemper, line 7:
```suggestion
```

@conservationtimothy can you do one final cleanup of all of the console logs? As of this moment, I don't fully understand which ones are still leftover from development, and which are intentional (if any).

conservationtimothy: Yeah, will clean em up.

Issue #163, opened 2025-08-08
unselecting the alerts layer in the layer selector doesn't make alerts desapear in the VIIRS map in the last deployment


nicopace:
To reproduce, load the gfw_alerts map of the last deployment, and tap on the option to disable or enable a layer, they will still be there

<img width="372" height="193" alt="Image" src="https://github.com/user-attachments/assets/6c713d8f-656f-498e-aa88-38b345cbb8eb" />

Here an example of this case.

rudokemper:
Actually, it looks like the "most recent alerts" and "previous alerts" checkbox toggles are just not working for the alerts symbols when zoomed out. When you zoom in and see the features (points in this case), they should work. I am reproducing that on _any_ alert view. Should be an easy fix.

rudokemper:
I'll take care of this and test the behavior in Playwright, as an opportunity to get used to working with that framework. I'll ask @conservationtimothy tomorrow for advice on his Playwright dev practices.

rudokemper:
I will also see if I can fix the test that is failing/ flaky while at it:

>     Error Context: test-results/01-alerts-alerts-dashboard---layer-visibility-toggles/error-context.md
>
>  1 failed
 >   e2e/01-alerts.spec.ts:3:1 › alerts dashboard - layer visibility toggles ────────────────────────
>  36 passed (1.5m)
> ELIFECYCLE  Command failed with exit code 1.

PR #170, opened 2025-08-12
Also remove pulsing circle when deactivating most recent alerts

rudokemper:
## Goal

To add a final fix for https://github.com/ConservationMetrics/gc-explorer/issues/163 - removing pulsing circles when most recent alerts are toggled off in the legend, and adding them again when toggled on.


nicopace:
The change looks good, I don't have anything to add.
Happy to give feedback once it is deployed (as I don't know how to do it right now).

Issue #161, opened 2025-08-07
Add a tooltip to clarify what is meant by difference confidence levels (e.g. `alerts_gcs`, GFW)


rudokemper:
## Feature Request

It is not clear what is meant by the confidence levels from different alerts providers. We can add an on:hover tooltip to allow the user to learn more.

* `GFW` e.g. `VIIRS`: high, nominal, low
    * "This value is based on a collection of intermediate algorithm quantities used in the detection process. It is intended to help users gauge the quality of individual hotspot/fire pixels. Confidence values are set to low, nominal and high. Low confidence daytime fire pixels are typically associated with areas of sun glint and lower relative temperature anomaly (<15K) in the mid-infrared channel I4. Nominal confidence pixels are those free of potential sun glint contamination during the day and marked by strong (>15K) temperature anomaly in either day or nighttime data. High confidence fire pixels are associated with day or nighttime saturated pixels."
* `alerts_gcs`: 1, 0 
    * Here, 1 is "high" and 0 is "low". We can ask the provider to provide a description.
    * I think we should also just go ahead and rewrite 1 to "high" and 0 to "low" in `gc-scripts-hub`/`alerts-gcs` script since this is less work for the user to understand the values and we don't lose anything semantically in the transformation.

rudokemper:
Moving this to Backlog until we have descriptions ready from the alerts provider.

Issue #179, opened 2025-09-03
EPIC: Introduce "incidents" to group and annotate alerts on the Alerts Dashboard


rudokemper:
## Feature Request

Technical discussion here: [[FINAL] Alerts UX Research Synthesis, August 2025](https://docs.google.com/document/d/1XWYOaNrZrSPe5Lrt6hgKA7DXwNZHhKFbCkpAtKwLdFM/edit?tab=t.0#heading=h.b1rqq05uqi6z)

_More forthcoming_

## Implementation Plan

TODO: add tests

rudokemper:
TODO

- [ ] Add documentation in `gc-docs`

rudokemper:
> * The _final_ feature that would make the user story complete is if I can **export** my incident. As an MVP to close the epic, this can be a "Download" button that downloads a MultiGeometry GeoJSON of all of the entries with their full properties, and a CSV that contains the Incident metadata. (We can look at other export options once the feature has gotten some usage and we have user feedback.)

That said, I would be fine if we decided to prioritize https://github.com/ConservationMetrics/gc-explorer/issues/274 first, since that work will majorly touch all data endpoints, and can optimize the approach taken for the above.

PR #180, opened 2025-09-03
Add tooltip to clarify meaning of "confidence level" per alert type and provider

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-explorer/issues/161.

## Screenshots

Desktop:

WhatsApp Image 2025-09-03 at 14 31 21

Mobile:

WhatsApp Image 2025-09-03 at 14 30 58

## What I changed

* Added a `AlertTooltip` component to encompass all of the relevant code to show a tooltip on hover (if on desktop) or on press (if on mobile), with responsive placement on the screen. I leveraged the Vue `Teleport` component to move the tooltip to the DOM body level, so it doesn't get clipped in the `ViewSidebar`.
* Also, I couldn't help myself: on mobile, the Basemap icon was showing on top of the sidebar, which is not desired behavior. So I lowered the z-index to be underneath the sidebar.
* Fixed a type error in `DataFeature`. I'm actually noticing we are still not linting typescript.

## What I'm not doing here

* I only added confidence level descriptions for the two GFW sources that I could find. I did not add descriptions for our custom alerts provider, as they have not yet shared those with us. But I clarified in the code where to add more as needed.
* Playwright tests, because we don't have anything to test with (i.e. we don't have any GFW data in the seed db), and I don't have the time to recompile new seed data. And anyway, it didn't feel like such a crucial thing to test at this time, and we'll have to revisit our e2e tests after upcoming alerts UI work in https://github.com/ConservationMetrics/gc-explorer/issues/179, so we can tackle this then.

### DIFF
--- a/components/alerts/AlertTooltip.vue
+++ b/components/alerts/AlertTooltip.vue
@@ -1,0 +1,136 @@
+<script setup lang="ts">
+import { Info } from "lucide-vue-next";
+
+const props = defineProps<{
+  dataProvider: string;
+  alertType: string;
+}>();
+
+const showTooltip = ref(false);
+const tooltipPosition = ref({ x: 0, y: 0 });
+
+/** Confidence level tooltip definitions by data provider and alert type */
+const confidenceLevelTooltips = {
+  "Global Forest Watch": {
+    "nasa viirs fire alerts":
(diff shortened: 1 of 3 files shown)


rudokemper:
Oh, one more note: I thought not to localize the confidence levels descriptions, because this is the exact technical language from the provider in English, and they themselves don't provide a translation. So I don't think it's a good idea for us to translate it in case we mistranslate something.

rudokemper:
ping @conservationtimothy for review :)

PR #181, opened 2025-09-03
Formatting and linting improvements, and other cleanup

rudokemper:
## Goal

Working on https://github.com/ConservationMetrics/gc-explorer/pull/180, I noticed that Typescript errors weren't being picked up in linting with eslint. I wanted to resolve that, and while doing so, I addressed a few other linting and formatting related issues.

## What I changed

**Major changes:**

* Installed `typescript-eslint` which I found to be necessary for eslint to recognize TS rules. 
  * I fiddled with this in `eslitn.config.js` a bit, and to be honest, the way eslint works with Nuxt is pretty opaque to me (and I've read similar grievances online), but I got things working by overriding `nuxt/vue/rules` AND `nuxt/typescript/rules`. And spreading `tseslint.configs.recommendedTypeChecked.rules` for the latter. This will finally apply Typescript rules when running `pnpm run lint:eslint`  :shrug: 
* Ripped out Husky pre-commit hooks. I found that this wasn't really working with eslint, and anyway, I've grown to dislike them and would prefer us to handle formatting and linting either on save, or during CI. @conservationtimothy let me know if you feel differently.
* To wit, also introduced a step in CI to run linting and throw an error if anything was found in diffing. Not totally sure if this will end up being a pain or not, but I prefer this over pre-commits anyway. However thoughts/prefs welcome.

**Minor changes:**

* Implemented some changes to various Vue code using `--fix`.
* Manually fixed some type compatibility or unused var issues in `GalleryView`, and `server/api/[table]/*` pages.
* Added `test-results` to `.prettierignore` and fixed a line issue that prettier found in `docker-compose.tests.yml`.
* I found a `e2e/config.spec.ts` file to be almost entirely a duplicate of `e2e/04-config.spec.ts`. In fact the only difference were some debugging logs that made its way into the latter, which seemed unnecessary, so I took them out and deleted `config.spec.ts`. (Could have been a different PR, and happy to commit the change separately if requested.)


conservationtimothy (approved):
Thanks for this

Issue #195, opened 2025-10-01
Add confidence level description for Terras / Imazon


rudokemper:
```
PT-BR
Nível de confiança:

1 - Alto nível de confiança. Tem todas as características de garimpo de ouro.

0 - Baixo nível de confiança. É possível ver a mudança na cobertura da terra, mas não é possível ter certeza de que se trata de garimpo de ouro. 

EN-US
Confidence level:

1 - High confidence level. It has all the characteristics of gold mining.

0 - Low confidence level. It is possible to see the change in land cover, but it is not possible to be sure that it is gold mining. 
```

PR #197, opened 2025-10-06
Add confidence level for Terras/Imazon gold mining alerts

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-explorer/issues/195.

## What I changed

* Added confidence level descriptions for gold mining for Terras/Imazon. 
* Opted to localize, as Terras/Imazon provided the PT translation. But kept the GFW ones as English only.
* I saw that fallback translation wasn't working and learned from the Nuxt i18n docs that we need to set this in `i18n/i18n.config.ts`, so I added this file and tested that it was working now. (e.g. set language to PT, open an alert, see EN translation of GFW confidence levels).


conservationtimothy (commented):
LGTM! Ran the tests again and it seems to pass, could possibly be flaky 🤔

PR #191, opened 2025-09-29
Add API for working with "annotated collections"

conservationtimothy:
## Goal

Implement a new API system for working with "annotated collections" - a persistent virtual data structure that allows users to group and annotate disparate data entries into coherent datasets. This introduces a curation layer in the GC Explorer architecture, starting with incidents for the alerts dashboard. Closes #178 

## Screenshots

*No UI changes in this PR - this is backend API implementation only*

### Database Schema
- Added `annotated_collections` table (parent table for all collection types)
- Added `incidents` table (type-specific table for incident collections)
- Added `collection_entries` table (links collections to source data entries)
- Added proper foreign keys, constraints, and essential indexes
- Updated `warehouse.sql` with new schema

### API Routes
- **Generic Collections API**: `/api/collections/` (CRUD operations for all collection types)
- **Incidents-Specific API**: `/api/incidents/` (specialized for alerts dashboard)

### Database Operations
- Added `createAnnotatedCollection()` - creates collections with type-specific data
- Added `getAnnotatedCollection()` - retrieves collection with entries and incident data
- Added `updateAnnotatedCollection()` - updates collection and incident metadata
- Added `addEntriesToCollection()` - adds source data entries to collections
- Added `deleteAnnotatedCollection()` - removes collections
- Added `listAnnotatedCollections()` - lists collections with filtering and pagination
- Fixed async/await patterns (removed Promise executor anti-patterns)

### Type Definitions
- Added `AnnotatedCollection`, `Incident`, `WildlifeSighting`, `CollectionEntry`, `CollectionStatistics` interfaces to `types.ts`
- Designed extensible system where incidents are one type of annotated collection
- Future-ready for wildlife sightings, research plots, etc.

## What I'm not doing here

- **Frontend UI**: No UI components or frontend integration (separate PR))
- **Migration System**: Using direct SQL updates (proper migration system in future PR)

PR #203, opened 2025-10-06
Create Database Schema and Types

### LINKED ISSUE #198: Create Database Schema and Types
## Feature Request
Create the database schema and the TS types for annotated collections

## Implementation Plan

conservationtimothy:
## Goal

Closes #198. This PR is the first PR in the attempt to cherry pick, staggered PR #191. It adds the database schema to `warehouse.sql` and adds TS types.

## Screenshots
N/A

## What I changed
- Adds database schema to warehouse.sql
- Adds TS types representation of Annotated Collections

### DIFF
--- a/db/init/warehouse.sql
+++ b/db/init/warehouse.sql
@@ -739,7 +739,6 @@ CREATE TABLE public.annotated_collections (
     name text NOT NULL,
     description text,
     collection_type text NOT NULL,
-    status text DEFAULT 'active'::text,
     created_by text NOT NULL,
     created_at timestamp with time zone DEFAULT now(),
     updated_at timestamp with time zone DEFAULT now(),
@@ -755,6 +754,7 @@ CREATE TABLE public.incidents (
     collection_id uuid NOT NULL,
     incident_type text,
     responsible_party text,
+    status text DEFAULT 'suspected'::text,
     impact_description text,
     supporting_evidence jsonb
 );
--- a/db/init/warehouse.sql
+++ b/db/init/warehouse.sql
@@ -727,6 +727,119 @@ ALTER TABLE ONLY public.seed_survey_data
 
 
 --
+-- Annotated Collections Tables
+--
+
+--
+-- Name: annotated_collections; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.annotated_collections (
+    id uuid DEFAULT gen_random_uuid() NOT NULL,
+    name text NOT NULL,
+    description text,
+    collection_type text NOT NULL,
+    status text DEFAULT 'active'::text,
+    created_by text NOT NULL,
+    created_at timestamp with time zone DEFAULT now(),
+    updated_at timestamp with time zone DEFAULT now(),
+    is_active boolean DEFAULT false,
+    metadata jsonb DEFAULT '{}'::jsonb
+);
+
+--
+-- Name: incidents; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.incidents (
+    collection_id uuid NOT NULL,
+    incident_type text,
+    responsible_party text,
+    impact_description text,
+    supporting_evidence jsonb
+);
+
+--
+-- Name: collection_entries; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.collection_entries (
+    id uuid DEFAULT gen_random_uuid() NOT NULL,
+    collection_id uuid NOT NULL,
+    source_table text NOT NULL,
+    source_id text NOT NULL,
+    source_data jsonb,
+    added_by text NOT NULL,
+    added_at timestamp with time zone DEFAULT now(),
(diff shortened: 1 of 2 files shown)


types/types.ts
```diff
@@ -171,3 +171,67 @@ export interface ToastOptions {
...
+    | "camera_trap"
+    | "direct_observation"
+    | "track"
+    | "scat"
+    | "call";
+  behavior_notes?: string;
+  habitat_type?: string;
+  weather_conditions?: string;
+  observer_confidence?: "high" | "medium" | "low";
+}
```
rudokemper, line 214: Remove for now. Although we mentioned this use case as an example, we won't be doing annotated collections for wildlife anytime soon and will need to understand the data schema better before adding.

types/types.ts
```diff
@@ -171,3 +171,67 @@ export interface ToastOptions {
     | "bottom-center"
     | "bottom-right";
 }
+
+// Annotated Collections Types
+export interface AnnotatedCollection {
+  id: string;
+  name: string;
+  description?: string;
+  collection_type: "incident" | "wildlife_sighting" | "research_plot";
```
rudokemper, line 180:
```suggestion
  collection_type: "incident";
```

(For now)

types/types.ts
```diff
@@ -171,3 +171,67 @@ export interface ToastOptions {
...
+  status: "active" | "archived" | "draft";
```
rudokemper, line 181: I'm not sure if we should support publication status.

In the case of incident, we have a different use case for "status": `"validated" | "suspected" | "resolved"`.

Maybe we should move `status` to `Incident` instead? Or be explicit and call it `incident_status`.

types/types.ts
```diff
@@ -171,3 +171,67 @@ export interface ToastOptions {
...
+  created_by: string;
+  created_at: string;
+  updated_at: string;
+  is_active: boolean;
```
rudokemper, line 185: Is this used to track whether the incident is "activated" on the front end, by an end user? I reckon not, right? In which case, I'd remove.

types/types.ts
```diff
@@ -171,3 +171,67 @@ export interface ToastOptions {
...
+  metadata: Record<string, string | number | boolean>;
+}
+
+export interface Incident {
+  collection_id: string;
+  incident_type?: string;
```
rudokemper, line 191: I think we want to make this a "select one" field with options like "mining", "logging", "petroleum drilling", "poaching", "land invasions", "wildlife trafficking". Which then raises the typical survey design question: what happens if it's not one of the above? Do we support "other" and then another field `incident_type_other` (String)?

types/types.ts
```diff
@@ -171,3 +171,67 @@ export interface ToastOptions {
...
+  responsible_party?: string;
+  impact_description?: string;
```
rudokemper, line 193: Is this redundant with `description` from `AnnotatedCollection`?

types/types.ts
```diff
@@ -171,3 +171,67 @@ export interface ToastOptions {
...
+  supporting_evidence?: {
+    photos?: string[];
+    documents?: string[];
+    testimonies?: string[];
+  };
```
rudokemper, line 198: Just to clarify -- the vision is that on the front end, the end user will be able to upload files. So I assume these will be used to store filenames?

db/init/warehouse.sql
```diff
@@ -726,6 +726,119 @@ ALTER TABLE ONLY public.seed_survey_data
...
+CREATE TABLE public.collection_entries (
+    id uuid DEFAULT gen_random_uuid() NOT NULL,
+    collection_id uuid NOT NULL,
+    source_table text NOT NULL,
+    source_id text NOT NULL,
+    source_data jsonb,
+    added_by text NOT NULL,
+    added_at timestamp with time zone DEFAULT now(),
+    notes text
+);
```
rudokemper, line 775: See my notes / questions about these fields in `types.ts` below.

rudokemper (changes requested):
Thanks @conservationtimothy! Having this cherry-picked enabled me to dig into the nitty gritty of the schema itself :)

types/types.ts
conservationtimothy: Yes in the future one will ideally be able to upload files.

types/types.ts
conservationtimothy: will move it to Incident bc i think it becomes slippery having type_status on the parent

conservationtimothy:
> Before merging, can we still give some thought to using a development branch for Annotated Collections per @IamJeffG's [suggestion](https://github.com/ConservationMetrics/gc-explorer/pull/191#issuecomment-3361670668)?
> 
> The reason being that we don't always control when a GC Explorer instance rebuilds using `latest` (it happens upon VM or CapRover restart, which can transpire unexpectedly), and I want to avoid anything inadvertently breaking something in production for our users (albeit unlikely).

Yes, i am in thr middle of doing this actually.

PR #213, opened 2025-10-08
Bug: Public datasets still asking for login & Bug: View Routes do not resolve without auth0 role

### LINKED ISSUE #189: Bug: view routes do not resolve for user without an auth0 Role set
Today I partnered with one of our partners in debugging an issue that they were having.

The issue raised when they were attempting to access one of their Explorer views. And then we noticed that it didn't happen with only one view, it happened with all Explorer views.

In order to manifest the issue, you had to have an account that is authorized in Auth0 but has no assigned roles. With the views configured in Explorer without any role-based access control configuration. So the defaults would apply. When doing that, what is expected is for the user to be able to access Explorer views. But what happened was that not only they were not able to see anything, even they were able to see the links to the views in the screen. 

But also it was a silent error. No logging issue, no, no, no issue was logged in the frontend and no issue was raised in the backend. Later we found out that if we assigned a role to the user, the issue disappeared. So we noticed that the default was not being set up. 

What we noticed at the end of the debugging session, now with @rudokemper , was that there was a log on the backend saying that the backend would apply default roles (or something like that)... but default roles did not get applied apparently, as no changes were seen in Auth0.

### LINKED ISSUE #210: Bug: public datasets still asking for login
#### Steps to Reproduce

1. Set a view to Visibility: "Public -- no sign-in-required"
2. Try to open the link to that view in an incognito browser; you will still be asked to sign in
3. OR, as a user with Public role, try to click on the link in the index; you will see an "Access Denied" toast

conservationtimothy:
## Goal
Fix the bug where public (i.e. dataset with `anyone` permissions) datasets were asking for log in. Closes #210 and a bug around view routes not rendering because the permission requirements were too high Closes #189. 

## What I changed
- Default permissions is `signed-in`
- Role already defaulted to "Viewer"
- Fix type casting which was using outdated variable i.e. `routeLevelPermission` -> `ROUTE_LEVEL_PERMISSION`

### DIFF
--- a/components/config/ConfigPermissions.vue
+++ b/components/config/ConfigPermissions.vue
@@ -23,7 +23,7 @@ const routeLevelPermission = ref<RouteLevelPermission | undefined>(
 const shouldShowPermissions = computed(() => {
   if (!user.value) return false;
   const typedUser = user.value as User;
-  const userRole = typedUser.userRole || Role.Viewer;
+  const userRole = typedUser.userRole ?? Role.Public;
   return userRole >= Role.Admin;
 });
 
--- a/middleware/oauth.global.ts
+++ b/middleware/oauth.global.ts
@@ -39,7 +39,7 @@ export default defineNuxtRouteMiddleware(async (to) => {
 
       const response = await $fetch<
         [
-          Record<string, { routeLevelPermission?: RouteLevelPermission }>,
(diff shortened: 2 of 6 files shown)


rudokemper (approved):
Thanks Osa for taking action so quickly. I figured it was something silly :)

Issue #218, opened 2025-10-14
Update Auth0 Roles Screenshots and Remove stale roles


conservationtimothy:
## Feature Request
We need to update the roles in Auth0 for our partners, remove redundant roles like "Public" and update the JSON config (e.g. changed `signed_in` to `guest` (if we decide that is the correct nomenclature)). Depends on #211 being closed

## Implementation Plan
(may be added later, before work begins)

PR #220, opened 2025-10-14
Add condition to handle when both tables exist

conservationtimothy:
## Goal
Add condition to handle when both tables exist and migrate data from config to view_config and drops the old table

## LLM use disclosure
None

### DIFF
--- a/server/db/migrations/0001_rename_config_to_view_config.sql
+++ b/server/db/migrations/0001_rename_config_to_view_config.sql
@@ -1,11 +1,18 @@
 -- Migration: Rename config table to view_config
 -- This migration renames the existing config table to view_config
 
--- Only rename if config table exists and view_config doesn't exist
 DO $$
 BEGIN
+    -- If config table exists and view_config doesn't exist, rename it
     IF EXISTS (SELECT 1 FROM information_schema.tables WHERE table_name = 'config')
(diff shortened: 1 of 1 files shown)


rudokemper (commented):
Thanks for doing this so quickly.

Part of me wonders if we should just resolve this manually, by SSHing into the ~10 servers and running

```sql
ALTER TABLE guardianconnector.config
RENAME TO view_config;
```

Since we have to manually modify some of the JSON content of `view_config` anyway per https://github.com/ConservationMetrics/gc-explorer/issues/218.

And then we don't need this migration at all.

Thoughts?

rudokemper (changes requested):
Since we decided to this manually (and I can do this, documented in https://github.com/ConservationMetrics/gc-explorer/issues/218), can you update this PR to just remove the relevant migrations altogether, since they will be superfluous?

conservationtimothy:
> Since we decided to this manually (and I can do this, documented in #218), can you update this PR to just remove the relevant migrations altogether, since they will be superfluous?

I vote we keep it in the migration history. It’s harmless, idempotent in practice, and serves as a record of our schema evolution; it was the first ever Drizzle migration. We can just close this PR. @rudokemper wdyt?

Issue #100, opened 2025-05-06
Allow multiple custom basemaps for Basemap Selector


rudokemper:
## Feature Request

Currently, the GuardianConnector Basemap Selector (as used by Map and Alerts views) supports adding only a single custom basemap (e.g. the default custom Mapbox style) in addition to the predefined options like "satellite" and "streets." To better support diverse user needs, we should update the component to support an arbitrary number of custom basemaps.

Assumptions we will make in doing this:

1. We are dealing only with Mapbox maps.
2. One Mapbox access token is sufficient to access all of the map styles (e.g., all of the map styles are on the same account).
3. Other map properties like zoom, center, pitch, bearing, projection, 3d will not change.

### User Stories

A community organization wants to explore their field data overlaid against several different styled maps; one with zonal demarcations, another with territory data and satellite imagery, and yet another with tiled raster imagery.

## Implementation Plan

* Modify the `BasemapSelector` component logic to accept an array of custom basemaps instead of a single one.
* Update the UI to list all available custom basemaps alongside the standard ones. Ensure each custom basemap can be uniquely identified by name.
* Update the `ConfigMap` component to allow for adding multiple Mapbox styles.
* Update the API routes and types.
* Add unit tests to cover the enhancement.

PR #226, opened 2025-10-30
Add ability to set (and toggle between) multiple basemaps for a map view 

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-explorer/issues/100.

## Screenshots

image

_(ignore the LastPass red-box-with-three-dots icons, that's from a browser extension...)_

image

## What I changed

* Deprecated `MAPBOX_STYLE` (but left in for backwards compatibility) in favor of `MAPBOX_BASEMAPS`, a JSON string of a new interface `BasemapConfig[]` consisting of `name`, `style`, and `isDefault`
* In `ConfigMaps`, created a user experience for basemaps management, allowing for adding up to 3 basemaps, each consisting of a name and style. The values are stored in `MAPBOX_BASEMAPS`
  * The maps can be drag-and-dropped in order, and the first one is always the default and cannot be removed. 
  * There are a bunch of validations checking for name uniqueness and validity of the fields. 
  * For current views configs where a `MAPBOX_STYLE` is set, that style is populated as the first entry in `MAPBOX_BASEMAPS`. 
* In the API, applied a new util function `parseBasemaps()` which parses `MAPBOX_BASEMAPS` into a `basemaps: BasemapConfig[]` object, falling back to legacy `MAPBOX_STYLE` for backwards compat.
* Basemap choices are added in `BasemapSelector` component, here also with fallback handling if `MAPBOX_BASEMAPS` is not yet used.
* Bugfix to apply 3d Terrain whenever a new style loads, using a new util function `applyTerrain()`. Previously, 3D terrain only loaded for the first map style.
* Added unit and e2e tests for new behavior.

## LLM use disclosure

* Used Claude Sonnet 4.5 in Cursor to help generate unit and e2e tests, and handle localization.


conservationtimothy:
This is really cool, I learnt something new today

PR #228, opened 2025-11-10
Determine why GC Explorer Docker image size has increased by ~4x, and resolve

### LINKED ISSUE #222: Determine why GC Explorer Docker image size has increased by ~4x, and resolve

**Rudo:**

> At some point 4 months ago, the image size of GC Explorer shot up by ~4x. I've been noticing this because when trying to upgrade GC Explorer on the VMs via CapRover, it sometimes fails due to lack of space. When I SSH into the VM to see what's going on and run `docker image ls`, I notice that the size of GC Explorer is 2.89gb, as opposed to GC Landing Page (also Nuxt 3) which is 342mb. I recall GC Explorer taking up about ~550mb on the VMs in the past.

Image

> We build GC Explorer with `node:20.15.0-slim` whereas Landing Page uses a smaller size distro for production, but that can't be the cause for the _increase_ as we've been using that Node distro for much longer than 4 months.

**Jeff:**

It seems to tell us that the increase was introduced in PR 129:
https://github.com/ConservationMetrics/gc-explorer/pull/129/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

@conservationtimothy  can all the testing libs become devDependencies?  And/or is the production build not tree-shaking them out?

conservationtimothy:
## Goal

Reduce GC Explorer Docker image size from 2.89GB to ~714mb to prevent deployment failures due to insufficient disk space on VMs. Closes #222 

## Screenshots

N/A (infrastructure change)

## What I changed

- Converted Dockerfile to multi-stage build (builder + production stages)
- Updated `package.json` to use conditional postinstall script that skips Playwright browser installation in production
- Enhanced `.dockerignore` to exclude test files, docs, and development-only files
- Optimized production stage to install only production dependencies (plus drizzle-kit for migrations)

## What I'm not doing here

- Not switching to distroless base image
- Not further optimizing node_modules size

## LLM use disclosure

Used LLM (Cursor/Auto) for consultation on Docker optimization strategies, multi-stage build patterns, and pnpm dependency management.


conservationtimothy:
we've dropped image to 1.16gb btw
Screenshot 2025-11-10 at 13 36 06

conservationtimothy:
> That's good! And I wonder if we can diagnose a bit more? Like running `du -sh node_modules/* | sort -h` on a container running that image to see if there are any obvious offenders. I hope we can still cut that down by half or so, to return to what it was? (Don't think much was added to the app that wasn't a dev dependency...)

We absolutely can, this is a first iteration for me, I will keep trying, my ideal goal is maybe sub 1GB. Fwiw it is a decently sized app. I am also open to ideas on diagnosing.

conservationtimothy:
cc: @IamJeffG , this is WIP but open to thoughts and ideas :)

assets/icons/warning_red.png
IamJeffG: this and warning_orange.png can be removed?

conservationtimothy:
> I highly recommend using https://github.com/wagoodman/dive to see where space is being used in a Docker image: that's both which "layers of history" and which folders on disk.
> 
> The big question I have is: are vitest and similar devDepenendencies included in the production image or no? I feel they ought not be. And we could optionally build a separate image for local development / testing and/or mount a local `node_modules` to the production image.

This PR should exclude Playwright in production, in the same vein we could exclude vitest.

rudokemper:
Building _this_ branch locally:

```
REPOSITORY         TAG                                   IMAGE ID       CREATED          SIZE
gc-explorer        feat-investigate-and-fix-image-size   753dc3e2d4e9   10 seconds ago   713MB
```

IamJeffG (approved):
I wrote a couple comments, but you've already achieved your goal and so actually I recommend to ignore my comments. 😃 


great work

PR #260, opened 2025-12-17
Handle case where there are no alerts in the table

rudokemper:
## Goal

For the alerts flow in Guardian Connector, there is a possible state where an alerts table is created, but no alerts are posted (yet). This is a **desirable** state because we _are_ running an alerts schedule in Windmill to fetch alerts, and there could be metadata indicating an alert detection range. So this PR makes that possible; before, GC Explorer would just loudly fail upon no `alertsData` being received.

Closes https://github.com/ConservationMetrics/gc-programs/issues/130

## Screenshots

image

## What I changed and why

In `prepareAlertsStatistics()`:

* Turned operations to get data providers and date range into helper functions for reusability
* Returned data for when the alerts table is empty, but using metadata table if available
* Only show slider when data is available

## LLM use disclosure

Done through prompt engineering with Cursor to win time.

### DIFF
--- a/server/dataProcessing/transformData.ts
+++ b/server/dataProcessing/transformData.ts
@@ -404,35 +404,56 @@ const prepareAlertsStatistics = (
   data: DataEntry[],
   metadata: AlertsMetadata[] | null,
 ): AlertsStatistics => {
+  // Extract data providers from metadata (reusable for empty data case)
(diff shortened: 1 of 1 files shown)


nicopace (approved):
looks ok :)

Issue #208, opened 2025-10-07
Write tests for new annotated collections / incidents collection


conservationtimothy:
## Feature Request
We need to (likely in the UI) write tests to be sure this new functionality is working as expected. So for this with our simple UI, we will right some Playwright and probably also component tests

## Implementation Plan
(may be added later, before work begins)

PR #262, opened 2025-12-30
Build Simple UI to be able to visually test new API for Incidents 

### LINKED ISSUE #190: Build Simple UI to be able to visually test new API for Incidents
## Feature Request
We have Figma designs but in the mean-time, I believe we should build a simple UI that A. lays the groundwork, consists of some of the features of "Incidents" e.g. Bounding Box, Multi-select B. Allows us to be able to visually create incidents as opposed to using the API. 

It should do a lot of the heavy logical lifting where all we need to do is change UI in certain places when the time comes for implementing the UI overhaul. This way the lift is not massive. To achieve this i have a `sketch` branch derived off the parent API changes branch. I think this model makes sense.

## Implementation Plan

conservationtimothy:
## Goal

Add a simple UI to test the annotated collections/incidents API endpoints in the Alerts Dashboard. Closes #190 

## Screenshots

Screenshot 2025-12-30 at 17 27 29
Screenshot 2025-12-30 at 16 38 42
Ctrl + click i.e. Multi select 👆🏿
Screenshot 2025-12-31 at 12 49 00
Screenshot 2025-12-31 at 12 47 35
Screenshot 2025-12-31 at 12 47 29

Bounding box 👆🏿

## What I changed and why

Added a simple UI to AlertsDashboard.vue for testing the annotated collections/incidents API. Includes three control buttons (View Incidents, Multi-select, Bounding Box) at 50% screen height, an IncidentsSidebar component for viewing/creating incidents, and multi-select (Ctrl+click) and bounding box which is WIP (Ctrl (Command)+drag) functionality to add map features to collections.

Also Merged main into feat/annotated-collections-feature-branch and ran database migrations to ensure tables exist.

## What I'm not doing here

This is a minimal testing UI so stuff can be more comprehensive 
## LLM use disclosure

Used LLM (Auto/Cursor) for code generation of the incidents UI components minor things like placing 50% of the screen as well as bounding box research


conservationtimothy:
> Note: this is WIP as i am not through with the bounding box implementation, right now it does not do what I want it to do. It draws the box but it zooms to the nearest element instead of selecting it. However it is mostly done where there is some merit in reviewing if you wish.

Ok this is ready : )

components/AlertsDashboard.vue
```diff
@@ -1973,8 +2090,454 @@ const calculateLineStringCentroid = (coordinates: number[][]) => {
...
+        });
+        const validFeatures = features.filter(
+          (f) =>
+            !f.properties?.cluster &&
+            f.properties?.cluster_id === undefined &&
+            !f.layer?.id?.includes("clusters") &&
+            !f.layer?.id?.includes("cluster-count"),
+        );
+        allFeatures.push(...validFeatures);
+        console.log(`Layer ${layerId}: found ${validFeatures.length} features`);
```
IamJeffG: I don't know if we've established team norms for logging, but at the very least I suggest replacing these `console.log` with a specific log level (likely `console.debug` in most cases)

components/AlertsDashboard.vue
```diff
@@ -1973,8 +2090,454 @@ const calculateLineStringCentroid = (coordinates: number[][]) => {
...
+    const pos = `translate(${minX}px, ${minY}px)`;
+    box.style.transform = pos;
+    box.style.width = maxX - minX + "px";
+    box.style.height = maxY - minY + "px";
+  };
+
+  /**
+   * Handles mouse up to finish selection
+   */
+  const onMouseUp = (e: MouseEvent) => {
```
IamJeffG: Functions like `onMouseUp`, `finish` are actually very specific to bounding box selection for incidents, not generic to all of the AlertsDashboard.

At the very least I'd suggest making the function names a bit more precise: e.g. `onMouseUpDuringBoundingBox` `finishBoundingBox`

However a better recommendation may be to try to split a lot of this out to 1 or more new component(s).  This file is almost 3000 lines long and really has a runaway scope.  I think you could fairly easily extract this PR's new stuff to one or more new components, and perhaps another time later we come back and also look at if we should break apart any of the code that was here prior.

components/AlertsDashboard.vue
```diff
@@ -2025,6 +2588,113 @@ onBeforeUnmount(() => {
...
+            stroke-width="2"
+          >
+            <path d="M9 12l2 2 4-4m6 2a9 9 0 11-18 0 9 9 0 0118 0z" />
+          </svg>
+        </button>
+        <div v-if="hoveredButton === 'incidents'" class="tooltip tooltip-left">
+          {{
+            showIncidentsSidebar
+              ? "Hide incidents sidebar"
+              : "View saved incidents and create new ones"
```
IamJeffG: I'm ok leaving i18n out of this PR. 👍

IamJeffG (changes requested):
Generally very good code.  I am very concerned about the length of `AlertsDashboard.vue` but would not block merge solely on that.

I am marking "request changes" but what I actually want is more context from you to help me review `server/database/annotatedCollections.ts`.   I did not really review that file, am not clear what's in-scope or what's going on there.

components/AlertsDashboard.vue
```diff
@@ -1973,8 +2090,454 @@ const calculateLineStringCentroid = (coordinates: number[][]) => {
...
+
+  if (feature.properties.alertID) {
+    sourceId = feature.properties.alertID;
+  } else if (feature.properties.id) {
+    sourceId = feature.properties.id;
+  }
+
+  if (sourceTable && sourceId) {
+    addSourceToSelection(sourceTable, sourceId);
+    highlightSelectedSource(feature, layerId);
```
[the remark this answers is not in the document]
conservationtimothy: good catch

components/AlertsDashboard.vue
```diff
@@ -1973,8 +2090,454 @@ const calculateLineStringCentroid = (coordinates: number[][]) => {
...
+ * Clears all selected sources
+ */
+const clearSelectedSources = () => {
+  selectedSources.value = [];
+  clearSourceHighlighting();
+};
+
+/**
+ * Handles multi-select feature selection
+ * Determines source table from layer ID and route params, extracts source ID from feature properties
```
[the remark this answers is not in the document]
conservationtimothy: +1 on documenting will do. On tests, there is an issue: https://github.com/ConservationMetrics/gc-explorer/issues/208 I plan to attend to once this is merged.

components/shared/IncidentsSidebar.vue
```diff
@@ -0,0 +1,538 @@
+<script setup lang="ts">
```
rudokemper: Unless I'm wrong, `IncidentsSidebar` will only be used in `AlertsDashboard`. So it is not a shared component. I would move this under `alerts/`

rudokemper:
Great work, Osa!

Some feedback on the UI, not code review:

---

When I am in the process of selecting records to be part of a new incident, the sidebar UI is a little confusing. I think that this state should no longer show "Saved Incidents". Maybe it is enough to remove that, and keep the button to add a new one, like so:

image

---

I think there should be a "Clear" control that deselects all of the records that I have selected until now. I didn't see a way to deselect anything outside of the "Clear All" button in the Incident sidebar - I intuited that I might be able to by ctrl-clicking, or drawing a bounding box over, already selected features, but that doesn't work.

---

I didn't see a way to "activate", open, or otherwise interact with existing saved incidents. Just making sure that this was out of scope for this PR, but forthcoming in another one?

PR #276, opened 2026-01-13
Writes tests for new annotated collections / incidents collection

### LINKED ISSUE #208: Write tests for new annotated collections / incidents collection

conservationtimothy:
## Goal
Write tests for the new annotated collections / incidents collection flow.  Closes #208 

## Screenshots
Screenshot 2026-01-14 at 16 58 16
Screenshot 2026-01-13 at 23 02 09

To check tests pass, I switched to main and ran unit and E2E tests then switched back to feature branch.

## What I changed and why
Add relevant test IDs for latching onto stuff,

## What I'm not doing here
Any additional logic. 

## LLM use disclosure
Used this to test out the Zed, OpenCode, OpenCode Zen workflow. specifically using 4.5 Opus and GPT 5.2

### DIFF
--- a/e2e/07-annotated-collections.spec.ts
+++ b/e2e/07-annotated-collections.spec.ts
@@ -461,12 +461,6 @@ test("annotated collections - remove individual source from selection", async ({
     return;
   }
 
-  const clickPoint = await projectLngLatToPagePoint(page, lngLat);
-  if (!clickPoint) {
-    test.skip();
-    return;
-  }
-
   // Select a feature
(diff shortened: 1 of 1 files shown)


rudokemper (approved):
For the record, I didn't review this very closely, but am trusting that the behavior is fully tested and happy if the tests pass and are not false positives.

PR #298, opened 2026-02-02
Fix incident link copying to exclude alertId parameter

### LINKED ISSUE #295: Fix incident link copying to exclude alertId parameter
## Feature Request

When copying an incident link, the system should exclude any `alertId` or `mapeoDocId` parameters from the URL to prevent confusion. Currently, if a user has both an incident and an alert selected, copying the incident link includes both parameters, causing the link to open the alert view instead of the incident view when shared.

**User Story:**
As a user copying an incident link, I want the copied link to only include the incident ID and not merge with any selected alert ID, so that when I share the link, it opens directly to the incident view without showing an unrelated alert.

**Current Behavior:**
When copying an incident link while an alert is also selected, the copied URL includes both `incidentId` and `alertId` parameters (e.g., `?incidentId=xxx&alertId=yyy`). When opening this link, it shows the alert instead of the incident, causing confusion.

**Expected Behavior:**
When copying an incident link, the URL should only include the `incidentId` parameter, excluding any `alertId` parameter that might be present in the current URL.

**Acceptance Criteria:**
- [ ] Copying an incident link only includes `incidentId` in the URL
- [ ] Any `alertId` parameter is excluded from the copied incident link
- [ ] Opening a copied incident link shows the incident view, not an alert view
- [ ] The behavior is consistent whether an alert is selected or not

## UI Sketch (or Screenshot)

N/A

conservationtimothy:
## Goal
This PR removes ensures an incident link does not copy over alertId param. Closes #295. 

## What I changed and why
Accepts a list of params to exclude from the URL and excludes them when copying over a URL.

## LLM use disclosure
None

### DIFF
--- a/composables/useFeatureSelection.ts
+++ b/composables/useFeatureSelection.ts
@@ -287,13 +287,11 @@ export function useFeatureSelection(
       ? featureObject.alertID
       : feature.id;
 
-    // Update URL with alertId or mapeoDocId
+    // Update URL with alertId or mapeoDocId; remove incidentId so address bar matches "copy link to alert"
     const query = { ...route.query };
-    // Remove any existing feature IDs first
     delete query.alertId;
     delete query.mapeoDocId;
-
-    // Add the new feature ID
+    delete query.incidentId;
     if (featureObject.alertID) {
       query.alertId = featureObject.alertID;
       isMapeo.value = false;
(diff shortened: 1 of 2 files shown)


conservationtimothy:
I have implemented the request, thank y'all for the feedback.

PR #306, opened 2026-02-09
Add Minimal Map Data Endpoint

### LINKED ISSUE #267: Add minimal map data endpoint (`GET api/[table]/map`)
## Feature Request

Create a purpose-specific Map data endpoint to replace the current “fetch everything” pattern for map views. This endpoint must return only the minimum fields required to render the map (geometry + stable identifier + styling/filter fields).

* **As a user viewing a dataset map**, I want the map to load quickly even for large datasets, so that the UI feels responsive.
  * **AC:** Map view loads using `GET api/[table]/map` and does **not** fetch full records by default.
  * **AC:** Response includes: `_id`, `g__type`, `g__coordinates`, and if set in config, the values of `COLOR_COLUMN` and `ICON_COLUMN`.
  * **AC:** Endpoint returns valid GeoJSON FeatureCollection.
  * **AC:** Payload size and server processing time are measurably smaller than the current endpoint for the same dataset.

## Implementation Plan

conservationtimothy:
**Goal**  
Implement a purpose-built map endpoint (issue 267): `GET api/[table]/map` returns only the data needed for the map (geometry, id, styling/filter fields) as a GeoJSON FeatureCollection instead of loading and transforming full records. Closes #267 

**Screenshots**  
Screenshot 2026-02-09 at 10 42 31
Screenshot 2026-02-09 at 10 40 20
Screenshot 2026-02-09 at 10 40 04

**What I changed and why**  
The map view used a single endpoint that did `SELECT *`, then filtering, survey transforms, and `prepareMapData`, and returned a big array. That didn’t scale and mixed data access with presentation. This PR adds `fetchMapData` in `dbOperations` to run a minimal SELECT (`_id`, `g__type`, `g__coordinates`, plus optional COLOR/ICON/FILTER columns from config). The map handler now uses that, applies only geo validation and value-based filtering, and returns a GeoJSON FeatureCollection as `response.data`. MapView accepts either the legacy array or a FeatureCollection and normalizes to the same internal shape so DataFilter and the map layers keep working. Result: smaller payloads, less server work, and a clear path to on-demand full records (268) and client-side transforms (269).

**What I'm not doing here**  
Spatial/viewport filtering, the single-record endpoint (268), or moving transforms to the client (269). Those stay in follow-up PRs.

**LLM use disclosure**  
Cursor with me driving


server/api/[table]/map.ts
```diff
@@ -1,19 +1,140 @@
...
+ * Builds a GeoJSON Feature from a minimal map row (raw DB columns).
+ * Assigns filter-color from the filter column value for styling and DataFilter.
+ *
+ * @param {Record<string, unknown>} row - One row from fetchMapData (e.g. _id, g__type, g__coordinates, optional styling columns).
+ * @param {ViewConfig} config - View config for this table (used for property keys the client expects).
+ * @param {{ colorColumn: string | null; iconColumn: string | null; filterColumn: string | null }} resolvedColumns - Resolved SQL column names.
+ * @param {Map<string, string>} filterColorMap - Map from filter value to hex color (populated as we iterate).
+ * @returns {Feature | null} A GeoJSON Feature, or null if geometry is invalid.
+ */
+function rowToFeature(
```
Comment (rudokemper) [inline server/api/[table]/map.ts]: Remember the repo-wide convention (here and `gc-landing-page`) to use arrow functions. Maybe add a custom instruction for Cursor.

rudokemper (changes requested):
On the right track. Some minor requested changes.

---

The biggest issue that I see with this code is something that I missed in the AC of https://github.com/ConservationMetrics/gc-explorer/issues/267:

The same logic will need to be applied to the alerts dashboard, e.g. the preparation of spatial data for `api/[table]/alerts.ts`. Both Mapeo and alerts data. This is actually where the biggest bottleneck is, as the alerts datasets tend to grow quite large. You can see this a bit in the payload for the `alerts/gfw_alerts_viirs` view on Demo.

So I think a task ahead of us is to take a lot of the "spatial payload shaping" code that you added to `api/[table]/map.ts` and make it a shared module, so that it can be called by multiple view endpoints. 

(The same applies to any code added to `MapView.vue` which might be used by `AlertsDashboard.vue`)

I think it'd be fine if you kept this PR as-is, and did the refactor in a subsequent PR. But either this PR doesn't fully close https://github.com/ConservationMetrics/gc-explorer/issues/267, or you need to add another sub-issue to document the need for shared modular code used by multiple endpoints + components, and implementation for the alerts dashboard.

---

I wish we had a large enough dataset of mapping data to tell a story about the before & after payload delta! (Well, we will once we call the endpoint via the alerts dashboard.

conservationtimothy:
> On the right track. Some minor requested changes.
> 
> The biggest issue that I see with this code is something that I missed in the AC of #267:
> 
> The same logic will need to be applied to the alerts dashboard, e.g. the preparation of spatial data for `api/[table]/alerts.ts`. Both Mapeo and alerts data. This is actually where the biggest bottleneck is, as the alerts datasets tend to grow quite large. You can see this a bit in the payload for the `alerts/gfw_alerts_viirs` view on Demo.
> 
> So I think a task ahead of us is to take a lot of the "spatial payload shaping" code that you added to `api/[table]/map.ts` and make it a shared module, so that it can be called by multiple view endpoints.
> 
> (The same applies to any code added to `MapView.vue` which might be used by `AlertsDashboard.vue`)
> 
> I think it'd be fine if you kept this PR as-is, and did the refactor in a subsequent PR. But either this PR doesn't fully close #267, or you need to add another sub-issue to document the need for shared modular code used by multiple endpoints + components, and implementation for the alerts dashboard.
> 
> I wish we had a large enough dataset of mapping data to tell a story about the before & after payload delta! (Well, we will once we call the endpoint via the alerts dashboard.

I think I am OK doing it in this 306

rudokemper (changes requested):

components/AlertsDashboard.vue
```diff
@@ -874,13 +874,13 @@ const addAlertsData = async () => {
 
 /**
  * Adds (optional) Mapeo data to the map by creating a GeoJSON source and a layer for Point features.
- * It also sets up event listeners for user interactions with the Mapeo data features.
+ * Expects mapeoData as Dataset (array with geotype, geocoordinates); if the API later returns
+ * a FeatureCollection, we could pass it through directly like MapView to avoid conversion.
  */
```
rudokemper: I don't understand. The current behavior in this PR:

1. In `server/api/[table]/alerts.ts` constructs a Mapeo FeatureCollection
2. It then converts it back "to mapeoData shape expected by AlertsDashboard." (Dataset type)
3. Then, here in `components/AlertsDashboard`, we receive the Dataset, and in `addMapeoData`, keep the old pattern of constructing a FeatureCollection

Why this redundancy for the alerts dashboard, yet for the map view, we streamline this and just have everything work with FeatureCollections? This is what I meant in https://github.com/ConservationMetrics/gc-explorer/pull/306#pullrequestreview-3774839178.

---

Additionally, for the alerts dashboard, we should also do a similar refactor for **AlertsData** -- that is, prepare a minimal FeatureCollection for the alerts features, and then leverage a single-record endpoint to get the record for `DataFeature` when the sidebar is opened. However, since this PR is already quite large, let's make a new issue to encapsulate that work.

components/MapView.vue
```diff
@@ -45,14 +46,76 @@ const props = defineProps<{
...
   mediaBasePath?: string;
   mediaBasePathIcons?: string;
   mediaColumn?: string;
   planetApiKey?: string;
 }>();
 
-const filteredData = ref([...props.mapData]);
+/**
+ * Converts mapData (FeatureCollection) to a Dataset-shaped array for DataFilter and ViewSidebar.
+ * Fails loudly if mapData is not a valid FeatureCollection.
```
rudokemper:
```suggestion
```

You don't have to add my code review notes as literal comments to the codebase. This behavior here should be obvious from just reading the code.

components/MapView.vue
```diff
@@ -142,17 +205,16 @@ const addDataToMap = () => {
     }
   }
 
-  // Create a GeoJSON source with all the features
+  // Use GeoJSON features directly; add .feature for sidebar/layer expressions
```
rudokemper: This is a LLM pattern I think we should avoid at all costs - when the LLM revises a comment in a way where the new copy is an update on what was there before. But if you don't know what was there before, if you lack the original context, how would you know what this means?

"Use GeoJSON features directly" -- as opposed to what? 

Ah -- before, it said "Create a GeoJSON source with all the features": only now does it make sense. But I had to retrieve that original context from `git blame` or commit history.

@conservationtimothy, as you are heavy on LLM usage, I am going to ask you to please manually review each and every comment revision that Cursor generates for you, and evaluate for yourself if it makes sense, before it gets to me.

components/MapView.vue
```diff
@@ -340,10 +402,9 @@ const addDataToMap = () => {
 const loadIconImages = async () => {
   if (!props.iconColumn || !props.mediaBasePathIcons || !map.value) return;
 
-  // Get unique icon filenames from data
   const iconFilenames = new Set<string>();
-  filteredData.value.forEach((item) => {
-    const iconFilename = item[props.iconColumn!];
+  displayedFeatures.value.forEach((f) => {
+    const iconFilename = (f.properties ?? {})[props.iconColumn!];
```
rudokemper: I notice that in this PR, you are using `v` for value, `f` for feature, `fc` for FeatureClass, etc.

This is a new pattern that you are introducing. I would prefer us to be consistent across the codebase. It adds to the mental load of the code reader to be using multiple patterns across the code, let alone one file alone.

server/api/[table]/alerts.ts
```diff
@@ -91,52 +90,66 @@ export default defineEventHandler(async (event: H3Event) => {
...
-      // Fetch Mapeo data
-      const rawMapeoData = await fetchData(mapeoTable);
-
-      // Filter data to remove unwanted columns and substrings
-      const filteredMapeoData = filterUnwantedKeys(
-        rawMapeoData.mainData,
-        rawMapeoData.columnsData,
-        viewsConfig[table].UNWANTED_COLUMNS,
-        viewsConfig[table].UNWANTED_SUBSTRINGS,
+      // Always use minimal fetch + shared spatial payload (config optional when Mapeo table has no view config).
```
Comment (rudokemper) [inline server/api/[table]/alerts.ts]: Here is another example of an LLM comment that doesn't make a lot of sense out of context.

Read this out loud to yourself and see if it makes sense.

"Always use minimal fetch + shared spatial payload"

As opposed to what? You have to know the whole history of the codebase to know that we did something else before.

rudokemper (changes requested):

PR #308, opened 2026-02-09
Move Filtering and Presentation Transformations to the Client

### LINKED ISSUE #269: Move filtering and presentation transformations to the client
## Feature Request
Move **filtering and human-readable transformations** out of the API and into the client, so the API serves canonical raw data and presentation logic is decoupled. This reduces backend coupling and supports raw exports and alternative consumers. This means that the business logic functions in `server/dataProcessing/transformData` will be called by the front end on an as-needed basis per record. (Hence, the module file should also move outside the `server/` directory.)

* **As a user filtering or viewing formatted fields**, I want consistent results without forcing the backend to pre-transform everything.
  * **AC:** API endpoints return raw values (no presentation formatting).
  * **AC:** Client applies filtering based on config params and transforms values/columns for display.
  * **AC:** There is a clear decision/implementation for transformation timing: render-time vs cached transformed results (documented in code/comments).
  * **AC:** “Download raw data” behavior is no longer blocked by API transformations.

## Implementation Plan

conservationtimothy:
## Goal
Implement subissue 269, move filtering and human-readable transformations out of the API and into the client so the API serves canonical raw data and presentation is decoupled. Closes #269 

## Screenshots
N/A – no UI changes.

**What I changed and why**

Moved survey transformation logic out of the server and into shared client code: `transformData` and `helpers` now live under `utils/dataProcessing/`, with the server re-exporting them. The API returns raw data; the client runs `transformSurveyData` for display. On the map, we still fetch a single record by ID; that raw response is transformed once for the sidebar and the same raw payload is used for “download raw.” The gallery API now returns raw rows; the gallery page transforms them with `transformSurveyData` before passing data to `GalleryView`/`DataFeature`. Added JSDoc to the shared helpers and a test that the utils and server versions of `transformSurveyData` match. Documented on `DataFeature` that the `feature` prop is expected to be client-transformed for display.

## LLM disclosure
Cursor with me driving


rudokemper (changes requested):
I don't follow why we kept `server/dataProcessing/helpers.ts` and `server/dataProcessing/transformData.ts`, and why we do the parity check in tests? It seems vestigial to keep these files around.

---

As with the other PRs, we will need to ensure that filtering / transformation is applied across all views. I am missing implementation for the alerts dashboard.

PR #316, opened 2026-02-12
Make Map Legend Scrollable and Responsive on Small Screens

### LINKED ISSUE #314: Make Map Legend Scrollable and Responsive on Small Screens
## Feature Request

Improve map legend usability on smaller screens to prevent it from obstructing core map interactions.

**Map Legend Responsiveness (QOL)**

* **User Story:** As a user on a smaller screen, I want the map legend to remain accessible without covering the map so that I can continue interacting with the interface comfortably.
* **Acceptance Criteria:**

  * Apply a max-height to the legend container.
  * Make the legend vertically scrollable when content exceeds the available height.
  * Ensure the legend does not override critical map areas specifically the incidents buttons, it should always give 10px minimum from where incidents buttons are.

### UI Sketch (or Screenshot)

*To be added.*

## Implementation Plan

conservationtimothy:
## Goal
Improve map legend usability so it remains accessible without obstructing key map controls or incident interactions across screen sizes. Closes #314.

## What I changed and why
I updated the map legend and incidents controls positioning so the legend stays right-aligned but has a consistently smaller footprint with constrained height and internal vertical scrolling, and I nudged incidents controls upward to reduce overlap risk; this keeps legend access intact on large and small screens while preserving usability of core map/incident controls, which was the core intent of the issue.

## What I'm not doing here
I am not changing feature selection logic, incident creation flow, or geometry handling behavior in this issue.

### DIFF
--- a/components/shared/MapLegend.vue
+++ b/components/shared/MapLegend.vue
@@ -249,6 +249,7 @@ watch(
   min-height: 0;
   overflow-y: auto;
   padding-right: 4px;
+  scrollbar-gutter: stable;
 }
 
 .slide-enter-active,
@@ -275,7 +276,7 @@ watch(
     right: 10px;
     width: min(265px, calc(100vw - 20px));
     max-height: min(32vh, 220px);
-    bottom: 16px;
+    bottom: 44px;
   }
 }
 </style>


conservationtimothy:
> I like your approach to calculate the width, but since the scrollbar is now taking up some of the real estate of the legend, I'd say we increase the width a bit to match the amount of space that was available for the legend content before.
> 
> Image

Screenshot 2026-02-15 at 20 31 10
Screenshot 2026-02-15 at 20 31 49

Pushed these improvements.

conservationtimothy:
> > Pushed these improvements.
> 
> I have not reviewed the code yet, but notes about that second screenshot:
> 
> 1. The map legend component is overlapping / on top of the Mapbox credit. That should not be the case. The same spacing should apply whether or not there is a scrollbar.
> 2. Why not just use the same width as the state where there is no scrollbar (first screenshot)? It doesn't make a lot of sense that if a scrollbar is needed, that the width of the component would actually _decrease_.

The 2nd screenshot is a case of on a small media device width, the first is the case on full sized screen i.e. Edge in full screen mode on a 16" Macbook .

PR #327, opened 2026-02-17
Add Filter-by-viewtype on the homepage

### LINKED ISSUE #321: Add filter-by-view-type on the home page
## Feature Request

The home page (`/`) lists all available dataset views as cards but provides no way to filter them. When the number of datasets grows, users may want to quickly find datasets that have a specific view type enabled (e.g., only datasets with a "map" view, or only those with "alerts").

- **As a** user on the home page, **I want** to filter the displayed dataset cards by view type (map, gallery, alerts), **so that** I can quickly find datasets that offer the view I'm interested in.
- **Acceptance Criteria:**
  - A set of filter buttons (e.g., "All", "Map", "Gallery", "Alerts") is displayed above the dataset cards grid.
  - Clicking a filter shows only datasets that have that view type in their `VIEWS` config.
  - Clicking "All" (or deselecting the filter) shows all datasets.
  - The active filter is visually highlighted.
  - The filter interacts correctly with the existing permission-based filtering.

conservationtimothy:
## Goal

Allow users to filter the home page dataset cards by view type (e.g. Map, Gallery, Alerts) so they can quickly find datasets that offer the view they're interested in. Closes #321 

## Screenshots
Screenshot 2026-02-17 at 08 31 36

## What I changed and why

I added a row of filter buttons ("All", plus one for each view type that exists across visible datasets) above the cards grid. Clicking a button filters the cards to only show datasets that include that view type in their `VIEWS` config. The filter state is synced to the URL via the `?view=` query parameter, so sharing a link like `/?view=map` or refreshing the page preserves the active filter. Selecting "All" removes the param for a clean URL. The buttons are dynamically generated from the data so new view types automatically appear. I also added the `"all"` i18n key to all four locale files.

## What I'm not doing here

Not changing the dataset card component, not adding multi-select filtering (only one filter active at a time).

## LLM use disclosure
Cursor with me driving providing a detailed plan, reading the code and testing in the browser.

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -223,6 +223,7 @@ useHead({
             {{ $t(viewType) }}
           </button>
         </div>
+        <!-- NuxtLink messes up the layout, hence the use of a regular anchor tag -->
         <a
           v-if="shouldShowConfigLink"
           href="/config"
--- a/e2e/03-index.spec.ts
+++ b/e2e/03-index.spec.ts
@@ -72,6 +72,98 @@ test("index page - displays available views and navigation flow", async ({
   expect(hasGallery || hasMap || hasAlerts).toBe(true);
 });
 
+test("index page - view type filter buttons are visible and functional", async ({
+  authenticatedPageAsAdmin: page,
+}) => {
+  await page.goto("/");
+  await page.waitForLoadState("networkidle");
+  await page.waitForSelector("[data-testid='dataset-card']", {
+    timeout: 15000,
+  });
+
+  // "All" button should be visible and active by default
+  const allButton = page.getByRole("button", { name: /^all$/i });
+  await expect(allButton).toBeVisible({ timeout: 5000 });
+
+  // At least one view-type filter button should exist besides "All"
+  const filterButtons = page.locator(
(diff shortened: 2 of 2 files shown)


pages/index.vue
```diff
@@ -139,13 +192,39 @@ useHead({
...
 
-      <!-- Manage Datasets Button - Right above cards on RHS -->
+      <!-- View Type Filter & Manage Datasets -->
       <div
-        v-if="shouldShowConfigLink && viewsConfig"
-        class="flex justify-end mb-4"
+        v-if="viewsConfig"
+        class="flex flex-wrap items-center justify-between gap-3 mb-4"
       >
-        <!-- NuxtLink messes up the layout, hence the use of a regular anchor tag -->
```
rudokemper: This comment was here for a reason. Do not remove it.

conservationtimothy:
> , I think should **test every new behavior**, and that we should make it a regular practice to add unit or e2e tests to every PR that adds new behavior.

More than happy to, I think the question is what constitutes new behaviour *and* where to write tests. Or rather which tests won't be redundant.

Issue #268, opened 2026-01-08
Add single-record raw data endpoint (`GET api/[table]/[recordId]`)


rudokemper:
Create a **single-record raw endpoint** to fetch the canonical record on-demand (i.e., when a user clicks a feature on the map opening `DataFeature` in the sidebar). This decouples view rendering from full-record transport and ensures consumers can access raw data.

### User Stories & Acceptance Criteria

* **As a user clicking a feature on the map**, I want to see full details without the app preloading the entire dataset.
  * **AC:** `GET api/[table]/[recordId]` returns the **raw, untransformed** record exactly as stored (no transformations, we will move those to the client -- see https://github.com/ConservationMetrics/gc-explorer/issues/269).
  * **AC:** Returns 404 for missing records; validates `[recordId]` input.
  * **AC:** Map view uses this endpoint only on demand (click/selection), not on initial page load.
  * **AC:** Endpoint behavior is covered by unit and/or integration tests (including error cases), and all tests pass in CI.

## Implementation Plan
(may be added later, before work begins)

PR #340, opened 2026-02-19
Add minimal map data endpoint (GET api/[table]/map)

### LINKED ISSUE #267: Add minimal map data endpoint (`GET api/[table]/map`)

conservationtimothy:
## Goal

Closes #267 — Create a purpose-specific Map data endpoint that returns a GeoJSON FeatureCollection instead of the previous "fetch everything" flat-array pattern, and extract spatial payload shaping into a shared module used by both the map and alerts endpoints.

## Screenshots

N/A (no visual changes; data format change is transparent to the UI)

## What I changed and why

This PR introduces `buildMinimalFeatureCollection` in `server/utils/spatialPayload.ts` — a shared utility that builds a valid GeoJSON FeatureCollection directly from raw database rows with MurmurHash-based stable IDs, configurable property inclusion, and deterministic filter-color assignment. Both `api/[table]/map.ts` and `api/[table]/alerts.ts` (Mapeo data path) now use this shared module, eliminating the `transformSurveyData` → `prepareMapData` → reconstruct-GeoJSON-in-component pipeline.

On the client side, `MapView.vue` and `AlertsDashboard.vue` now receive FeatureCollections directly, removing the redundant GeoJSON reconstruction that previously happened in `addDataToMap()` and `addMapeoData()`. The map click handler reads properties directly from the clicked GeoJSON feature instead of parsing a nested `feature` property. `MapIntroPanel.vue` and `ViewSidebar.vue` were updated to accept the new `FeatureCollection` prop types. All existing and new behavior is covered by tests (`spatialPayload.test.ts` and updated `MapView.test.ts`), and all 205 tests pass.

## What I'm not doing here

- **Single-record raw endpoint** (`GET api/[table]/[recordId]`) — scoped for #268. For now, `includeAllProperties: true` is used to keep full property data in the FeatureCollection so feature clicks continue to show complete details without regression.
- **Client-side caching and request de-duping** — deferred t /#271.

## LLM use disclosure

Opus 4.6 with me driving and, making adjustments.

### DIFF
--- a/components/AlertsDashboard.vue
+++ b/components/AlertsDashboard.vue
@@ -26,6 +26,7 @@ import IncidentsControls from "@/components/alerts/IncidentsControls.vue";
 import { useIncidents } from "@/composables/useIncidents";
 import { useFeatureSelection } from "@/composables/useFeatureSelection";
 import { useAlertsDateFilter } from "@/composables/useAlertsDateFilter";
+import { useRecordCache } from "@/composables/useRecordCache";
 
 import type { Layer, MapMouseEvent } from "mapbox-gl";
 import type {
@@ -58,9 +59,11 @@ const props = defineProps<{
   mapbox3d: boolean;
   mapbox3dTerrainExaggeration: number;
   mapeoData: FeatureCollection | null;
+  mapeoTable?: string;
   mediaBasePath: string | undefined;
   mediaBasePathAlerts: string | undefined;
   planetApiKey: string | undefined;
+  table: string;
 }>();
 
 const localAlertsData = ref<Feature | AlertsData>(props.alertsData);
@@ -77,6 +80,9 @@ const route = useRoute();
 const router = useRouter();
 
 const isMapeo = ref(false);
+const selectedFeatureLoading = ref(false);
+
+const { fetchRecord } = useRecordCache();
 
 // Get API key from runtime config
 const config = useRuntimeConfig();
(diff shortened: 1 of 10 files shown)


components/MapView.vue
```diff
@@ -143,36 +156,22 @@ const addDataToMap = () => {
...
-        type: feature.geotype,
-        coordinates: JSON.parse(feature.geocoordinates),
-      },
-      properties: {
-        feature,
-      },
-    })),
-  };
-
-  // Add the source to the map
```
rudokemper: The wrong comment was removed here.

What happens next is we add the source to the map. We are no longer creating a GeoJSON source with all the features.

components/MapView.vue
```diff
@@ -190,26 +189,14 @@ const addDataToMap = () => {
...
             1.0,
-            [
-              "==",
-              ["slice", ["get", props.iconColumn, ["get", "feature"]], -4],
-              ".svg",
-            ],
+            ["==", ["slice", ["get", props.iconColumn], -4], ".svg"],
             0.2,
-            1.0, // default size if neither .png nor .svg
+            1.0,
```
rudokemper: This was supposed to be a helpful comment. It isn't at all obvious to anyone that `1.0` is the default size if neither .png nor .svg, unless you are really deep into Mapbox.

Please be more careful in removing comments. Air on the side of cautious. Many / most comments in this codebase are there for a reason.

components/AlertsDashboard.vue
```diff
@@ -932,36 +894,22 @@ const addAlertsData = async () => {
 };
 
 /**
- * Adds (optional) Mapeo data to the map by creating a GeoJSON source and a layer for Point features.
- * It also sets up event listeners for user interactions with the Mapeo data features.
+ * Adds Mapeo data to the map. The data arrives as a GeoJSON FeatureCollection
+ * from the server (built by the shared spatial payload module), so it is passed
+ * directly to the map source with no reconstruction.
```
rudokemper: This is another case of a comment that doesn't make a whole lot of sense out of context, c.f.  https://github.com/ConservationMetrics/gc-explorer/pull/306/changes#r2818944282

", so it is passed directly to the map source with no reconstruction."

Writing this only makes sense if the reader is expected to know that there was reconstruction before. Otherwise, why should I need to spend mental load trying to make sense of why this is being mentioned? Since we are no longer doing that, no need to mention that.

PR #341, opened 2026-02-19
Add single-record raw endpoint

### LINKED ISSUE #268: Add single-record raw data endpoint (`GET api/[table]/[recordId]`)

conservationtimothy:
## Goal

Closes #268 — Create a single-record raw endpoint for on-demand fetching with client-side caching.

## Screenshots

N/A

## What I changed and why

Added `GET /api/[table]/[recordId]` to return a single raw record by `_id` (404 if missing, permission-validated). Introduced a `useRecordCache` composable that caches records client-side by `table::recordId` and de-dupes concurrent fetches.

Map clicks are now async: the sidebar opens with a skeleton, fetches the full record on demand, and renders `DataFeature` once loaded. This enables the map endpoint to return a minimal FeatureCollection payload (`_id`, `colorColumn`, `iconColumn`, `filterColumn` only).

The same on-demand fetch + skeleton pattern was applied to Mapeo feature clicks in the alerts dashboard. Pre-transformed alert features still render immediately. The gallery view remains unchanged since it already receives full records, and per-card lazy loading would require a bulk `GET-many` endpoint to avoid N API calls.

## What I'm not doing here

- **Bulk records endpoint** (`POST api/[table]/records` or `GET api/[table]?ids=...`) — deferred to a future sub-issue. Needed before the gallery can leverage on-demand fetching.
- **Client-side filtering and transformations** (#269) — API still returns raw data; transformations remain server-side for now.

## LLM use disclosure
Opus 4.6 with me driving

### DIFF
--- a/components/AlertsDashboard.vue
+++ b/components/AlertsDashboard.vue
@@ -146,7 +146,7 @@ const {
 
 // Fetch full Mapeo record on demand when a Mapeo feature is selected.
 // Clears the feature display and shows a loading skeleton while the
-// full record is fetched, preventing a jarring flash of stale data.
+// full record is fetched.
 let skipNextWatch = false;
 
 watch(


components/AlertsDashboard.vue
```diff
@@ -138,6 +144,36 @@ const {
   isMapeo,
 );
 
+// Fetch full Mapeo record on demand when a Mapeo feature is selected.
+// Clears the feature display and shows a loading skeleton while the
+// full record is fetched, preventing a jarring flash of stale data.
```
rudokemper:
```suggestion
// Fetch full Mapeo record on demand when a Mapeo feature is selected.
// Clears the feature display and shows a loading skeleton while the
// full record is fetched.
```

Once again, a case of LLM adding a comment that refers back to past behavior and is not necessary to include.

rudokemper (approved):
This is very good. Well scoped and great solution. I like the caching included here.

conservationtimothy:
> This is very good. Well scoped and great solution. I like the caching included here.

Thank you very much,

PR #343, opened 2026-02-21
Move filtering and Presentation Transforms to the Client

### LINKED ISSUE #269: Move filtering and presentation transformations to the client

conservationtimothy:
## Goal

Move filtering and presentation transformations to the client. Closes #269 

## Screenshots
Screenshot 2026-02-21 at 18 41 30
Screenshot 2026-02-21 at 18 41 13

## What I changed and why
Moved `transformSurveyData` from the server to a new client module (`utils/transforms.ts`) so APIs now return raw, canonical data and all presentation formatting (key renaming, capitalization, bracket cleanup) happens at render time. This decouples display logic from data delivery and ensures downloads receive untransformed records.

`GalleryView`, `MapView`, and `AlertsDashboard` now apply `transformRecord` client-side (including on-demand single-record fetches). Removed unused server helpers (`transformSurveyData`, `prepareMapData`). Also fixed Mapeo image rendering in the alerts sidebar by re-extracting photo URLs from the transformed record, since the initial selection only exposed raw keys like `p__photos`.

## LLM use disclosure

Opus 4.6 with me driving

### DIFF
--- a/components/AlertsDashboard.vue
+++ b/components/AlertsDashboard.vue
@@ -27,7 +27,7 @@ import { useIncidents } from "@/composables/useIncidents";
 import { useFeatureSelection } from "@/composables/useFeatureSelection";
 import { useAlertsDateFilter } from "@/composables/useAlertsDateFilter";
 import { useRecordCache } from "@/composables/useRecordCache";
-import { transformRecord } from "@/utils/transforms";
+import { transformSurveyEntry } from "@/utils/transforms";
 
 import type { Layer, MapMouseEvent } from "mapbox-gl";
 import type {
@@ -169,7 +169,7 @@ watch(
     // Skip the next watcher trigger caused by setting the full record
     skipNextWatch = true;
     const displayRecord = fullRecord
-      ? transformRecord(fullRecord)
+      ? transformSurveyEntry(fullRecord)
       : minimalFeature;
     selectedFeature.value = displayRecord;
     if (displayRecord.photos) {
--- a/components/GalleryView.vue
+++ b/components/GalleryView.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { getFilePathsWithExtension } from "@/utils";
 import { prepareCoordinatesForSelectedFeature } from "@/utils/mapFunctions";
-import { transformRecord } from "@/utils/transforms";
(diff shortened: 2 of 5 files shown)


conservationtimothy:
Alert-specific transforms remain server-side. In my opinion, these are tightly coupled to the alerts pipeline and will be addressed in a follow-up issue for minimal alerts. @rudokemper

IamJeffG (commented):
I see the issue and PR title both talk about "filtering" (which I admit I don't know if this is referring to row-filtering or column-filtering or what), but am I correct that actually this PR makes no change in this to where filtering happens?  (that might be fine)

The  only reference I see to "filtering" is in the functions like `prepareMapData` that were unused and now deleted.

To be clear: I'm not asking for a change, but I just want to better understand.

components/AlertsDashboard.vue
```diff
@@ -247,6 +254,7 @@ const selectInitialAlertFeature = (alertId: string) => {
  * (via MurmurHash) for Mapbox feature-state compatibility.
  */
 const selectInitialMapeoFeature = (mapeoDocId: string) => {
+  console.log("selectInitialMapeoFeature", mapeoDocId, props.mapeoData);
```
IamJeffG: remove

IamJeffG (changes requested):
Close! Minor cleanup marked inline.

As I mentioned in one of the comments, you do not need to undo the refactor of `transformSurveyData` this time.

PR #345, opened 2026-02-23
Implement client-side record caching and request de-duplication

### LINKED ISSUE #271: Implement client-side record caching and request de-duplication
## Feature Request

Add **client-side caching + request de-duping** for record fetches to prevent repeated network calls across map clicks, gallery cards, and navigation. This improves perceived performance and reduces backend load.

* **As a user interacting with multiple features**, I don’t want the app to re-fetch the same record repeatedly.
  * **AC:** Client maintains a cache keyed by `_id` for raw records.
  * **AC:** In-flight requests for the same `_id` are de-duped (subsequent calls await the same promise).
  * **AC:** Cache invalidation strategy exists and is implemented (tied to what? Let's discuss. If we can't figure it out, mark as a TODO in the code).
  * **AC:** Client-side caching and request de-duping behavior is covered by tests, including:
    * cache hits return data without issuing a network request,
    * concurrent requests for the same `_id` are de-duped and resolve from a single in-flight fetch,
    * cache invalidation correctly triggers a refetch when conditions change,
    * all tests pass in CI.
  
## Implementation Plan

conservationtimothy:
## Goal

Closes #271 — Implement client-side record caching and request de-duplication.

## What I changed and why

Refactored `useRecordCache` from a per-component instance into a module-level singleton so that map, alerts, and gallery all share one record cache keyed by `table::recordId`.  Added `fetchRecords` for bulk fetching (used by gallery pagination), `getCachedRecord` for synchronous template access, and `invalidateRecord` for cache busting. GalleryView was migrated off its local `loadedRecords` Map to use the shared cache.

## What I'm not doing here

Cache invalidation tied to a TTL, route change, or server-sent events — left as a TODO for future evaluation. The current strategy relies on explicit `clearCache()` or `invalidateRecord()` calls.

## LLM use disclosure

Opus 4.6 with me driving

### DIFF
--- a/server/api/[table]/records.post.ts
+++ b/server/api/[table]/records.post.ts
@@ -4,7 +4,7 @@ import { validatePermissions } from "@/utils/auth";
 import type { H3Event } from "h3";
 
 const MAX_IDS = 500;
-
+/** NOTE: The endpoint does not guarantee that records are returned in the same order as requested IDs. Consumers must not rely on response ordering. */
 export default defineEventHandler(async (event: H3Event) => {
   const { table } = event.context.params as { table: string };
 
@@ -52,13 +52,12 @@ export default defineEventHandler(async (event: H3Event) => {
     const records = await fetchRecords(table, stringIds);
     return records;
   } catch (error) {
-    if (error instanceof Error && "statusCode" in error) {
-      throw error;
+    if (error instanceof Error) {
+      console.error("Error fetching records:", error);
+      return sendError(event, error);
+    } else {
+      console.error("Unknown error fetching records:", error);
+      return sendError(event, new Error("An unknown error occurred"));
     }
-    console.error("Error fetching records:", error);
-    throw createError({
-      statusCode: 500,
-      statusMessage: "Internal server error",
-    });
   }
 });
--- a/server/database/dbOperations.ts
+++ b/server/database/dbOperations.ts
@@ -133,26 +133,7 @@ export const fetchRecords = async (
     WHERE _id IN (${idPlaceholders})
   `);
 
-  if (!result || result.length === 0) {
-    return [];
-  }
-
-  // Preserve request order: build a lookup and map back to input order
-  const recordMap = new Map<string, DataEntry>();
-  for (const row of result) {
-    const entry = row as DataEntry;
-    recordMap.set(String(entry._id), entry);
-  }
-
-  const ordered: DataEntry[] = [];
-  for (const id of ids) {
-    const record = recordMap.get(id);
-    if (record) {
-      ordered.push(record);
-    }
-  }
-
-  return ordered;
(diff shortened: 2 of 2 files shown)


conservationtimothy:
> Sorry I should have seen the GH issue earlier, but: this feature request is a pretty standard thing in webapps and I don't know we **need to hand-roll** our own solution. Are you familiar with
> 
> * [TanStack Query](https://tanstack.com/query/v5/docs/framework/vue/overview) is very popular
> * [swrv](https://github.com/Kong/swrv) ("stale while invalidate for vue") is lighterweight
> 
> (Disclaimer: I did not review the code yet, so it's possible that TanStack Query or swrv would be more disruptive in the short-term, but I do think it's worth a quick evaluation to at least be able to document that decision)


I am familiar with both, a couple considerations is the issue itself was rather small in scope i.e. specifically what we want cached. I know that of course, you can incrementally add these tools. 

Also these tools while easy to add they introduce their own mental overhead to learn especially around query keys tripping you up. 

> * If others are editing records (regardless whether or not we are too), then TTL is absolutely necessary, which I see you added.  Also even if data is static, still need TTL (or max cache size) to ensure the cache doesn't grow too large.

I do not think others are editing. I do agree on adding a max size

conservationtimothy:
> I am good with @conservationtimothy 's answer about building our own instead of a 3rd party lib.
> 
> Does this PR replace #312 or are they both relevant?
> 
> In any case: before proceeding let's pin down answers about cache invalidation. (my 3 specific questions above, and those were inspired #271's AC: "Cache invalidation strategy exists and is implemented (tied to what? Let's discuss. If we can't figure it out, mark as a TODO in the code)." At very least I think @rudokemper should chime in on the questions. (I know he's in meetings all day, so I'll try to sit tight until he can respond!) Do we need a call about this?

We do have eviction via max size now. May not be the best.

IamJeffG:
Thanks @rudokemper .  So I agree with the high-level strategy that @conservationtimothy describes here; I'll actually review the code next.

We should not use redis because this cache is entirely client side (inside the user's browser).

composables/useRecordCache.ts
```diff
@@ -1,21 +1,54 @@
...
+    const firstKey = cache.keys().next().value;
+    if (firstKey) {
+      cache.delete(firstKey);
+    }
+  }
+};
+// Lazy-initialized reactive counter; avoids calling ref() at import time
+// (which would fail in test environments where Vue auto-imports are not yet
+// registered on globalThis).
+let cacheCount: Ref<number> | null = null;
```
IamJeffG: This is hacky. Seems to me the better fix is to not use module-level side effects in a composable at all.

I confess I did not think through workarounds (i.e. specifically _how_ to avoid the module-level var), but am happy to brainstorm with you if you like!

PR #349, opened 2026-03-02
UX Improvement: Explicit Client-Side Error Rendering for Failed API Requests

### LINKED ISSUE #348: UX Improvement: Explicit Client-Side Error Rendering for Failed API Requests
## Feature Request
Currently, our APIs correctly return error responses when data fetching fails. However, the client does not render a visible error state in the UI. As a result, users may see blank screens or incomplete pages with no clear indication that something went wrong.

We should introduce a consistent, user-facing error page/state when API requests fail.
### User Story

As a user,
I want to see a clear and informative error page when data fails to load,
So that I understand that something went wrong and what I can do next.

### Acceptance Criteria

* If a dataset API request fails (e.g. 4xx, 5xx, network error), the user is shown a dedicated error state.
* The error page should:

  * Clearly communicate that data could not be loaded.
  * Avoid exposing sensitive internal error details.
  * Provide an actionable next step (e.g. “Retry”, “Go back”, or “Return to home”).
* The error UI should be visually consistent with the rest of the application.
* Errors should not result in blank screens or silent failures.
* Optional (nice-to-have):

  * Log the error for monitoring/observability.
  * Display a friendly fallback message for known cases (e.g. dataset not found vs. permission denied).

## Implementation Plan

conservationtimothy:
## Goal

Introduce a consistent, user-facing error state when API or data-fetching fails so users see a clear, safe message and actions (Retry, Go back, Return to home) instead of blank screens or silent failures. Closes #348 

## Screenshots
Screenshot 2026-03-02 at 10 10 19
Screenshot 2026-03-02 at 10 10 39
Incidents refetching successfully 

Screenshot 2026-03-02 at 10 07 45
Screenshot 2026-03-02 at 10 07 34
Screenshot 2026-03-02 at 10 16 41
On index, "Go Home" is hidden

## What I changed and why

I added a shared `DataLoadError` component and wired it on all seven pages that use `useFetch` for initial data (map, gallery, alerts, index, dataset, config index, config dataset). When the fetch fails, the page shows the error component with Retry (using `refresh()`), Go back, and Return to home. For config mutations (submit/remove/add) and incidents (create, fetch list, fetch details), I added toasts or inline error states with Retry where we previously only logged to the console. I added i18n keys for all new copy in en, es, nl, and pt. Optionally, DownloadMapData now shows a warning toast when download is attempted with no valid data. Unit tests were updated to mock `useToast` and `useI18n` where needed.

## LLM use disclosure

Pretty minimal Cursor auto usage used for stuff like i18n and tests

### DIFF
--- a/components/shared/DataLoadError.vue
+++ b/components/shared/DataLoadError.vue
@@ -9,7 +9,10 @@ const props = withDefaults(
   { showRetry: true },
 );
 
+const route = useRoute();
 const router = useRouter();
+
+const isHomeRoute = computed(() => route.path === "/" || route.path === "");
 
 const isRetrying = ref(false);
 
@@ -59,6 +62,7 @@ const handleGoBack = () => {
           {{ $t("goBack") }}
         </button>
         <NuxtLink
+          v-if="!isHomeRoute"
           to="/"
           class="inline-flex items-center justify-center rounded-lg border border-gray-300 bg-white px-4 py-2.5 text-sm font-medium text-gray-700 transition-colors hover:bg-gray-50"
         >


conservationtimothy:
> This matches my expectation. I agree that all the additional state vars to store error messages are in fact necessary.

Thanks for code review !

PR #351, opened 2026-03-03
Create dedicated streaming export endpoints for raw data downloads

### LINKED ISSUE #272: Create dedicated streaming export endpoints for raw data downloads
## Feature Request

Create **dedicated export endpoints** for download buttons that stream raw records to disk and return a download link, instead of loading entire datasets into memory. This protects browser memory and supports very large exports.

* **As a user downloading dataset data**, I want exports to work reliably for large datasets and reflect canonical raw records.
  * **AC:** Download buttons call export endpoints (not map/view endpoints).
  * **AC:** Export streams data server-side (no “load all records into memory” on client).
  * **AC:** Export output contains raw, untransformed values.
  * **AC:** Endpoint returns a download link or initiates a streamed download with correct headers.
  * **AC:** Large exports do not crash the browser and do not require rendering the dataset first.

## Implementation Plan

conservationtimothy:
## Goal

Closes #272 — Create dedicated streaming export endpoints for raw data downloads.

## Screenshots
Alerts: 
Screenshot 2026-03-03 at 13 44 53
Screenshot 2026-03-03 at 13 44 36
Map: 
Screenshot 2026-03-03 at 13 43 48
Screenshot 2026-03-03 at 13 43 30
Screenshot 2026-03-03 at 13 43 16

## What I changed and why

Added a new server endpoint `GET /api/[table]/export?format=csv|geojson|kml` that queries the database, formats the response server-side, and returns it as a file download with correct `Content-Type` and `Content-Disposition` headers. CSV exports include all raw columns; GeoJSON and KML exports include only records with valid coordinates, with all non-geometry fields as properties. This replaces the previous approach where the client built CSV/GeoJSON/KML strings in JavaScript from data already loaded in browser memory, which risked crashes on large datasets and required the view to be rendered before downloading.

Updated `DownloadMapData.vue` so that bulk downloads (FeatureCollection or AlertsData) call the export endpoint via `$fetch` with blob response, while single-feature downloads remain client-side. Added a `downloading` loading state and i18n key (en, es, nl, pt) to disable buttons and show feedback during export. Added tests for the export endpoint covering CSV output, GeoJSON feature filtering, format validation, and error handling.

## What I'm not doing here

Database-level cursor streaming (postgres.js cursors) — the server still loads the full table via `fetchData` before formatting. This is a future optimization for very large datasets. View-specific filtering (date ranges, filter columns) is not supported on the export endpoint; it exports the full raw table.

## LLM use disclosure

Osa & Opus.

### DIFF
--- a/test/server/exportEndpoint.test.ts
+++ b/test/server/exportEndpoint.test.ts
@@ -42,9 +42,9 @@ const sampleColumns = [
   { original_column: "Coordinates", sql_column: "g__coordinates" },
 ];
 
-// Import the endpoint helper functions by importing the module directly
-// We test the buildCsv and buildGeoJson logic via the mocked fetchData
 import { escapeCSVValue } from "@/utils/csvUtils";
+// @ts-expect-error - tokml does not have types
+import tokml from "tokml";
 
 describe("GET api/[table]/export", () => {
   beforeEach(() => {
@@ -190,15 +190,10 @@ describe("GET api/[table]/export", () => {
   });
 
   describe("KML format", () => {
-    it("builds KML from records with valid coordinates", async () => {
-      const { mainData } = await mockFetchData("test_table");
-
-      const features = mainData
-        .filter(
-          (entry: Record<string, unknown>) =>
-            entry.g__type && entry.g__coordinates,
-        )
-        .map((entry: Record<string, unknown>) => {
+    const buildGeoJsonFromData = (data: Record<string, unknown>[]) => {
+      const features = data
+        .filter((entry) => entry.g__type && entry.g__coordinates)
+        .map((entry) => {
           const properties: Record<string, unknown> = {};
           for (const [key, value] of Object.entries(entry)) {
             if (!key.startsWith("g__")) {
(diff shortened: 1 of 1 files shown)


components/shared/DownloadMapData.vue
```diff
@@ -9,11 +9,15 @@ import { escapeCSVValue } from "@/utils/csvUtils";
 const route = useRoute();
 const { t } = useI18n();
 const { warning: showWarningToast } = useToast();
+const config = useRuntimeConfig();
+const apiKey = config.public.appApiKey as string;
```
rudokemper: A good number of lines in this PR is adding API key stuff that is removed by https://github.com/ConservationMetrics/gc-explorer/pull/350.

If you ask me, once #350 is merged, you should rebase your `feat/api-refactor` branch with `main` and remove all of this, and instead implement use of `validateUserSession()`

test/server/exportEndpoint.test.ts
```diff
@@ -0,0 +1,255 @@
...
+      expect(row).toContain('"has, comma"');
+    });
+
+    it("returns empty string for empty dataset", () => {
+      const data: Record<string, unknown>[] = [];
+      expect(data.length).toBe(0);
+    });
+  });
+
+  describe("GeoJSON format", () => {
```
rudokemper: Can you add a KML test?

components/shared/DownloadMapData.vue
conservationtimothy: Yes, this was indeed my plan ! Merge #350 , merge this into `feat/api-refactor` and then rebase. Does it sound good?

test/server/exportEndpoint.test.ts
rudokemper: Left a comment on the latest commit, surfacing here in case Github hides it: https://github.com/ConservationMetrics/gc-explorer/commit/d13da53e419b10af2672b92733e50c114ac6d5a6#commitcomment-178605312

conservationtimothy: https://github.com/ConservationMetrics/gc-explorer/pull/351/commits/464d92f1f97fa8573b644e04fb398ab99e0335ef – Full posterity is i threw this at an LLM.

rudokemper (approved):

PR #353, opened 2026-03-04
Enable API response compression (gzip / brotli)

### LINKED ISSUE #273: Enable API response compression (gzip / brotli)
## Feature Request

Enable **response compression** (gzip/br) for API responses to reduce transfer size and improve load times, especially for map payloads and bulk record fetches. We can look at `@nuxtjs/precompress` for this.

* **As a user on slower connections**, I want dataset views to load faster without changing functionality.
  * **AC:** API responses include appropriate `Content-Encoding` when clients support it.
  * **AC:** Compression is enabled for relevant endpoints (at minimum: map endpoint + get-many endpoint).
  * **AC:** No regressions in response correctness; content-length/payload differences are validated in dev tools.

## Implementation Plan

conservationtimothy:
## Goal
Enable API response compression. Closes #273

## Screenshots
Screenshot 2026-03-04 at 16 29 47
Screenshot 2026-03-04 at 16 45 28

## What I changed and why

Added a Nitro `beforeResponse` plugin that compresses response bodies using Node.js built-in `zlib` (brotli preferred, gzip fallback) based on the client's `Accept-Encoding` header. We initially tried `h3-compression` but its `useCompression` helper doesn't work in the `beforeResponse` hook which I explain bellow. Nuxt internal routes (`/_nuxt`, `/__nuxt`) are skipped since those assets are already pre-compressed at build time via `compressPublicAssets`, which is also now enabled in `nuxt.config.ts`.

Also fixed a pre-existing `useRuntimeConfig` mock gap in `AlertsDashboard.test.ts` that surfaced after merging the auth refactor from `main`.

## LLM use disclosure
Osa and Opus

### DIFF
--- a/server/plugins/compression.ts
+++ b/server/plugins/compression.ts
@@ -4,8 +4,13 @@ import { brotliCompressSync, gzipSync } from "node:zlib";
  * Nitro plugin that enables brotli/gzip compression for responses.
  *
  * Reads the client's Accept-Encoding header and compresses the
- * serialized response body, preferring brotli over gzip.
+ * serialized response body, preferring brotli over gzip. Uses
+ * Node.js built-in zlib instead of h3-compression because that
+ * library only supports plaintext/HTML — it silently skips JSON
+ * API responses in the beforeResponse hook.
  *
+ * @see https://github.com/CodeDredd/h3-compression/issues/8
+ * @see https://github.com/ConservationMetrics/gc-explorer/pull/355#issuecomment-2700637498
  * @see https://nitro.build/guide/plugins
  */
 export default defineNitroPlugin((nitro) => {


conservationtimothy:
@IamJeffG @rudokemper , Moved away from `h3-compression` to Node.js built-in `zlib` (brotli/gzip). The plugin was loading and the `beforeResponse` hook was firing, but `useCompression` wasn't setting `Content-Encoding` on API responses.

This turns out to be a [known limitation](https://github.com/CodeDredd/h3-compression/issues/8) — the library [only supports plaintext content](https://github.com/CodeDredd/h3-compression/issues/8#issuecomment-2407698624), not JSON. Its `render:response` example is designed for SSR HTML where the body is already a serialized string. For API routes via `beforeResponse`, the body is still a raw JavaScript object (not yet serialized by h3), so `useCompression` silently skips it. There were also [performance concerns raised](https://github.com/CodeDredd/h3-compression/issues/5#issuecomment-2244384751) about response times jumping from 150ms to 1200ms when using `beforeResponse` with the library.

The fix: we serialize the body ourselves, compress with `brotliCompressSync`/`gzipSync` from `node:zlib`, and set the headers directly. No external dependency needed. I have updated the PR body to reflect this
Screenshot 2026-03-04 at 16 29 47

PR #390, opened 2026-03-24
Filter saved incidents by parent alerts table (view

### LINKED ISSUE #374: Filter saved incidents by parent alerts table (view)
## Feature Request

Allow the incidents list to be scoped to the current view’s alerts table (e.g. on `/alerts/fake_alerts` only show incidents whose entries belong to that context). Right now we don’t store a “parent table” on the collection, so we can’t filter that way. Check the below screenshot, Incidents includes 1. fake_alerts, GFW, and confidential incidents.

Image

### Two approaches

**Option 1 – Derive from entries**  
Infer relevance from `collection_entries.source_table`: e.g. only show incidents that have at least one entry with `source_table === currentAlertsTable` (and optionally same mapeo table).  
- **Pros:** No schema change, no backfill.  
- **Cons:** We’ve discussed supporting multiple tables per view. Once we can mix tables, “does this incident match this view?” becomes ambiguous (which entry’s table wins? what if entries span two tables?). So option 1 is probably easier to implement short-term but doesn’t give a clear, stable definition of “this incident belongs to this view.”

**Option 2 – Store parent table on the collection (recommended)**  
Add a column to the collection (or incident) that records the “parent” table/view when the incident is created (e.g. `alerts_table` or `view_name` / `primary_table`), set from the current route. List/filter incidents by that column.  
- **Pros:** Clear semantics: one canonical table per incident for filtering. Works cleanly with “one view = one alerts table” and stays consistent if we later allow mixed-table entries.  
- **Cons:** Schema change, and we need to decide:  
  - **Backfilling:** Existing incidents have no value. Either backfill (e.g. from entries’ `source_table` with some rule), or accept that old incidents don’t match any table filter until we backfill.  
  - **Nullable vs not null:** If we don’t backfill, the column must be nullable so existing rows are valid. If we backfill, we could make it non-null for new rows and leave legacy rows null (and either hide them when filtering by table or treat null as “show in all views” depending on product choice).

**Recommendation:** Prefer **option 2** for a clear, future-proof “filter saved incidents to only match the parent table” behavior. Document the chosen approach for backfilling and nullability before implementation.

- As a user on an alerts view (e.g. `/alerts/fake_alerts`), I can see only incidents that belong to that view’s context (or a clear, defined behavior for “no parent table” / legacy incidents).
- Incidents created from that view are associated with that view’s alerts table (or equivalent identifier) so they appear when filtering by parent table.
- Engineering agree on: backfill strategy for existing incidents (if any) and whether the new column is nullable and how null is treated in the list filter.

## Implementation Plan

conservationtimothy:
## Goal

Filter saved incidents by the current alerts table context (view) and persist that parent alerts table on incident creation. Closes #374.

## What I changed and why

This change introduces a canonical `parent_alerts_table` field on incidents and uses it as the source of truth for view scoping. On create, the frontend sends the current route table and the backend persists it to `incidents.parent_alerts_table`. On list, the incidents API accepts `parent_alerts_table` and applies it in the joined incidents query, so each alerts view only shows incidents that belong to that parent table context. This gives us deterministic behavior and avoids the ambiguity of inferring ownership from mixed entry tables.

## What I'm not doing here

This PR does not enforce non-null `parent_alerts_table` for legacy data and does not include a mandatory backfill policy for mixed-table incidents. Existing incidents without a parent table remain nullable until backfill/product rules are finalized.

## LLM use disclosure

Osa and Opus


components/alerts/AlertsIntroPanel.vue
```diff
@@ -49,64 +49,41 @@ const emit = defineEmits(["dateRangeChanged"]);
...
-            }}</span>
-          </div>
-          <div class="flex items-center gap-2">
-            <span class="font-bold text-sm">{{ $t("recentAlertsDate") }}:</span>
-            <span class="text-sm text-muted-foreground">{{
-              props.alertsStatistics.recentAlertsDate
-            }}</span>
-          </div>
-          <div class="flex items-center gap-2">
-            <span class="font-bold text-sm"
```
rudokemper: These are the changes from https://github.com/ConservationMetrics/gc-explorer/pull/382. Were those pulled in here, but not `feat/370-selection-mode-visual-separation`?

i18n/locales/en.json
```diff
@@ -221,7 +221,7 @@
   "pleaseMatchFormat": "Por favor, siga el formato esperado",
   "previewImagerySource": "Preview Imagery Source",
   "previousAlerts": "Previous alerts",
-  "recentAlertsDate": "Date of most recent alerts published",
+  "recentAlertsDate": "Date of most recent alerts",
```
rudokemper: Same here, changes from https://github.com/ConservationMetrics/gc-explorer/pull/382. I'd pull that into the parent branch so you have a clean diff for this PR.

rudokemper (changes requested):
Little tricky to review because this branch has code from https://github.com/ConservationMetrics/gc-explorer/pull/382 but the parent branch seemingly doesn't. But I got the gist of it. I'd resolve that before merging though.

My other biggest request is to add a test. I'd like to see a test that asserts that if there are 2 Incidents, but with different parent tables `foo` and `bar`, then if I load `alerts/foo`, only the 1 Incident associated with `foo` shows.

PR #413, opened 2026-04-13
Download button for individual record on MapView does not return all columns

### LINKED ISSUE #411: Download button for individual record on `MapView` does not return all columns

Something we missed in https://github.com/ConservationMetrics/gc-explorer/issues/256: the download generated by the Download CSV / GeoJSON / KML button for an **individual** record does not return all of the data. I think we might need a specific endpoint, or can the streaming export endpoint introduced by https://github.com/ConservationMetrics/gc-explorer/issues/272 be adapted to return a single record on demand?

Image

#### Steps to Reproduce

1. Open a Map view
2. Click on a point
3. Click any of the Download CSV / GeoJSON / KML buttons
4. See the result in the screenshot above

## Expected Behavior

The downloaded file returns all of the **raw** columns and values for the given record.

conservationtimothy:
## Goal

Fix individual map-record downloads so CSV, GeoJSON, and KML include the full raw warehouse row instead of only the reduced map feature properties. Closes #411 

## Screenshots

Screenshot 2026-04-13 at 09 59 46
Screenshot 2026-04-13 at 09 59 27

## What I changed and why

I updated the export endpoint to accept an optional `recordId` and scope the export to a single warehouse row when present. The sidebar download flow now passes that record id through to the existing export endpoint, This keeps CSV, GeoJSON, and KML behavior consistent and ensures per-record downloads include the same raw attributes users expect from full exports.

## What I'm not doing here

I am not introducing a new export endpoint.
I am not changing the bulk export format beyond reusing its raw-row behavior for single-record downloads.

## LLM use disclosure

Reasoned with GPT 5.4

### DIFF
--- a/server/api/[table]/export.get.ts
+++ b/server/api/[table]/export.get.ts
@@ -12,7 +12,13 @@ import tokml from "tokml";
 
 import type { H3Event } from "h3";
 import type { DataEntry, ColumnEntry } from "@/types";
-import type { Feature, FeatureCollection, Geometry, Position } from "geojson";
+import type {
+  Feature,
+  FeatureCollection,
+  GeoJsonProperties,
+  Geometry,
+  Position,
+} from "geojson";
 
 const SUPPORTED_FORMATS = ["csv", "geojson", "kml"] as const;
 type ExportFormat = (typeof SUPPORTED_FORMATS)[number];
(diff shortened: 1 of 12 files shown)


conservationtimothy:
We do not use`alertID` or other URL params as the export lookup source because in my opinon, the selected feature object is the more resilient source here: it is already the object we rely on for selection, and for map/sidebar export it carries the _id/id we actually need to fetch the raw record.

server/api/[table]/export.get.ts
```diff
@@ -45,58 +51,114 @@ const buildCsv = (data: DataEntry[], columns: ColumnEntry[] | null): string => {
...
+    "<Document>" +
+    `<name>${xmlEncode(documentName)}</name>` +
+    "<Placemark>" +
+    "<name>Exported record</name>" +
+    "<description>No valid map geometry in source; attributes only.</description>" +
+    `<ExtendedData>${dataTags}</ExtendedData>` +
+    "</Placemark>" +
+    "</Document></kml>"
+  );
+};
```
Comment (rudokemper) [inline server/api/[table]/export.get.ts]: I think this is not necessary. As a request to download a KML for a single feature with missing geometry will never be called.

The only way a "Download KML" button is surfaced is if a user clicks on a feature on a map to open the sidebar. So if the feature did not have correct geometry, that interaction never would have happened to begin with.

server/api/[table]/export.get.ts
```diff
@@ -45,58 +51,114 @@ const buildCsv = (data: DataEntry[], columns: ColumnEntry[] | null): string => {
 
 /**
  * Builds a GeoJSON FeatureCollection from raw database rows.
- * Only includes records with valid coordinates. All non-geometry
- * fields are included as Feature properties.
+ * Rows without valid geometry are skipped unless `allowNullGeometry` is true
+ * (single-record export), in which case one Feature per row uses `geometry: null`
+ * and still carries all non-`g__` properties.
```
Comment (rudokemper) [inline server/api/[table]/export.get.ts:56]: Here too I think the null geometry checks are sort of superfluous, for the same reason as mentioned below for KML.

But I'm more inclined to keep them here as a defensive measure.

utils/identifierUtils.ts
```diff
@@ -88,3 +91,39 @@ export const titleToCamelCase = (str: string): string => {
...
+const isGeoJsonFeature = (
+  featureGeojson?: Feature | AlertsData | null,
+): featureGeojson is Feature => {
+  return (
+    !!featureGeojson &&
+    typeof featureGeojson === "object" &&
+    "type" in featureGeojson &&
+    featureGeojson.type === "Feature"
+  );
+};
```
rudokemper: I think this is better off in `geoUtils`

rudokemper (changes requested):
Looks good, some changes requested.

Now, we also forgot one thing in the acceptance criteria:

The same code should be implemented for `AlertsDashboard` where there is an option to download data for Mapeo points:

Image

Can you file a quick PR into this branch to apply the code to `AlertsDashboard` Mapeo data downloads?

And once that is merged, and the other changes requested are resolved, we can merge this branch.

---

Side note #1 - it is nice to see the Toast working well here, even if the desired behavior is not correct! 

Side note #2 - noticing that in this screenshot, the query param `?mapeoId=` is no longer working. And that it says "Copy link to alert" which is not expected. I see that is true outside of this PR, so at some point we had a regression. Can you file an issue and solve this during this sprint? We can prioritize it before starting on the Multiple views epic.

utils/identifierUtils.ts
conservationtimothy: Moved

rudokemper (approved):

PR #415, opened 2026-04-14
Fix/alerts dashboard mapeo download export table

### LINKED ISSUE #414: Alerts dashboard: Mapeo deep link query param broken; copy-link UI shows alert copy for Mapeo

Deep-linking into the alerts dashboard with a Mapeo document id in the URL no longer selects/opens the expected Mapeo point. In practice, `?mapeoIDocd=` appears to have stopped working

Separately, when viewing a **Mapeo** selection in the sidebar, the UI still presents **“Copy link to alert”** (and related alert-oriented copy) instead of Mapeo-oriented copy (e.g. “Copy Mapeo link”). That is incorrect for Mapeo context. This appears to be a **pre-existing regression** (not introduced only by the current export [work](https://github.com/ConservationMetrics/gc-explorer/pull/413)): the same wrong label can be seen outside the single-record export [PR](https://github.com/ConservationMetrics/gc-explorer/pull/413).

#### Steps to Reproduce

1. Open the alerts dashboard view that includes Mapeo points on the map.
2. Open the app (or a saved link) with a Mapeo deep link in the query string, e.g. `?mapeoDocId=<documentId>` 
3. Observe whether the corresponding Mapeo feature is selected and the sidebar shows the expected record.
4. Select a Mapeo point from the map (if needed).
5. Observe the copy-link section in the sidebar: note the label still reads like the alert flow (“Copy link to alert”) instead of the Mapeo flow.

## Expected Behavior

- Query-based deep link to a Mapeo document should reliably open/select that Mapeo feature when the id is valid and data is loaded.
- Sidebar copy-link affordance and strings should match context: **Mapeo** selections should use Mapeo copy/link behavior and wording, not the alert-specific “Copy link to alert”.

conservationtimothy:
## Goal

Apply the single-record export flow to AlertsDashboard Mapeo point downloads. This also implictly fixes https://github.com/ConservationMetrics/gc-explorer/issues/414 and closes https://github.com/ConservationMetrics/gc-explorer/issues/414 as while working on it, the issues overlapped i.e. the error we got which affected URL also affected downloading so it got fixed. 

## Screenshots
Screenshot 2026-04-14 at 12 54 17

## What I changed and why

This follow-up applies the same server-side single-record export flow to AlertsDashboard Mapeo selections. The missing piece was not record-id resolution, but table resolution: when downloading from the AlertsDashboard sidebar, the export code still defaulted to the route table, which is correct for alerts but wrong for Mapeo points. I added an explicit `exportTableName` override through the shared download path so that Mapeo sidebar downloads target `mapeoTable` while keeping the existing route-based behavior everywhere else.

### DIFF
--- a/components/AlertsDashboard.vue
+++ b/components/AlertsDashboard.vue
@@ -1710,6 +1710,7 @@ onBeforeUnmount(() => {
       :allowed-file-extensions="allowedFileExtensions"
       :calculate-hectares="calculateHectares"
       :date-options="dateOptions"
+      :export-table-name="isMapeo ? mapeoTable : table"
       :feature="selectedFeature"
       :feature-loading="selectedFeatureLoading"
       :feature-geojson="localAlertsData"
--- a/components/shared/DataFeature.vue
+++ b/components/shared/DataFeature.vue
@@ -11,6 +11,7 @@ import type { Feature } from "geojson";
 
 const props = defineProps<{
   allowedFileExtensions?: AllowedFileExtensions;
+  exportTableName?: string;
   feature: DataEntry;
   featureGeojson?: Feature | AlertsData;
   filePaths?: Array<string>;
@@ -195,6 +196,7 @@ const exportRecordId = computed(() =>
       <DownloadMapData
         :data-for-download="featureForDownload"
         :export-record-id="exportRecordId"
+        :export-table-name="exportTableName"
         :filename-prefix="exportRecordId"
       />
     </div>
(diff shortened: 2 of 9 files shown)


rudokemper (approved):
Great fix! Thank you.

rudokemper:
Remember that when merging a PR into branch that isn't `main`, closing keywords don't work so you have to manually add the linked issue in Development :)

PR #419, opened 2026-04-15
Explorer tries to fetch data for non-existent views

### LINKED ISSUE #412: Explorer tries to fetch data for non-existent views

We don't enforce a view being configured before GC Explorer tries to fetch data from the database.

For example, in `pages/map/[tablename].vue`, the first thing we do is fetch data from the API:

```ts
const { data, error, refresh } = await useFetch(`/api/${table}/map`, {
  params: { limit: ROW_LIMIT },
});
```

On the API endpoint, we call `fetchConfig()` before fetching the data,

```ts
  try {
    const viewsConfig = await fetchConfig();

    // Check visibility permissions
    const permission = viewsConfig[table]?.ROUTE_LEVEL_PERMISSION ?? "member";

    // Validate user authentication and permissions
    await validatePermissions(event, permission);

    const { mainData } = await fetchData(table, limit);
    ...
}
...
```

But `fetchConfig()` only logs an error and returns an empty object. So nothing is stopping the API from steaming ahead and fetching the data anyway.

#### Steps to Reproduce

1. Let's assume a table `gfw-alerts` exists and has data.
2. On GC Explorer, there is no view configured for this table.
3. I access `/alerts/gfw-alerts`.
4. Although the front end will show a screen with "We couldn't load this data", server-side the data is still fetched.

conservationtimothy:
## Goal

Prevent GC Explorer from querying data for tables that do not have a configured view, so access fails early and no backend data is fetched for unsupported routes. Closes #412 

## Screenshots
Screenshot 2026-04-15 at 15 29 51
Screenshot 2026-04-15 at 15 26 02

## What I changed and why

I added a shared server-side guard (`requireTableViewConfig`) that validates a table has a non-empty view configuration before any endpoint proceeds to permission checks or warehouse queries. If no configuration exists, the API now throws a `404` with a clear error message instead of continuing into data fetch logic.

I then integrated this guard across table-scoped API endpoints (`map`, `gallery`, `alerts`, `data`, `records`, `record detail`, and export endpoints) and switched each handler to use a validated `tableConfig` object. I also added a unit test for the guard to verify both success and missing-config behavior so this regression is covered going forward.

## What I'm not doing here

- Not changing frontend route behavior or UI messaging.
- Not altering permission role semantics beyond requiring an existing view config first.
- Not introducing new config schema fields or migration logic.

## LLM use disclosure

None


server/api/[table]/alerts.ts
```diff
@@ -108,33 +109,33 @@ export default defineEventHandler(async (event: H3Event) => {
...
+      planetApiKey: tableConfig.PLANET_API_KEY,
       table,
-      routeLevelPermission: viewsConfig[table].ROUTE_LEVEL_PERMISSION,
+      routeLevelPermission: tableConfig.ROUTE_LEVEL_PERMISSION,
     };
   } catch (error) {
     if (error instanceof Error) {
       console.error("Error fetching data on API side:", error.message);
-      return sendError(event, new Error(error.message));
+      return sendError(event, error);
```
Comment (conservationtimothy) [inline server/api/[table]/alerts.ts:137]: This was introduced because I learnt in the process of implementing this pattern of creating a new error swallows the existing error code. So if it was a 404, it becomes a 500 which is incorrect.

server/api/[table]/statistics-export.get.ts
```diff
@@ -27,7 +28,8 @@ export default defineEventHandler(async (event: H3Event) => {
 
   try {
     const viewsConfig = await fetchConfig();
-    const permission = viewsConfig[table]?.ROUTE_LEVEL_PERMISSION ?? "member";
+    const tableConfig = requireTableViewConfig(viewsConfig, table);
```
Comment (IamJeffG) [inline server/api/[table]/statistics-export.get.ts]: If I understand correctly:
- fetchConfig fetches ALL datasets' configs
- requireTableViewConfig just extracts the one requested key from that

If that's correct then this is another example of "our functions are pulling back too much information from the database -- more than is needed to fulfill the request".

Would it work to roll this all up into `fetchConfig`? A single function `await fetchTableConfig(table)` that
- loads only this table's config from the database
- throw an error if the table doesn't exist.

In this way, `requireTableViewConfig` either isn't needed, or becomes only an error handler (converting the error from fetchConfig to a 404 response).

Or is there a reason that it's too difficult to extract just a single table's config in our database query?

server/api/[table]/map.ts
```diff
@@ -22,9 +22,10 @@ export default defineEventHandler(async (event: H3Event) => {
 
   try {
     const viewsConfig = await fetchConfig();
+    const tableConfig = requireTableViewConfig(viewsConfig, table);
 
     // Check visibility permissions
```
Comment (IamJeffG) [inline server/api/[table]/map.ts:26]: I agree with all these changes in this file below this point.  This view handler should be dealing with just one table config,  not all of them.

server/api/[table]/statistics-export.get.ts
Comment (conservationtimothy) [reply server/api/[table]/statistics-export.get.ts]: `fetchConfig()` does a full `select().from(viewConfig)`, so it does load all dataset configs.  
 We now use `fetchTableConfig(table)` on table-scoped API routes, which queries only one row and returns 404 when missing/empty.  
 
 I kept `fetchConfig()` only for config-index endpoints that actually need the full map.
 
 I also verified everything loads correctly.

server/utils/index.ts
```diff
@@ -8,17 +8,12 @@ export type ParsedBasemaps = {
 };
 
 /**
- * Parse basemaps configuration from ViewConfig for a given table, with legacy fallback.
+ * Parse basemaps configuration from a table ViewConfig, with legacy fallback.
```
rudokemper: The below code change is fine, but the docstring change doesn't make sense to me. I'd keep it how it was written.

rudokemper (approved):
Two great improvements here. Thanks.

PR #423, opened 2026-04-27
193.3: Update alerts API to read from new config table

### LINKED ISSUE #418: 193.3: Update alerts API to read from new config table
## Feature Request

Update alerts view API endpoints to read from `view_config_alerts` instead of the legacy JSON config column. Return primary and secondary datasets as distinct structured fields.

- As a frontend consumer, the alerts view API returns `primary_dataset` and `secondary_dataset` as explicit fields (not buried in a JSON blob)
- Existing alerts views return correct data post-migration
- API does not break if `secondary_dataset` is null
- Old JSON config column is no longer read for alerts views

### UI Sketch (or Screenshot)

N/A — API only.

## Implementation Plan

- Update relevant query/resolver to join on `view_config_alerts`
- Update API response shape to expose `primary_dataset` and `secondary_dataset`
- Add test coverage for null secondary dataset case

conservationtimothy:
## Goal

Implement subissue 193.3 by making the alerts API read dataset linkage from `view_config_alerts` and return explicit dataset fields. Closes #418 

## Screenshots

N/A (API-only change)

## What I changed and why

- Created a generic config accessor in `server/database/dbOperations.ts`:
  - Added `fetchViewDatasets(table, viewType)` for `alerts`, `map`, and `gallery`.
  - Added generic missing-config error helper (`createMissingViewDatasetConfigError`).
- Introduced generic types in `types/index.ts`:
  - `ViewType`
  - `ViewDatasets`
- Updated `server/api/[table]/alerts.ts` to use `fetchViewDatasets(table, "alerts")` 
- Preserved explicit API response fields for frontend use:
  - `primary_dataset`
  - `secondary_dataset` (nullable)
- Updated unit test `tests/unit/server/alertsEndpointDatasets.test.ts` to validate:
  - generic accessor is called with `("alerts_confidential", "alerts")`
  - null secondary dataset is handled without breaking API response.

This keeps the 193.3 scope working while aligning implementation with the epic’s reusable architecture direction.

## What I'm not doing here

- No removal of other legacy alert display/config fields yet.
- No map/gallery endpoint migration in this PR.

## LLM use disclosure

Used LLM assistance in writing tests

### DIFF
--- a/server/database/dbOperations.ts
+++ b/server/database/dbOperations.ts
@@ -10,6 +10,8 @@ import type {
   ViewType,
 } from "@/types";
 import { CONFIG_LIMITS } from "@/utils";
+import { splitCsv } from "@/utils/csvUtils";
+import type { AnyPgColumn } from "drizzle-orm/pg-core";
 
 import {
   viewConfig,
@@ -49,6 +51,79 @@ const createMissingViewDatasetConfigError = (
   error.statusCode = 404;
   error.statusMessage = statusMessage;
   return error;
+};
+
+type ViewDatasetSpec = {
+  table:
+    | typeof viewConfigAlerts
+    | typeof viewConfigMap
+    | typeof viewConfigGallery;
+  viewIdColumn:
+    | typeof viewConfigAlerts.viewId
+    | typeof viewConfigMap.viewId
+    | typeof viewConfigGallery.viewId;
+  primaryColumn:
+    | typeof viewConfigAlerts.primaryDataset
+    | typeof viewConfigMap.primaryDataset
+    | typeof viewConfigGallery.primaryDataset;
+  secondarySelections?: Record<string, AnyPgColumn>;
+  extractSecondary?: (row: Record<string, unknown>) => string[];
+};
+
+const VIEW_DATASET_SPECS: Record<ViewType, ViewDatasetSpec> = {
+  alerts: {
+    table: viewConfigAlerts,
+    viewIdColumn: viewConfigAlerts.viewId,
+    primaryColumn: viewConfigAlerts.primaryDataset,
+    secondarySelections: {
+      secondaryDataset: viewConfigAlerts.secondaryDataset,
+    },
+    extractSecondary: (row) => {
+      const value = row.secondaryDataset;
+      return typeof value === "string" && value.length > 0 ? [value] : [];
+    },
+  },
+  map: {
+    table: viewConfigMap,
+    viewIdColumn: viewConfigMap.viewId,
+    primaryColumn: viewConfigMap.primaryDataset,
+    secondarySelections: { secondaryDatasets: viewConfigMap.secondaryDatasets },
+    extractSecondary: (row) =>
+      splitCsv(
+        typeof row.secondaryDatasets === "string"
+          ? row.secondaryDatasets
+          : null,
+      ),
+  },
+  gallery: {
+    table: viewConfigGallery,
+    viewIdColumn: viewConfigGallery.viewId,
+    primaryColumn: viewConfigGallery.primaryDataset,
+  },
+};
+
+const queryViewDatasets = async (
+  spec: ViewDatasetSpec,
+  viewId: string,
+): Promise<ViewDatasets | null> => {
+  const [row] = (await configDb
+    .select({
+      primaryDataset: spec.primaryColumn,
+      ...(spec.secondarySelections ?? {}),
+    })
+    .from(spec.table)
+    .where(eq(spec.viewIdColumn, viewId))
+    .limit(1)) as Array<Record<string, unknown>>;
+
+  const primary = row?.primaryDataset;
+  if (typeof primary !== "string" || primary.length === 0) {
+    return null;
+  }
+
+  return {
+    primaryDataset: primary,
+    secondaryDatasets: spec.extractSecondary?.(row) ?? [],
+  };
 };
 
 /**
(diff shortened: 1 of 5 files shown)


rudokemper:
@conservationtimothy unrelated to the code itself -- I know we once discussed having the contents of "What I changed and why" be more human-readable and written by the code author, rather than a slew of LLM-generated bullets with implementation details -- but I think the one-paragraph format that you've been outputting recently could actually benefit from using bullets.

Like, this would read much better:

## What I changed and why
- Added `fetchAlertsViewDatasets(table)` to `server/database/dbOperations.ts` to read `primary_dataset` and `secondary_dataset` from `view_config_alerts`, with a corresponding `AlertsViewDatasets` type in `types/index.ts`. 
- The alerts endpoint (`server/api/[table]/alerts.ts`) now uses these values directly — fetching main alert records from `primaryDataset` instead of the implicit route table, and using `secondaryDataset` for the optional mapeo fetch — and returns both as explicit fields in the response. 
- Unit tests in `alertsEndpointDatasets.test.ts` cover the null secondary dataset case, verifying the endpoint succeeds and skips the secondary fetch when none is configured.

rudokemper (changes requested):
Currently, this PR unnecessarily hard-codes behavior specific to alert datasets that should be generalizable. 

Like, `createMissingAlertsDatasetConfigErrior()` could be something we use for _any_ dataset config request. And more importantly, `fetchAlertsViewDatasets()` basically just encapsulates behavior to get a primary dataset (which happens to be alerts for the Alerts dashboard) and a secondary dataset (which happens to be Mapeo data for the Alerts dashboard). 

Knowing that this Epic will touch **all** endpoints and Views, let's refactor this PR to provide general purpose functional code that any dataset view can use to get the primary and secondary dataset(s) they need.

conservationtimothy:
> Currently, this PR unnecessarily hard-codes behavior specific to alert datasets that should be generalizable.
> 
> Like, `createMissingAlertsDatasetConfigErrior()` could be something we use for _any_ dataset config request. And more importantly, `fetchAlertsViewDatasets()` basically just encapsulates behavior to get a primary dataset (which happens to be alerts for the Alerts dashboard) and a secondary dataset (which happens to be Mapeo data for the Alerts dashboard).
> 
> Knowing that this Epic will touch **all** endpoints and Views, let's refactor this PR to provide general purpose functional code that any dataset view can use to get the primary and secondary dataset(s) they need.

Implemented ur feedback 🙂

server/database/dbOperations.ts
```diff
@@ -364,6 +386,112 @@ export const fetchTableConfig = async (table: string): Promise<ViewConfig> => {
...
+      throw error;
+    }
+    console.error(
+      `Error fetching ${viewType} datasets for table "${table}":`,
+      error,
+    );
+    throw error;
+  }
+};
+
```
rudokemper, line 524: This is a slight improvement, but what I had in mind was a single map lookup that we can easily expand in the future, not an `if/if/else` chain with a lot of repeated code.

The ~25 line blocks for `if (viewType === "map")` and `if (viewType === "alerts")` are nearly identical. Whenever you see something like this, you should take the opportunity to ask yourself - how can this code be more DRY?

Take a refactor like the following:

```typescript

// This you could add to `csvUtils`
const splitCsv = (value: string | null | undefined): string[] =>
  (value ?? "")
    .split(",")
    .map((v) => v.trim())
    .filter(Boolean);

type ViewDatasetSpec = {
  table: PgTable;
  viewIdColumn: AnyPgColumn;
  primaryColumn: AnyPgColumn;
  // Omit both for the "1 primary, 0 secondary" default:
  secondarySelections?: Record<string, AnyPgColumn>;
  extractSecondary?: (row: Record<string, unknown>) => string[];
};

const VIEW_DATASET_SPECS: Record<ViewType, ViewDatasetSpec> = {
  alerts: {
    table: viewConfigAlerts,
    viewIdColumn: viewConfigAlerts.viewId,
    primaryColumn: viewConfigAlerts.primaryDataset,
    secondarySelections: { secondaryDataset: viewConfigAlerts.secondaryDataset },
    extractSecondary: (row) =>
      typeof row.secondaryDataset === "string" && row.secondaryDataset
        ? [row.secondaryDataset]
        : [],
  },
  map: {
    table: viewConfigMap,
    viewIdColumn: viewConfigMap.viewId,
    primaryColumn: viewConfigMap.primaryDataset,
    secondarySelections: { secondaryDatasets: viewConfigMap.secondaryDatasets },
    extractSecondary: (row) => splitCsv(row.secondaryDatasets as string | null),
  },
  gallery: {
    table: viewConfigGallery,
    viewIdColumn: viewConfigGallery.viewId,
    primaryColumn: viewConfigGallery.primaryDataset,
  },
};

const queryViewDatasets = async (
  spec: ViewDatasetSpec,
  viewId: string,
): Promise<ViewDatasets | null> => {
  const [row] = await configDb
    .select({ primaryDataset: spec.primaryColumn, ...spec.secondarySelections })
    .from(spec.table)
    .where(eq(spec.viewIdColumn, viewId))
    .limit(1);
  const primary = row?.primaryDataset;
  if (typeof primary !== "string" || !primary) return null;
  return {
    primaryDataset: primary,
    secondaryDatasets:
      spec.extractSecondary?.(row as Record<string, unknown>) ?? [],
  };
};

export const fetchViewDatasets = async (
  table: string,
  viewType: ViewType,
): Promise<ViewDatasets> => {
  if (process.env.CI) {
    const tableConfig = (await fetchConfig())[table];
    if (!tableConfig) {
      throw createMissingViewDatasetConfigError(table, viewType);
    }
    return {
      primaryDataset: table,
      secondaryDatasets:
        viewType === "alerts" && tableConfig.MAPEO_TABLE
          ? [tableConfig.MAPEO_TABLE]
          : [],
    };
  }
  try {
    const datasets = await queryViewDatasets(VIEW_DATASET_SPECS[viewType], table);
    if (!datasets) {
      throw createMissingViewDatasetConfigError(table, viewType);
    }
    return datasets;
  } catch (error) {
    if (error instanceof Error && "statusCode" in error) throw error;
    console.error(
      `Error fetching ${viewType} datasets for table "${table}":`,
      error,
    );
    throw error;
  }
};
```

This will make it easy as pie to add a future view type. 

In making changes, please also add more information on how to use these functions in the docstring.

```
 * Fetches dataset linkage for one table from view-type-specific config tables.
```

Only tells you what is happening in the code at a very basic level.

rudokemper (changes requested):

server/database/dbOperations.ts
conservationtimothy: Ah gotcha — implemented !

I refactored `fetchViewDatasets` to the DRY/spec-driven style. Please LMK if it's more aligned with your expectations.

rudokemper (approved):

PR #428, opened 2026-04-30
Push Column Projection to DB for `fetchData` (Eliminate Full-Table Loads)

### LINKED ISSUE #393: Push Column Projection to DB for `fetchData` (Eliminate Full-Table Loads)
## Feature Request

https://github.com/ConservationMetrics/gc-explorer/issues/256 aims to prevent loading entire tables into server memory. We should enforce database-level column projection for all `fetchData` usage so we do not depend on application-layer filtering after oversized reads.

- As a maintainer, I want every server-side fetch to select only required columns so memory usage scales with the minimal payload, not table width.
- As a maintainer, I want guardrails that prevent future accidental full-table / SELECT * style reads.

**Acceptance Criteria**
- Audit all `fetchData` callsites and document each current projection behavior.
- Update `fetchData` API and/or callsites to require explicit column selection (or a safe default that is never `SELECT *`).
- Remove code paths that fetch broad rows and then trim/filter in application memory.
- Add guardrails (tests and/or lint/check) that fail if a new `fetchData` usage performs unrestricted full-row retrieval without justification.
- Update docs/comments describing the performance rule: “projection must be pushed down to DB.”

## Implementation Plan

conservationtimothy:
## Goal

Enforces database-level projection for all `fetchData` usage so server reads no longer depend on broad row loads (`SELECT *`) followed by in-memory trimming, and add guardrails to prevent regressions. Closes #393 

## Screenshots

N/A (server-side performance and API behavior changes).

## What I changed and why

- Refactored `fetchData` to require an explicit options object with mandatory `mainColumns`, removing implicit unrestricted reads
- Projection columns are now validated/sanitized, and side tables (`__columns`, `__metadata`) are only read when explicitly requested with explicit projections
- Projection is pushed down to SQL for all route-level consumers, preventing accidental wildcard access patterns that could scale memory use with table width
- Migrated all server `fetchData` callsites (`data`, `map`, `gallery`, `alerts`, `statistics-export`, `export`) to pass explicit projections based on each endpoint's needs
- Added schema-aware column resolution at callsites where table variants differ
- Added guardrail tests to enforce `mainColumns` usage in route fetches, and updated existing unit tests for the new `fetchData` API shape
- Fixed alerts regressions during rollout by making alerts/statistics projection selection schema-aware while preserving the explicit projection path (no wildcard queries)

## What I'm not doing here

- Not introducing `SELECT *` fallback behavior.
- Not redesigning endpoint response contracts or frontend rendering logic.
## LLM use disclosure
Osa and Codex 5.3

### DIFF
--- a/server/api/[table]/alerts.ts
+++ b/server/api/[table]/alerts.ts
@@ -37,6 +37,29 @@ const ALERTS_MAIN_PROJECTION = [
   "g__coordinates",
 ];
 
+/**
+ * Keeps preferred projection columns that exist on the target table.
+ * Falls back to all available columns when none of the preferred columns exist.
+ *
+ * @param {string[]} preferredColumns - Columns this route wants to project.
+ * @param {string[]} availableColumns - Columns available on the target table.
+ * @returns {string[]} Safe projection columns to send to fetchData.
+ */
+const resolveProjectedColumns = (
+  preferredColumns: string[],
+  availableColumns: string[],
+): string[] => {
+  const projectedColumns = preferredColumns.filter((columnName) =>
+    availableColumns.includes(columnName),
+  );
+
+  if (projectedColumns.length > 0) {
+    return projectedColumns;
+  }
+
+  return availableColumns;
+};
+
 export default defineEventHandler(async (event: H3Event) => {
   const { table } = event.context.params as { table: string };
   const limit = parseAndValidateLimit(event);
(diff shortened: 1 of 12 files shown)


server/database/dbOperations.ts
```diff
@@ -88,40 +145,143 @@ const fetchDataFromTable = async (
...
 };
 
+export type FetchDataOptions = {
+  mainColumns: string[];
+  limit?: number;
+  includeColumnsData?: boolean;
+  includeMetadata?: boolean;
+  columnsTableColumns?: string[];
+  metadataColumns?: string[];
+};
```
rudokemper: Currently all of our types are in `types/index.ts`. I'd say to keep doing that to be consistent.

server/api/[table]/alerts.ts
```diff
@@ -17,6 +22,44 @@ import type { H3Event } from "h3";
...
+  "territory_name",
+  "alert_type",
+  "area_alert_ha",
+  "g__type",
+  "g__coordinates",
+];
+
+/**
+ * Keeps preferred projection columns that exist on the target table.
+ * Falls back to all available columns when none of the preferred columns exist.
```
Comment (rudokemper) [inline server/api/[table]/alerts.ts]: I'd rather we handle this more deterministically, and fail loudly if none of the preferred columns exist. That means that this is NOT an Alerts Dashboard compliant dataset and it will cause lots of load problems downstream. Can we instead show the error screen you created, client-side?

We may even want to be more strict: some columns should be **required** like the geometry ones, `_id`, `month_detec`, `year_detec`, `alert_id`. (Maybe `alert_type`?).

rudokemper (changes requested):

server/api/[table]/alerts.ts
Comment (conservationtimothy) [reply server/api/[table]/alerts.ts]: Absolutely

rudokemper (approved):

PR #429, opened 2026-05-05
193.4: Integrate primary & secondary dataset params in alerts dashboard client

### LINKED ISSUE #425: 193.4: Integrate primary & secondary dataset params in alerts dashboard client
## Feature Request

The server is now returning `primary_dataset` and `secondary_dataset` from the updated alerts API, but the client-side alerts dashboard is still passing the old `table` and `mapeo_table` parameters. Update the client to consume the new fields end-to-end.

- As a user of the alerts dashboard, data continues to load correctly after the client is updated to use the new API shape
- The client no longer references `table` or `mapeo_table` params when fetching alerts data
- `primary_dataset` is used where `table` was previously passed
- `secondary_dataset` is used where `mapeo_table` was previously passed
- Null `secondary_dataset` is handled gracefully on the client (no fetch attempted, no error thrown)
- This confirms the full end-to-end alerts refactor is functional before gallery and map work begins

### UI Sketch (or Screenshot)

N/A — no visual change expected.

## Implementation Plan

- Locate all client-side calls passing `table` / `mapeo_table` for the alerts dashboard
- Replace with `primary_dataset` / `secondary_dataset` from the API response
- Verify alerts dashboard loads correctly with and without a secondary dataset configured

conservationtimothy:
## Goal

Update the alerts dashboard client to consume `primary_dataset` and `secondary_dataset` from the alerts API so downstream client behavior no longer depends on legacy `table` / `mapeo_table` response fields. Closes #425 

## Screenshots

N/A (no visual changes expected)

## What I changed and why

- Updated the alerts page to read `primary_dataset` and `secondary_dataset` from `/api/[table]/alerts` and pass them through to the dashboard as the canonical dataset linkage
- Dashboard now uses those dataset values anywhere it previously relied on old table-style params for follow-up record fetch/export behavior
- Updated incidents composable wiring so source-table decisions align with the new dataset contract
- Added graceful handling for when `secondary_dataset` is null
- Preserves current UX while confirming the end-to-end alerts refactor path before map/gallery follow-on issues

## What I'm not doing here
N/A
## LLM use disclosure
None

### DIFF
--- a/composables/useIncidents.ts
+++ b/composables/useIncidents.ts
@@ -106,12 +106,7 @@ export const useIncidents = (
   const SOURCE_ID_KEYS = ["alertID", "_id", "source_id", "sourceId"] as const;
 
   const getCurrentAlertsTable = (): string | undefined => {
-    if (primaryDatasetRef?.value) {
-      return primaryDatasetRef.value;
-    }
-    const tableRaw = route.params.tablename;
-    if (!tableRaw) return undefined;
-    return Array.isArray(tableRaw) ? tableRaw.join("/") : String(tableRaw);
+    return primaryDatasetRef?.value;
   };
 
   /**
--- a/composables/useIncidents.ts
+++ b/composables/useIncidents.ts
@@ -40,7 +40,8 @@ export const useIncidents = (
   route: RouteLocationNormalizedLoaded,
   router: Router,
   mapLegendLayerIds?: Ref<string | undefined>,
-  mapeoTableRef?: Ref<string | undefined>,
+  primaryDatasetRef?: Ref<string | undefined>,
+  secondaryDatasetRef?: Ref<string | null | undefined>,
 ) => {
   // Incidents state management
   const incidents = ref<AnnotatedCollection[]>([]);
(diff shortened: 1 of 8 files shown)


composables/useIncidents.ts
```diff
@@ -104,12 +105,10 @@ export const useIncidents = (
...
-   *
-   * @returns Alerts table slug string or undefined.
-   */
   const getCurrentAlertsTable = (): string | undefined => {
+    if (primaryDatasetRef?.value) {
+      return primaryDatasetRef.value;
+    }
     const tableRaw = route.params.tablename;
     if (!tableRaw) return undefined;
     return Array.isArray(tableRaw) ? tableRaw.join("/") : String(tableRaw);
```
rudokemper: I might be wrong so take this question / comment at face value, but will this code ever be called again? Because if a primaryDataset was not set and passed to the Alerts Dashboard, then this incident code would never be called to begin with.

rudokemper (approved):

composables/useIncidents.ts
conservationtimothy: It is actually dead code, removed it.

Issue #437, opened 2026-05-11
[EPIC] Implement UI changes for multiple views per dataset


rudokemper:
## Feature Request

Traditionally we start with a dataset, and make a view for it. Our new mental model is to create a new view type then specify a primary (and maybe secondary) dataset.

> “I have camera trap data. I want to explore it on the same map as (something else)”

We already had this for Alerts only, with Mapeo as the secondary.

**Already done:** In the database, we’ve de-coupled views from dataset. You see this today in the “Views Management” screen as different cards for the same primary dataset.

This epic is focused on **UI changes** to reflect the new mental model and the new dataset ↔ view relationship:

1. **When seeing all existing views** — risk of too many views; need filter, search, or some way to explore without endless scrolling.
2. **When adding a new view** — flow should start from view type creation, then choose primary (and optional secondary) dataset.

Sub-issues will break this down; this epic is the umbrella only.

### User Stories & Acceptance Criteria

_(to be filled via sub-issues)_

### UI Sketch (or Screenshot)

_(to be added in sub-issues)_

nicopace:
Two ideas:
- Config: Next to the metadata, you can hint users that there are other maps, gallery or ..., or whether they can create them or not.
- Icon and color would help visually filter instead of only icon (3 options, easy to spot). Icon instead of the circular C.

rudokemper:
How do we reflect the mental model on the front end? Currently we still group `view_type` per dataset (slug). 

<img width="416" height="331" alt="Image" src="https://github.com/user-attachments/assets/6bd51e43-cd5d-4aad-bc27-a4db342871f8" />

(And notice that here, `- Map` is the title of **one** of the dataset views.

rudokemper:
> How do we reflect the mental model on the front end? Currently we still group `view_type` per dataset (slug).

Related to this, "Edit Config" on `/dataset/{dataset}` no longer makes sense

<img width="929" height="733" alt="Image" src="https://github.com/user-attachments/assets/a5524339-5855-4b62-867c-976f5643f6cf" />

rudokemper:
Also, when there are two views available, `/config/{dataset}` resolves to one of the two views available.

rudokemper:
> Also, when there are two views available, `/config/{dataset}` resolves to one of the two views available.

I am noticing a bug where for **Media** config, the Filebrowser share URL or hash is not populating, even when there is a value saved:

[...]

PR #446, opened 2026-05-13
193.9: Integrate primary_dataset param in map client

### LINKED ISSUE #443: 193.9: Integrate `primary_dataset` param in map client
### Description
The map API should return structured dataset linkage (`primary_dataset`, `secondary_datasets`), but map client code still relies on legacy route/table assumptions in parts of the flow. Update map client data fetching to consume the new API field `primary_dataset` end-to-end. `secondary_dataset` is null so does not need to be consumed. 

- As a user of the map view, map data and overlays continue to load correctly after client refactor.
- Client no longer relies on legacy dataset-link params/fields where `primary_dataset` / `secondary_datasets` are available.
- `primary_dataset` is used where base table dataset identity was previously inferred from route table.
- `secondary_datasets` is currently expected to be empty and is ignored by FE logic in this issue.

### UI Sketch (or Screenshot)
N/A — no visual redesign expected.

### Implementation Plan
- Locate map client fetch/caching paths using legacy table linkage assumptions.
- Replace with API-provided `primary_dataset`.
- Ensure map layers/overlays load correctly for:
  - primary-only configs
  - primary + one/multiple secondary datasets
- Add/update unit tests for map client dataset wiring and null/empty secondary behavior.

conservationtimothy:
## Goal

Update the map client to consume API-provided `primary_dataset` as the dataset source of truth for map record retrieval, aligning map UI behavior with the view-config dataset refactor. Closes #443 

## Screenshots

Screenshot 2026-05-13 at 09 46 22

## What I changed and why

- Map page now stores `primary_dataset` from `/api/[table]/map` and passes it down to `MapView` as `primaryDataset`, replacing route-derived table assumptions for dataset identity
- `MapView` record cache fetch path now uses `primaryDataset` when requesting full records for selected map features, consistent with alerts and gallery

## What I'm not doing here

- No frontend usage of `secondary_datasets` yet because current map configs do not yet define active secondary tables..
- No backend/migration changes in this PR.

## LLM use disclosure

None

### DIFF
--- a/components/MapView.vue
+++ b/components/MapView.vue
@@ -62,7 +62,7 @@ const props = defineProps<{
   mediaBasePathIcons?: string;
   mediaColumn?: string;
   planetApiKey?: string;
-  table: string;
+  primaryDataset: string;
   timestampColumn?: string;
 }>();
 
@@ -346,7 +346,7 @@ const addDataToMap = () => {
         // Fetch the full raw record from the single-record endpoint
         // and apply presentation transforms for display
         if (recordId) {
-          const record = await fetchRecord(props.table, recordId);
+          const record = await fetchRecord(props.primaryDataset, recordId);
           if (record) {
             const displayRecord = transformSurveyEntry(record);
             delete displayRecord["filter-color"];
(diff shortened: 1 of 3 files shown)


rudokemper (approved):
Recommend adding a TODO comment in MapView to support `secondaryDataset` with a link to the epic https://github.com/ConservationMetrics/gc-explorer/issues/437.

PR #456, opened 2026-05-14
Resolve a few type errors relating to filters and 3d exaggeration

rudokemper:
## Goal

I saw in the browser console a few Nuxt warning complaints about `filterColumn` being undefined if not set, and a mismatch with its declared type (String); and Mapbox 3d Terrain not being the right expected value always (a number - 0 if not set). This PR resolves both issues.

## LLM use disclosure

None

### DIFF
--- a/components/MapView.vue
+++ b/components/MapView.vue
@@ -42,7 +42,7 @@ import type {
 const props = defineProps<{
   allowedFileExtensions: AllowedFileExtensions;
   colorColumn?: string;
-  filterColumn: string;
+  filterColumn?: string;
   iconColumn?: string;
   mapStatistics: MapStatistics;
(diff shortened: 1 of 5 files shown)


components/MapView.vue
```diff
@@ -56,7 +57,7 @@ const props = defineProps<{
   mapboxBasemaps?: BasemapConfig[];
   mapboxZoom: number;
   mapbox3d: boolean;
-  mapbox3dTerrainExaggeration: number;
+  mapbox3dTerrainExaggeration?: number | null;
```
conservationtimothy: This type is now `number` | `null` | `undefined`. Just pointing out, it might be expected behaviour,

conservationtimothy (approved):

PR #458, opened 2026-05-14
view_config writes do not sync view_config_* dataset tables after migrations

### LINKED ISSUE #457: view_config writes do not sync view_config_* dataset tables after migrations

After initial migrations, editing or adding config rows in `view_config` does not automatically create/update matching rows in `view_config_alerts`, `view_config_map`, or `view_config_gallery`.

This creates an edge case where a table can be correctly configured in `view_config` (including `VIEWS`) but still fail dataset lookup at runtime because the per-view table row is missing.

#### Steps to Reproduce

1. Start from a database where `view_config_*` migrations have already run.
2. Add a new table via config APIs (`/api/config/new_table/[table]`) or update an existing one via `/api/config/update_config/[table]`.
3. Set `VIEWS` to include `alerts`, `map`, and/or `gallery` (and `MAPEO_TABLE` for alerts when applicable).
4. Query the corresponding endpoint (for example `/api/[table]/alerts` or `/api/[table]/gallery`).
5. Observe missing dataset linkage errors because expected row(s) are absent in `view_config_alerts` / `view_config_map` / `view_config_gallery`.

## Expected Behavior

Whenever `view_config` is inserted/updated, the corresponding per-view config tables should be synchronized in the same write path:

- `VIEWS` includes `alerts` -> upsert `view_config_alerts`
- `VIEWS` includes `map` -> upsert `view_config_map`
- `VIEWS` includes `gallery` -> upsert `view_config_gallery`
- View types removed from `VIEWS` should remove stale rows from the corresponding `view_config_*` table

So migrations only backfill historical data once, while runtime config writes keep all config tables consistent going forward.

conservationtimothy:
## Goal

Fix the runtime config sync gap where `view_config` updates could leave `view_config_alerts`, `view_config_map`, and `view_config_gallery` out of date after initial migrations. Closes #457 

## Screenshots

N/A 

## What I changed and why

- Added a centralized sync helper in `server/database/dbOperations.ts` that derives configured view types from `VIEWS` and keeps per-view config tables aligned with each `view_config` row
- Helper upserts rows into `view_config_alerts`, `view_config_map`, and `view_config_gallery` when their view types are present, and deletes stale rows when view types are removed
- For alerts, `MAPEO_TABLE` is carried over into `secondary_dataset` where applicable
- Wired the sync into both `updateConfig()` and `addNewTableToConfig()` so per-view tables are maintained at write time instead of depending on one-time migration backfills

## What I'm not doing here

- Not changing migration files or startup migration behavior.
- Not modifying frontend behavior.

## LLM use disclosure

Codex 5.3 and Osa

### DIFF
--- a/components/AlertsDashboard.vue
+++ b/components/AlertsDashboard.vue
@@ -102,11 +102,11 @@ const props = defineProps<{
   mapbox3d: boolean;
   mapbox3dTerrainExaggeration: number;
   mapeoData: FeatureCollection | null;
-  mapeoTable?: string;
+  primaryDataset: string;
+  secondaryDataset?: string | null;
   mediaBasePath: string | undefined;
   mediaBasePathAlerts: string | undefined;
   planetApiKey: string | undefined;
-  table: string;
 }>();
 
 const localAlertsData = ref<Feature | AlertsData>(props.alertsData);
@@ -167,7 +167,8 @@ const {
   route,
   router,
   toRef(props, "mapLegendLayerIds"),
-  toRef(props, "mapeoTable"),
+  toRef(props, "primaryDataset"),
+  toRef(props, "secondaryDataset"),
 );
 
 // Use feature selection composable
@@ -206,7 +207,7 @@ watch(
     }
     if (!feature) return;
 
-    const isMapeoFeature = isMapeo.value && props.mapeoTable;
+    const isMapeoFeature = isMapeo.value && !!props.secondaryDataset;
     const isMinimalAlert = !isMapeo.value && feature.alertID && feature._id;
 
     if (!isMapeoFeature && !isMinimalAlert) return;
@@ -214,7 +215,9 @@ watch(
     const recordId = feature._id || feature.id;
     if (!recordId) return;
 
-    const fetchTable = isMapeoFeature ? props.mapeoTable! : props.table;
+    const fetchTable = isMapeoFeature
+      ? props.secondaryDataset!
+      : props.primaryDataset;
     const minimalFeature = { ...feature };
     selectedFeature.value = null;
     selectedFeatureLoading.value = true;
@@ -233,7 +236,7 @@ watch(
       );
     } else {
       displayRecord = fullRecord
-        ? transformAlertEntry(fullRecord, props.table)
+        ? transformAlertEntry(fullRecord, props.primaryDataset)
         : minimalFeature;
       imageUrl.value = [];
       if (displayRecord.t0_url)
@@ -1710,7 +1713,9 @@ onBeforeUnmount(() => {
       :allowed-file-extensions="allowedFileExtensions"
       :calculate-hectares="calculateHectares"
       :date-options="dateOptions"
-      :export-table-name="isMapeo ? mapeoTable : table"
+      :export-table-name="
+        isMapeo && secondaryDataset ? secondaryDataset : primaryDataset
+      "
       :feature="selectedFeature"
       :feature-loading="selectedFeatureLoading"
       :feature-geojson="localAlertsData"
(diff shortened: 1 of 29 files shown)


rudokemper:
Hey Osa,

I've been looking at how we did this refactor, and what is going to set us up for success in the future. And I think we should revise our approach.

As it stands, having 3+ tables -- `view_config_map`, `view_config_alerts`, `view_config_gallery`, `view_config_<future_dashboard_type>` -- exist _in addition to_ a main `view_config` that stores the config JSON is not very efficient. In my understanding, we will have to JOIN `view_config_<type>.primary_dataset` with `view_config.table_name` every time we want the config, and this is both inefficient and just not in line with the goal of getting away from dataset-centric config to begin with!

I've been thinking about this, and also got @IamJeffG's insights on it, and think we can do this in a much simpler way. This is what I propose:

We have just one table, `gc_explorer_views`. It has the following columns:

- `view_id` (Integer) e.g. 1 (for DB joins only)
- `view_name` (Text) e.g. "My biodiversity monitoring map" (shown in frontend)
- `view_type` (Text, one of "map", "gallery", "alert")
- `primary_dataset` (Text) e.g. `comapeo_data`
- `secondary_dataset` (Text) e.g. `camera_trap_data`
- `view_config` (Text) e.g. the current JSON blobs that we already store in `view_config`

That way, all of the views are in one place, which means we don't have to JOIN 3+ tables to get an index like we provide on the `/` route. 

It's not a problem that the `secondary_dataset` column is available to a view which is a Gallery, because the app will ensure it is never possible for the user to set a secondary dataset when configuring a Gallery view. So for that row, the value will just be null.

`view_config` can just stay a giant JSON blob because we don't do any indexing based on any of those values anyway; they are largely just passed to the client as props to be used for state management.

In doing this, one thing we will have to resolve is that `public_views` should reference the `view_id` of the relevant record in the `gc_explorer_views` table. ~~And I think it could be renamed as `gc_explorer_views_public`.~~

---

Doing what I propose will have this PR close un-merged, and require an additional refactor. But I think it will set us up for success in the future. For example, if we add a new view type, it won't be another table and Drizzle migration. And it's just better aligned with our new model of how views work in GC Explorer.

Can you re-open the existing Epic, or create a new Issue to track this proposed refactor? I hope it won't be too much additional work - I think a lot of the code on the API side can largely remain, and it's moreso on the DB Operations side in terms of what is being queried.

Since we didn't really roll the merged code in `main` out to any instances except `demo` and `cmistaging`, I would be fine if we didn't make this another migration on top of the existing ones. 

I.e. instead of the migration order being "Construct and upsert view_config_alerts", etc. and then dropping them again, let's just start from where we were before the epic, with plain ol' dataset-centric `view_config`. And then we can manually adjust Demo and CMI Staging, that will only take a few minutes.

Let me know if you have any questions.

---

<details>
<summary>Hidden additional requests removed by @IamJeffG as out-of-scope</summary>

In proximity of this work, we may also want to rename the other Explorer tables,

- `annotated_collections` -> `gc_explorer_annotated_collections`
- `collection_entries` -> `gc_explorer_collection_entries`
- `incidents` -> `gc_explorer_incidents`

(Because we also use this database to store Guardian Connector stuff that isn't used by GC Explorer.)

</details>

conservationtimothy:
@rudokemper Will probably re-read to get a deeper understanding of requirements but initial thoughts:
 
1. I can probably hard reset to this https://github.com/ConservationMetrics/gc-explorer/commit/cbe7ce88035d335a3ecec5d95d83dd972c0b07c8 (Swahili commit) on main. It'll exclude the latest commit which effectively *rolls back*. I still have my local feature branch , the PR feature branch as well and can cherry pick or at the very least re-use certain API logic.

IamJeffG:
Starting from Rudo's comment above, I opened https://github.com/ConservationMetrics/gc-explorer/issues/459 to replace this PR.

PR #461, opened 2026-05-20
Remove pattern attributes in media config

rudokemper:
## Goal

I was seeing some errors in Chrome browser console logs when saving a config:

```js
Pattern attribute value ^(https?://[^\s]+/(?:share|api/public/dl)/[a-zA-Z0-9_-]+|[a-zA-Z0-9_-]+)$ is not a valid regular expression: Uncaught SyntaxError: Invalid regular expression: /^(https?://[^\s]+/(?:share|api/public/dl)/[a-zA-Z0-9_-]+|[a-zA-Z0-9_-]+)$/v: Invalid character in character classUnderstand this error
```

This is because we were using `pattern` attribute for media inputs, and Chrome rejects `/` and `:` as bare characters inside character classes. Anyway, turns out these added `pattern` attributes are not even useful, since we already handle the exact same validations in lifecycle computation, so this PR rips them out.

## LLM use disclosure

None

### DIFF
--- a/components/config/ConfigMedia.vue
+++ b/components/config/ConfigMedia.vue
@@ -265,7 +265,6 @@ onMounted(() => {
             type="text"
             :value="shareInputBasePath"
             placeholder="https://files.example.com/share/abc123 or abc123"
-            pattern="^(https?://[^\s]+/(?:share|api/public/dl)/[a-zA-Z0-9_-]+|[a-zA-Z0-9_-]+)$"
             @input="
               handleInput('basePath', ($event.target as HTMLInputElement).value)
             "
@@ -366,7 +365,6 @@ onMounted(() => {
             type="text"
             :value="shareInputAlerts"
             placeholder="https://files.example.com/share/abc123 or abc123"
-            pattern="^(https?://[^\s]+/(?:share|api/public/dl)/[a-zA-Z0-9_-]+|[a-zA-Z0-9_-]+)$"
             @input="
               handleInput('alerts', ($event.target as HTMLInputElement).value)
             "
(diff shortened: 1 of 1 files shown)


components/AlertsDashboard.vue
```diff
@@ -102,11 +102,11 @@ const props = defineProps<{
   mapbox3d: boolean;
   mapbox3dTerrainExaggeration: number;
   mapeoData: FeatureCollection | null;
-  mapeoTable?: string;
+  primaryDataset: string;
```
conservationtimothy: @rudokemper Not sure the best recommendation but we have `primaryDataset` in this PR which shouldn't be. Remember we reset `main` to the https://github.com/ConservationMetrics/gc-explorer/commit/cbe7ce88035d335a3ecec5d95d83dd972c0b07c8 commit which didn't have any primaryDataset work. We need to ensure we git pull from then. Current main seems fine as I cannot see `primaryDataset` https://github.com/ConservationMetrics/gc-explorer/blob/9a66f9b563507b8c999e5df19f14866483f43fed/components/AlertsDashboard.vue so this probably needs some work.

conservationtimothy (changes requested):

components/AlertsDashboard.vue
rudokemper: Oh, oops, on the plane I didn't pull latest `main` with the reset, so all this got pulled in. No worries, I will fix - this PR was a 3 line change just ripping out the `pattern` attributes.

conservationtimothy (approved):

PR #463, opened 2026-05-20
On `MapView`, move down filter group on mobile

rudokemper:
## Goal

Just a minor improvement to bump the filter group down a bit on mobile viewports, so as to avoid it from covering the "Reset Map" button.

## Screenshots

Before:
image

After: 
image

## LLM use disclosure

None


conservationtimothy (approved):

conservationtimothy:
LGTM but lint is failing @rudokemper

PR #467, opened 2026-05-22
Remove vestigial API keys

rudokemper:
## Goal

To remove some leftover API keys in headers that are no longer used in the app.


conservationtimothy (changes requested):
Unit tests are failing for `useRecordCache` probably because of these changes.

conservationtimothy (approved):

Issue #459, opened 2026-05-18
Re-implement "multiple datasets" from previous learnings


IamJeffG:
## Feature Request
> I think we can do this in a much simpler way. This is what I propose:
> 
> We have just one table, `views` (renamed from view_config). It has the following columns:
> 
> - `view_id` (Integer) e.g. 1 (for DB joins only)
> - `view_name` (Text) e.g. "My biodiversity monitoring map" (shown in frontend)
> - `view_type` (Text, one of "map", "gallery", "alert")
> - `primary_dataset` (Text) e.g. `comapeo_data`
> - `secondary_dataset` (Text) e.g. `camera_trap_data`
> - `view_config` (Text) e.g. the current JSON blobs that we already store in `view_config`
> 
> That way, all of the views are in one place, which means we don't have to JOIN 3+ tables to get an index like we provide on the `/` route. 
> 
> It's not a problem that the `secondary_dataset` column is available to a view which is a Gallery, because the app will ensure it is never possible for the user to set a secondary dataset when configuring a Gallery view. So for that row, the value will just be null.
> 
> `view_config` column can just stay a giant JSON blob because we don't do any indexing based on any of those values anyway; they are largely just passed to the client as props to be used for state management. 
>
> I.e. instead of the migration order being "Construct and upsert view_config_alerts", etc. and then dropping them again, let's just start from where we were before the epic, with plain ol' dataset-centric `view_config`. And then we can manually adjust Demo and CMI Staging, that will only take a few minutes.

_Originally posted by @rudokemper in https://github.com/ConservationMetrics/gc-explorer/issues/458#issuecomment-4453813656_

## Implementation Plan

- [x] Close #458 without merge, close #457 as wontfix. This Issue replaces it.
- [x] To undo the merge and deploy of `view-config-api-parent-branch`, apply the reverse migration to `demo` and `staging` VMs, before downgrading them to the latest release from `main` branch (either the current main or before you merged `view-config-api-parent-branch` -- they ought to be equivalent).  Now all our deployments of this app ought to be in the same state.
- [x] PR: To `view_config`, add the new columns described above + migration. App should still be runnable for single-dataset views.
   - OPTIONAL (feel free to punt til later): In doing this, `public_views` may now reference the `view_id` of the relevant record in the `views` table.
- [x] PR: Re-implement 193.3, 193.4 and 193.5 (all Alerts view features) over this revised data model.
[...]

PR #469, opened 2026-05-26
[DO NOT MERGE] Drop view_config_* tables

conservationtimothy:
## Goal

Phase 1 of #459 is to add a safe cleanup migration for environments that received the old per-view config table rollout, removing `view_config_alerts`, `view_config_map`, and `view_config_gallery` before returning those deployments to the current mainline schema.

## Screenshots
Screenshot 2026-05-26 at 12 34 20
Screenshot 2026-05-26 at 12 09 10

## What I changed and why

This PR adds a new migration that drops the old per-view config tables with `DROP TABLE IF EXISTS`. These tables came from the abandoned multi-table view config approach and should be removed before Demo/Staging are downgraded back to the current mainline app state. Using `IF EXISTS` makes the migration safe to run even if an environment has already been cleaned up or never had those tables created.

The migration is intentionally narrow and only touches the abandoned `view_config_alerts`, `view_config_map`, and `view_config_gallery` tables. It does not alter `view_config`, `public_views`, or any warehouse data.

I ran the migration on staging per the screenshots and it dropped successfully. 

## What I'm not doing here

- Not changing application code.
- Not yet implementing the new single-table `views` model work.

## LLM use disclosure

None


conservationtimothy:
We won't be merging this so no real need for review, we'd get the Docker image and deploy for staging and demo

IamJeffG (approved):

PR #470, opened 2026-05-27
Move from dataset-centric `view_config` rows to a single `views` table and related config updates

conservationtimothy:
## Goal

Replace the legacy single-blob view config model (one `view_config` row per dataset, with a comma-separated `VIEWS` key inside the JSON) with a normalized `views` table that has one row per view, making `view_type` the single source of truth for which views a dataset exposes.

## Screenshots

Screenshot 2026-05-28 at 20 04 43
Screenshot 2026-05-28 at 19 47 09
Screenshot 2026-05-28 at 19 35 15

## What I changed and why

- Normalized view configuration so each row in the `views` table represents a single view (`view_id`, `view_name`, `view_type`, `primary_dataset`, `secondary_dataset`, `view_config`); a dataset exposing multiple views is simply multiple rows sharing a `primary_dataset`
- `view_type` (one of `alerts`, `map`, `gallery`) is now the sole source of truth for a view's type — the `VIEWS` key is removed from stored JSON entirely and never reconstructed
- Deleted the `alert`/`alerts` translation shim and its helper module since tokens now match routes and i18n keys directly
- `/api/config` now returns `{ views, availableTables }`; a new `/api/config/[table]` endpoint returns a single dataset's views filtered in the database
- Cards derive identity, title, and pills from `primary_dataset`, `view_name`, and `view_type` columns, only touching JSON for content with no dedicated column (description, permission, header image)
- Rewrote the migration to create the `views` table fresh and backfill from the legacy table in a single transactional pass: expands each legacy row's comma-separated `VIEWS` list into one row per view type via `CROSS JOIN LATERAL unnest(...)`, sets `secondary_dataset` from `MAPEO_TABLE` for alerts, strips the `VIEWS` key from JSON, and drops the legacy table
- Write path now inserts an empty settings JSON on view creation and locates rows by `(primary_dataset, view_type)` on update — never writing a `VIEWS` key
- `updateConfig` requires a `view_type` and updates exactly the one `(primary_dataset, view_type)` row; `removeTableFromConfig` takes an optional `view_type`, deletes only that row, and only drops the dataset's `public_views` entry once no views remain
- Dataset-level endpoints (`data`, `records`, `export`, `[recordId]`, `statistics-export`) are now view-aware: view pages thread their route-resolved view type down as an optional `view_type` query param and the server resolves that exact view's config and permission
- Cross-table reads (e.g. an alerts page loading a record from its secondary Mapeo table) correctly send no `view_type`
- `fetchTableConfig` falls back to deterministic `ORDER BY view_id` when no view type is supplied, replacing arbitrary row selection

A fresh-create-and-backfill migration is provably correct because the entire transform is expressed as one declarative `SELECT` over the old table, the final schema is written out literally, and it runs all-or-nothing in a transaction (so no half-applied state and no need for `IF NOT EXISTS`). Unit coverage was changed from asserting Drizzle call shape (an implementation detail) to a pure-function test of the config-to-row transformation (alerts get a secondary dataset, others get null, the serialized JSON contains no `VIEWS` key, and the name falls back to the dataset name). `pnpm lint` passes, all 383 unit tests pass, and `nuxi typecheck` reports no errors in any changed file.

## What I'm not doing here

I did not add a true write-then-read-back integration test: in CI the config read paths are stubbed to a hardcoded seed while writes hit the real database, so that round-trip is best validated by a DB-backed test or by adjusting the CI read stub, which is out of scope here. 

## LLM use disclosure

LLM assistance was used to implement the changes across the backend, frontend, and migration, and to draft this PR body.


server/database/dbOperations.ts
```diff
@@ -31,23 +32,29 @@ const createMissingViewConfigError = (table: string) => {
 };
 
 /**
- * Ensures the config database contains a `view_config` table before reads/writes.
+ * Ensures the config database contains a `views` table before reads/writes.
  *
  * @returns {Promise<void>} Resolves when the table exists.
  */
 const ensureViewConfigTableExists = async (): Promise<void> => {
```
IamJeffG, line 104: Maybe this is because I'm not a Drizzle/Nuxt programmer but I don't understand why we have this function to create a table.  The migration creates the table.  How would it not exist by the time our app is running?

server/database/schemas/viewConfig.ts
```diff
@@ -1,7 +1,13 @@
-import { pgTable, text } from "drizzle-orm/pg-core";
+import { integer, pgTable, text } from "drizzle-orm/pg-core";
 
-// Guardian Connector view config table (existing table)
-export const viewConfig = pgTable("view_config", {
+// Guardian Connector views table.
+export const viewConfig = pgTable("views", {
+  viewId: integer("view_id").unique(),
```
IamJeffG: I haven't used DrizzleORM myself, but in other ORMs I've worked with, you would be telling the ORM here that this field has a sequence default (in dbOperations, you use "view_id INTEGER GENERATED BY DEFAULT AS IDENTITY UNIQUE," and in the migration you manually create the sequence).

Is https://orm.drizzle.team/docs/generated-columns relevant here?

server/database/dbOperations.ts
```diff
@@ -31,23 +32,29 @@ const createMissingViewConfigError = (table: string) => {
...
+      CREATE TABLE views (
+        view_id INTEGER GENERATED BY DEFAULT AS IDENTITY UNIQUE,
         table_name TEXT PRIMARY KEY,
-        views_config TEXT
+        view_name TEXT,
+        view_type TEXT,
+        primary_dataset TEXT,
+        secondary_dataset TEXT,
+        view_config TEXT,
+        views_config TEXT NOT NULL
```
IamJeffG: I don't understand why the table maintains two different blob cols. I think you may have misread the issue, or maybe the issue was poorly written.

My understanding is we are just keeping the **one** old JSON blob column.  I think the confusion is that its actual name had been `views_config` but the GH issue misspelled it as `view_config`; I could go either way with whether we _rename_ it to `view_config`.  My main point is that we want to _keep the same column as before_; there should not be two columns.

> _(from the PR description):_
> Config reads prefer the new view_config JSON column when present, falling back to views_config, and config writes now keep both JSON columns plus the new metadata columns in sync.

Likewise the 2-column fallback is not needed once there's only one column.  (I'd go as far as to say that "backward compatible" is not important when we're talking about the DB schema of this app, since the webapp is the only thing reading the DB tables and we always upgrade the app when we run DB migrations.  There is no "other" reader expecting the old version.)

server/database/dbOperations.ts
IamJeffG: I asked Claude and she tells me that in CI, `migrate-and-start.sh` explicitly skips migrations when CI=true, and so when tests hit the database ther's no `views` table yet. (I did not check this claim.)

If that's true, it seems to me a better practice is to just run the migrations before starting the app -- when running tests, when running the app locally, and when running in production. That would avoid having to maintain code like this and would also avoid the possibility of divergence between the table created here versus by the migration.  IMO the migration is the source of truth, full stop.

Well, this function existed long before this PR, so removing it would be for a different PR, not this one.  I'd be interested to hearing what you think about this. If you agree, we would open a new issue.

server/database/migrations/0007_add_view_config_view_columns.sql
```diff
@@ -0,0 +1,54 @@
...
+    "table_name"
+  ),
+  "view_type" = COALESCE(
+    NULLIF(BTRIM("view_type"), ''),
+    CASE
+      WHEN BTRIM(SPLIT_PART(COALESCE("views_config"::jsonb ->> 'VIEWS', ''), ',', 1)) = 'alerts' THEN 'alert'
+      ELSE NULLIF(BTRIM(SPLIT_PART(COALESCE("views_config"::jsonb ->> 'VIEWS', ''), ',', 1)), '')
+    END
+  ),
+  "primary_dataset" = COALESCE(NULLIF(BTRIM("primary_dataset"), ''), "table_name"),
```
IamJeffG: You said in the PR description

> What I'm not doing here:
> Not removing table_name

As I understand the issue, we're just renaming table_name -> primary_table.  Similalr to my comment about `view_config/views_config` I see no reason to keep both.  A simple `ALTER TABLE ... RENAME COLUMN table_name TO primary_dataset` should be sufficient.

And that operation will remove table_name for you.

IamJeffG (changes requested):
I think this issue misunderstands the new table schema proposed in the issue: in particular it should have been simple rename operations for table_name->primary_table and for views_config->view_config.

server/database/dbOperations.ts
conservationtimothy: Yeah I thought for some reason, probably since we are adding new columns that we wanted a new column. I can delete the column and the fallback. I have already ensured it works so I will happily just add to the migration that we should drop the old column.

server/database/schemas/viewConfig.ts
conservationtimothy: Yes. `.unique()` only describes a uniqueness constraint. It did not tell Drizzle that the column is generated by the database.

server/database/dbOperations.ts
conservationtimothy: I do feel like we should 100% run migrations on startup but I do think we should wait till Rudo is back to consult him on his thoughts as well. 

my guess is maybe on a super fresh database our database exists i.e. warehouse and gc but there are no tables. 

Basically i think these are two different distinct issues. The *why* we run this and whether we should run migrations on startup.

server/database/dbOperations.ts
IamJeffG: I bet the revised migration file will look far simpler after this change. Make sure it's not preserving hints of the misguided path that we're fixing.

server/database/dbOperations.ts
IamJeffG: I am not sure I understand the distinction. (My guess is: the _why_ we run this method _is because_ we haven't been running migrations on startup (e.g. test suites).

Anyway, please find a venue where you can follow-up (meeting or GH discussion) with @rudokemper next week. To me this smells like an antipattern that will continue to cause more and more work for us (keeping yet another codeblock synced with the source-of-truth, and new bugs not discovered in testing because they diverged), unless we tackle it

We can take the discussion to that new venue, I don't need to harp on it in this PR.

tests/unit/server/viewConfigModel.test.ts
IamJeffG, line 1: Thank you for adding tests, but I don't think these are doing much towards checking for correctness.

These tests are reading source files and asserting certain substrings in the source.  This is a linter, not a test 😉    By that, I mean it's verifying that code _looks_ a certain way, but does nothing to check that it _behaves_ correctly.

Please think whether there's a function or code path _that we can execute_ and check for correctness. Since this PR is mostly a DB refactor I could believe that not much is needed.  One example is `getPrimaryViewType` is a pure function so would be easy to test -- dunno if that's worth testing.

server/database/dbOperations.ts
```diff
@@ -76,6 +81,53 @@ const checkTableExists = async (
...
+    .split(",")
+    .map((view) => view.trim().toLowerCase())
+    .find((view) => view.length > 0);
+
+  if (firstView === "alerts" || firstView === "alert") {
+    return "alert";
+  }
+  if (firstView === "map" || firstView === "gallery") {
+    return firstView;
+  }
```
IamJeffG: Why is this normalization logic needed?  Where does `firstView` come from and could we have just normalized "alert" there?

UPD: ConfigViews.vue hard-codes `value="alerts"` (plural) and this code here is fixing that spelling.  So this is self-inflicted, but I'm fine with it; there's probably value in having checked these values, e.g. to avoid invalid strings ending up in the DB.

BUT, this is not about "legacy" data as the docstring says. If I understand correctly, the docstring is wrong.

server/database/migrations/0007_add_view_config_view_columns.sql
```diff
@@ -0,0 +1,97 @@
...
+  INTO existing_primary_key
+  FROM information_schema.table_constraints
+  WHERE table_name = 'views'
+    AND constraint_type = 'PRIMARY KEY'
+  LIMIT 1;
+
+  IF existing_primary_key IS NOT NULL THEN
+    EXECUTE format('ALTER TABLE "views" DROP CONSTRAINT %I', existing_primary_key);
+  END IF;
+END $$;
```
conservationtimothy: This code looks up whichever primary key currently exists on views, regardless of its name (`config_pkey`, `view_config_pkey`, `views_pkey`, etc.), drops it if present, and then adds `views_pkey` on `view_id`.

IamJeffG (changes requested):
This still needs work.  Please make sure to update the PR description (it contains outdated info), make sure to re-review your own PR before requesting review again.

One thing I am looking for in these later revisions is that the code & comments don't contain vestiges of the initial misguided path (backward compatibility, words like "legacy").

server/database/migrations/0007_add_view_config_view_columns.sql
IamJeffG: Are we actually in a scenario where we don't know the primary key  name in our deployments?  That would be problematic; how did we get here?

server/database/dbOperations.ts
conservationtimothy: > Why is this normalization logic needed? Where does firstView come

This is a transitional mapping from the old dataset-centric config shape to the new view-centric schema. In the old model, `VIEWS` could contain multiple values because one dataset config could power multiple views. In the new model, `view_type` is singular. For this first migration, we preserve current behavior by choosing the first configured view as the row’s `view_type`; later config UI/API work should stop treating VIEWS as multi-select and instead create/update one row per view.

server/database/migrations/0007_add_view_config_view_columns.sql
conservationtimothy: We know the primary key name, `table_name` (renamed to `primary_dataset`. The dynamic portion is only resolving the current constraint name from Postgres metadata before dropping it and then creating a new Pkey.

IamJeffG: Cool, this is fine.

conservationtimothy:
Next step: preserve every configured view type from the legacy `VIEWS` JSON value instead of choosing the first comma-separated entry. The `views` migration should create one row per `(primary_dataset, view_type)`, so `"gallery,map"` becomes two rows and `"alerts"` becomes one normalized `alert` row. Config updates should also target `(primary_dataset, view_type)`, with URLs like `/config/dataset?view_type=alert`, so saves update the intended view row.

server/database/migrations/0007_add_view_config_view_columns.sql
```diff
@@ -0,0 +1,202 @@
...
+DO $$
+BEGIN
+  IF EXISTS (
+    SELECT 1
+    FROM information_schema.columns
+    WHERE table_name = 'views' AND column_name = 'table_name'
+  ) AND NOT EXISTS (
+    SELECT 1
+    FROM information_schema.columns
+    WHERE table_name = 'views' AND column_name = 'primary_dataset'
```
IamJeffG: I can't imagine a reason this check is necessary.

We can expect migration 0007 to run exactly once. As it's this migration that creates `primary_dataset` (below), we can be certain that since this migration _is running now_, the column doesn't yet exist.

IamJeffG: TBH I can't imagine that the other `IF EXISTS ... views.table_name` (above) and `IF EXISTS ... views.views_config` (below) checks are necessary either.  Like, what was the scenario that would have led to this check being false at the time that migration 0007 runs?

server/database/migrations/0007_add_view_config_view_columns.sql
```diff
@@ -0,0 +1,202 @@
...
+  ) THEN
+    ALTER TABLE "views" RENAME COLUMN "views_config" TO "view_config";
+  END IF;
+END $$;
+--> statement-breakpoint
+ALTER TABLE "views"
+  ADD COLUMN IF NOT EXISTS "view_id" integer,
+  ADD COLUMN IF NOT EXISTS "view_name" text,
+  ADD COLUMN IF NOT EXISTS "view_type" text,
+  ADD COLUMN IF NOT EXISTS "secondary_dataset" text;
```
IamJeffG: Similar as above: Why would these columns exist already? What might have created them before this migration ran?

server/database/migrations/0007_add_view_config_view_columns.sql
```diff
@@ -0,0 +1,202 @@
...
+--> statement-breakpoint
+CREATE SEQUENCE IF NOT EXISTS "views_view_id_seq";
+--> statement-breakpoint
+ALTER SEQUENCE "views_view_id_seq" OWNED BY "views"."view_id";
+--> statement-breakpoint
+CREATE OR REPLACE FUNCTION normalize_view_config_for_type_0007(
+  source_config text,
+  normalized_view_type text
```
IamJeffG: I was not clear what "normalized" is supposed to mean here.  The view_types are {map, alert, gallery}, right? 

After reading more, I think maybe you are trying to rename old cases of "alert" to "alerts".  Can we just keep these map_type singular? If we did that, then no "normalization" is needed, is that right? 

Zooming out: The core idea of this issue is not about renaming "alert" to "alerts". If the issue said that, it's a typo.  I want this PR to stay focused on normalizing one `view_config` row to multiple `view` rows.  And how can we reduce the noise in this PR that's exogenous to that?

server/database/migrations/0007_add_view_config_view_columns.sql
```diff
@@ -0,0 +1,202 @@
...
+  ADD COLUMN IF NOT EXISTS "view_id" integer,
+  ADD COLUMN IF NOT EXISTS "view_name" text,
+  ADD COLUMN IF NOT EXISTS "view_type" text,
+  ADD COLUMN IF NOT EXISTS "secondary_dataset" text;
+--> statement-breakpoint
+CREATE SEQUENCE IF NOT EXISTS "views_view_id_seq";
+--> statement-breakpoint
+ALTER SEQUENCE "views_view_id_seq" OWNED BY "views"."view_id";
+--> statement-breakpoint
+CREATE OR REPLACE FUNCTION normalize_view_config_for_type_0007(
```
IamJeffG: What is going on here? What docstring would you write for this function?

Here's what I think is happening:

```suggestion
-- Reduces an (old-style, combined)  views_config JSON to the given target view_type.
--
-- For 'alert' target: Set {"VIEWS" : "alerts"} (switch to plural), clearing any other keys.
-- For 'map' target: Set {"VIEWS" : "map"} (name unchanged), clearing any other keys.
-- For other keys: Set {"VIEWS": "gallery"}, and KEEP other keys, namely
-- those related to MAPBOX_, PLANET_, and color/icons.
CREATE OR REPLACE FUNCTION normalize_view_config_for_type_0007(
```

So the idea that different view_types require different keys in their config JSON.  I agree.

Things I'd challenge you to reconsider:
- why renaming "alert" to "alerts"? That's the only one that is plural. (It could be that I'm misunderstanding. Your PR descriptions writes that you're renaming from "alerts" to "alert" -- which makes sense to me -- but I think I see this code doing the opposite?)
- I see this is maintaining the "VIEWS" key in the JSON. Seems we could drop that from the  JSON, since we already have a bona-finde  column `views_type` to store the same information.  In that case, config would be an empty JSON for those types.  This is important because we don't want two different sources of truth, they might conflict.

server/database/migrations/0007_add_view_config_view_columns.sql
IamJeffG: Also, I'm a little confused how does _the caller_ know the target normalized_view_type already? Did they have to parse this JSON already? If so, why do we need to parse it again? can we try to do the parsing in just 1 place?

Both my questions in this thread are not going to affect the overall result of the migration, and it's not like we're going to spend engineering time maintaining this migration after this is merged and run. So my line of questioning is less about "This code could be cleaner" and more around "the fact that we are doing all these convoluted things makes me question if we really understand what we're doing here".

server/database/migrations/0007_add_view_config_view_columns.sql
```diff
@@ -0,0 +1,202 @@
...
+    END
+  ),
+  "secondary_dataset" = COALESCE(
+    NULLIF(BTRIM("secondary_dataset"), ''),
+    CASE
+      WHEN LOWER(BTRIM(SPLIT_PART(COALESCE("view_config"::jsonb ->> 'VIEWS', ''), ',', 1))) IN ('alert', 'alerts')
+        THEN NULLIF(BTRIM("view_config"::jsonb ->> 'MAPEO_TABLE'), '')
+      ELSE NULL
+    END
+  )
```
IamJeffG: When a single input row had multiple comma-separated values in the {"Views"} piece of the  JSONB, I believe those need to be extracted to their own separate rows, each with its own view_type.  That's not happening here: SPLIT_PART(..., ..., 1) always takes the first element only.

UPDATE: I believe the row expansion happens below with the `CROSS JOIN LATERAL`.

In that case, what is this UPDATE actually trying to do?


This migration is very complicated, I suggest overly so.   Ideally can it be streamlined and made more clear? Concrete things across this file that would make a hueg difference:
* assume the input schema, don't guess;
* reduce or eliminating all this string munging -- or if it's truly necessary explain very clearly why. (I find myself doubting its necessity)
* Code comments explaining what we're doing at each step in simple prose!

pages/dataset/[tablename].vue
```diff
@@ -28,8 +30,16 @@ const { data, error, refresh } = await useFetch("/api/config");
 if (data.value && !error.value) {
   const fetchedViewsData = data.value[0] as Views;
   viewsConfig.value = fetchedViewsData;
+  viewRows.value = (data.value[2] ?? []) as ViewConfigRow[];
```
Comment (IamJeffG) [inline pages/dataset/[tablename].vue]: What is this magic number 2?

I mean, it's the 3rd element of an array, but _what is that_ and _how do we know it's always at that index and wont change_?  

---

Same question about value[0] a few lines up.


I wonder if things could be more clear if there was a way to destructure `data` to named vars.

pages/dataset/[tablename].vue
```diff
@@ -28,8 +30,16 @@ const { data, error, refresh } = await useFetch("/api/config");
...
 
-  if (fetchedViewsData[tableName]) {
+  const matchingRows = viewRows.value.filter(
+    (row) => row.primaryDataset === tableName,
+  );
```
Comment (IamJeffG) [inline pages/dataset/[tablename].vue]: pls add some inline code comments to your added code in this file.  MIght also revisit some very generic variable names like `data` and `matchingRows`

Here's what I'm gleaning:


```suggestion
  // `viewRows` contains ALL rows of the entire database table.
  // Now we filter to those whose primaryDataset is the table the user is viewing.
  const viewsOnThisTable = allViewRows.value.filter(
    (view) => view.primaryDataset === tableName,
  );
```


Out of scope: that filter operation should have been pushed to the database.  Dumping entire DB tables is a bad habit for us :-)

pages/dataset/[tablename].vue
```diff
@@ -28,8 +30,16 @@ const { data, error, refresh } = await useFetch("/api/config");
...
+
+  if (matchingRows.length > 0) {
+    datasetConfig.value = matchingRows[0].viewConfig;
```
Comment (IamJeffG) [inline pages/dataset/[tablename].vue]:
```suggestion
    // There are multiple views on this table! Pick one arbitrarily, ignore the other.
    datasetConfig.value = matchingRows[0].viewConfig;
```

pages/dataset/[tablename].vue
```diff
@@ -28,8 +30,16 @@ const { data, error, refresh } = await useFetch("/api/config");
...
+    dataFetched.value = true;
+  } else if (fetchedViewsData[tableName]) {
     datasetConfig.value = fetchedViewsData[tableName];
     dataFetched.value = true;
```
Comment (IamJeffG) [inline pages/dataset/[tablename].vue]: What is fetchedViewsData?  (related to my comment above about `data.values[0]`  Sorry, but I am absolutely lost in this file...

A reason I am pushing back is that all the if/elseif/else has me worried there might still be some vestiges of the old "backward compatibility fallback" stuff we had talked about.  I guess I am surprised there are so many different ways to look up the views for a table.

pages/dataset/[tablename].vue
```diff
@@ -77,6 +87,14 @@ const canViewDataset = computed(() => {
...
+  const rowViews = viewRows.value
+    .filter((row) => row.primaryDataset === tableName)
+    .map((row) => toConfigView(row.viewType));
+
+  if (rowViews.length > 0) {
+    return Array.from(new Set(rowViews)).sort();
+  }
+
   if (!datasetConfig.value?.VIEWS) return [];
   return datasetConfig.value.VIEWS.split(",").map((v) => v.trim());
```
Comment (IamJeffG) [inline pages/dataset/[tablename].vue]: Is this another vestige of "the old way", something that the migration should have made unnecessary?

utils/viewTypes.ts
```diff
@@ -0,0 +1,7 @@
+import type { ViewType } from "@/types";
+
+export const toViewType = (view: string): ViewType =>
+  (view === "alerts" ? "alert" : view) as ViewType;
+
+export const toConfigView = (viewType: ViewType): string =>
+  viewType === "alert" ? "alerts" : viewType;
```
IamJeffG: Similar to my comments elsewhere, why do we need to expect different spellings at this point?  After the normalization had run, where are the other spellings coming from?

server/database/dbOperations.ts
```diff
@@ -576,14 +656,28 @@ export const updateConfig = async (
       }
     }
 
-    const configString = JSON.stringify(config);
+    const configViewType = toViewType(typedConfig.VIEWS as string);
+    const currentViewType = viewType ?? configViewType;
```
IamJeffG: What are these lines doing?

Per a comment in the migrations/0007 file, I don't think that the `config.VIEWS` JSON need to maintain the view_type anymore.  (It seems to me that is what line 659 is reading)

components/ConfigDashboard.vue
```diff
@@ -175,16 +188,16 @@ watch(tableNameToAdd, (newVal) => {
...
             </div>
           </div>
         </div>
 
         <div class="flex flex-wrap gap-1.5 mb-4 overflow-hidden">
           <span
-            v-for="view in config.VIEWS
-              ? config.VIEWS.split(',')
+            v-for="view in row.viewConfig.VIEWS
+              ? row.viewConfig.VIEWS.split(',')
```
IamJeffG: Again, after migration 0007 has run, I never expect there to be comma-separated view-types.  (Indeed, my other comments also strongly suggest to not even have teh "VIEWS" key at all in viewConfig column.)

tests/unit/server/viewConfigModel.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        MAPEO_TABLE: "mapeo_data",
+        ROUTE_LEVEL_PERMISSION: "anyone",
+        VIEWS: "alerts",
+      }),
+      viewName: "Fake Alerts",
+      viewType: "alert",
+    });
+    expect(mocks.where).toHaveBeenCalledTimes(1);
+    expect(mocks.insert).toHaveBeenCalledTimes(1);
+    expect(mocks.values).toHaveBeenCalledWith({ tableName: "fake_alerts" });
```
IamJeffG: Making assertions on implementation details of the ORM is fragile. (correct me if I'm wrong and that's not what is happening)

I'd expect a test like this to (1) not mock the ORM, but instead actually write to a testing database , (2) call updateConfig, (3) read the database table and assert what was written.

IamJeffG (changes requested):
To be honest, I find it confusing what's going on. The PR description makes perfect sense, but the code doesn't reflect that, at least by my reading.

Possible next steps:
* If you want to take another pass to try to recover this PR, feel free, but our next code review would be a synchronous call where you walk me through the changeset.  (I'd suggest this one if you believe that my reading of the code is wrong,  I'm happy to have you push back if you think my comments are misguided).
* Or we can abandon this PR and try again fresh. The fresh start would first involve a written implementation plan that we review together synchronously before starting coding. (I'd suggest this one if you think you are simply not sure about either the goal or about most of my comments)

server/database/dbOperations.ts
IamJeffG: https://github.com/ConservationMetrics/gc-explorer/issues/472

server/database/migrations/0007_add_view_config_view_columns.sql
```diff
@@ -0,0 +1,202 @@
...
+      jsonb_set(
+      jsonb_set(
+      jsonb_set(
+      jsonb_set(
+      jsonb_set(
+      jsonb_set(
+      jsonb_set(
+      jsonb_set(
+        jsonb_set(source_config::jsonb, '{VIEWS}', to_jsonb('gallery'::text), true),
+        '{MAPBOX_STYLE}', to_jsonb(''::text), true),
```
IamJeffG: "" is not the same as absent. We would want to delete the key, right?

server/database/dbOperations.ts
```diff
@@ -356,152 +389,257 @@ export const fetchTableNames = async (): Promise<string[]> => {
...
+};
 
-    return viewsConfig;
+/**
+ * Fetches every configured view for one dataset as its own row.
+ *
+ * @param {string} primaryDataset - Dataset table whose views are loaded.
+ * @returns {Promise<ViewConfigRow[]>} View rows scoped to the dataset.
+ */
+export const fetchViewConfigRowsForTable = async (
```
conservationtimothy, line 631: the per-dataset config page (`pages/dataset/[tablename].vue`) only needs one dataset's views. Previously the page fetched the entire config map and filtered client-side but this function does the filter in SQL

server/database/dbOperations.ts
```diff
@@ -592,11 +748,15 @@ export const updateConfig = async (
   }
 };
 
-export const addNewTableToConfig = async (tableName: string): Promise<void> => {
+export const addNewTableToConfig = async (
```
conservationtimothy, line 823: `addNewTableToConfig(table, viewType)`. Adding a view inserts a brand-new row with an empty `{}` config

server/database/dbOperations.ts
```diff
@@ -576,14 +715,31 @@ export const updateConfig = async (
       }
     }
 
-    const configString = JSON.stringify(config);
+    // A view type is required: a dataset can have several views (e.g. map +
+    // gallery), so we must identify exactly one (primary_dataset, view_type) row.
+    // Without it we would update every view of the dataset and null their type.
+    if (!viewType) {
```
conservationtimothy, line 793: in the old code, a save without a view type would have matched every row of the dataset and rewritten all of them (and nulled `view_type`). With multiple views per dataset that's silent cross-view corruption. The guard makes that state unreachable

server/database/dbOperations.ts
```diff
@@ -553,6 +691,7 @@ export const fetchPublicViewTableNames = async (): Promise<string[]> => {
 export const updateConfig = async (
```
conservationtimothy, line 763: `viewType` is now required, and the `UPDATE` targets a single (`primary_dataset`, `view_type`) row

server/database/dbOperations.ts
```diff
@@ -606,14 +766,32 @@ export const addNewTableToConfig = async (tableName: string): Promise<void> => {
...
+    // Only drop the dataset's public_views entry once no views remain for it.
+    const remainingViews = await configDb
+      .select({ viewId: viewConfig.viewId })
+      .from(viewConfig)
+      .where(eq(viewConfig.primaryDataset, tableName));
+    if (remainingViews.length === 0) {
+      await configDb
+        .delete(publicViews)
+        .where(eq(publicViews.tableName, tableName));
+    }
```
conservationtimothy, line 872: This fixes asymmetry as "add" was per-view but "remove" used to nuke the whole dataset. Now both are per-view, and we don't accidentally revoke public access to a dataset's other still-existing views.

server/api/[table]/[recordId].get.ts
```diff
@@ -2,19 +2,21 @@ import { fetchRecord, fetchTableConfig } from "@/server/database/dbOperations";
 import { validatePermissions } from "@/utils/accessControls";
 
 import type { H3Event } from "h3";
+import type { ViewType } from "@/types";
```
Comment (conservationtimothy) [inline server/api/[table]/[recordId].get.ts:5]: The data endpoints — /`api/[table]/data`, `records`, `export`, `[recordId]`, `statistics-export` -- each start by calling `fetchTableConfig(table, viewType?)` to resolve permission + render settings before touching `warehouseDb`. 

The problem was with multiple views per dataset, which view's config/permission applies?

The frontend supplies `view_type` so UI traffic always hits the right view. The view page's URL segment (/map/:table, /gallery/:table, /alerts/:table) is the view type, so `resolveViewTypeForTable` derives it from the route and the `composables` append it as a query param -- but only for the route's own dataset (an alerts page reading its secondary Mapeo table sends nothing, because that table has no alerts view).

server/database/dbOperations.ts
```diff
@@ -76,6 +82,33 @@ const checkTableExists = async (
...
+/**
+ * Builds the new single-table view metadata columns from existing config shape.
+ *
+ * @param {string} primaryDataset - Dataset table used as the route identifier.
+ * @param {ViewConfig} config - Parsed view config.
+ * @param {string} configString - Serialized config JSON.
+ * @param {ViewType | null} viewType - View type for the row.
+ * @returns New view metadata column values for views.
+ */
+export const buildViewConfigColumns = (
```
conservationtimothy, line 159: For a bit more context: this turns a parsed view config plus its view type into the exact column values for one `views` row: `viewName` (the config's `DATASET_TABLE` trimmed, else the dataset name), `viewType`, `primaryDataset`, `secondaryDataset` (the `MAPEO_TABLE` only when `viewType === "alerts"`, else `null`), and `viewConfig` (the serialized settings JSON, which never contains a `VIEWS` key). It is the single place that decides how a config maps to row columns

server/database/dbOperations.ts
conservationtimothy: to-do: fetchViewConfigRows, document what it does a lot better i.e. returning data without going to the DB

server/database/dbOperations.ts
```diff
@@ -576,14 +715,31 @@ export const updateConfig = async (
...
 
     await configDb
       .update(viewConfig)
-      .set({
-        viewsConfig: configString,
-      })
-      .where(eq(viewConfig.tableName, tableName));
+      .set(viewColumns)
+      .where(
+        and(
```
conservationtimothy, line 810: to-do composite key on viewType and primarydataset

server/database/dbOperations.ts
```diff
@@ -576,14 +715,31 @@ export const updateConfig = async (
...
+          eq(viewConfig.primaryDataset, tableName),
+          eq(viewConfig.viewType, viewType),
```
IamJeffG, line 812: you will need a  UNIQUE constraint on these joint columns. That should be done in this PR, declard on the drizzle model schema.


In the future we may want to refactor these endpoints to be GET and POST-ing URLs that include the view_id itself.  No need to do that here.

components/ConfigDashboard.vue
```diff
@@ -25,15 +26,13 @@ const emit = defineEmits([
...
-    .sort()
-    .reduce(
-      (accumulator, key) => {
-        accumulator[key] = props.viewsConfig[key];
-        return accumulator;
-      },
-      {} as Record<string, ViewConfig>,
+  return [...props.viewRows].sort((first, second) => {
+    const firstName = first.viewName || first.primaryDataset;
+    const secondName = second.viewName || second.primaryDataset;
```
IamJeffG: Correct me if I'm wrong but I seem to recall we talked about the view name always being set in the database.  So by the time data gets to the frontend, it will never be the case that viewName is falsey.

If you agree that is what we had decided, then the database can even enforce this, with a `view_name NOT NULL` column type.

components/ConfigDashboard.vue
```diff
@@ -138,16 +144,16 @@ watch(tableNameToAdd, (newVal) => {
       class="grid grid-cols-1 md:grid-cols-2 lg:grid-cols-3 gap-4 sm:gap-6 items-stretch mb-6"
     >
       <div
-        v-for="(config, tableName) in sortedViewsConfig"
-        :key="tableName"
+        v-for="row in sortedViewsConfig"
+        :key="row.viewId ?? `${row.primaryDataset}-${row.viewType}`"
```
IamJeffG: Would viewId ever be falsey?  Maybe if we are adding a new View and it hasn't been written to the DB yet? If so you might document that.

composables/useViewType.ts
```diff
@@ -0,0 +1,56 @@
...
+  map: "map",
+  gallery: "gallery",
+  alerts: "alerts",
+};
+
+/**
+ * Resolves the view type a data request should carry, given the current route
+ * and the table being fetched.
+ *
+ * Why this exists: a dataset can now expose several views (e.g. map + gallery),
```
IamJeffG:
```suggestion
 * Why this exists: a dataset can expose several views (e.g. map + gallery),
``

this is an example where LLMs like to describe the change they make, instead of the current state of the system.

File: composables/useViewType.ts [hunk 34]
```diff
@@ -0,0 +1,56 @@
...
+ * each with its own row and potentially its own settings/permission. The
+ * dataset-level endpoints (data, records, export, [recordId], statistics-export)
+ * therefore need to know WHICH view is asking so the server resolves the right
+ * row instead of an arbitrary one. A view page's first path segment identifies
```
IamJeffG: This is exactly the use case to support eventual refactor of the routes to be keyed off view_id instead of tableName + type. (Later though)

composables/useViewType.ts
```diff
@@ -0,0 +1,56 @@
...
+ * the view it renders (`/map/:table`, `/gallery/:table`, `/alerts/:table`), so
```
IamJeffG: Do all the dataset-level endpoints you mentioned on line 15 look like this?  Trying to understand.

ALSO, is this composable ever called with the secondary `table`, or only with the primary?  If only the primary, then guard #1 is never expected to actually be triggered. It's purely defensive against a software bug/error, right?  That would be worth calling out, or possibly remove it.

If we do call it with the secondary `table` then it's correct and keep it.

composables/useViewType.ts
```diff
@@ -0,0 +1,56 @@
...
+ * we derive the view type from the route and pass it down.
+ *
+ * Two guards make this safe; both are deliberate:
```
IamJeffG:
```suggestion
 * Two guards make this safe:
```

This is a vapid comment that LLMs like to say. I hope all our code is deliberate!

server/database/migrations/0007_add_view_config_view_columns.sql
```diff
@@ -0,0 +1,47 @@
...
+-- becomes multiple rows that share the same primary_dataset.
+CREATE TABLE views (
+  view_id integer GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+  view_name text,
+  view_type text NOT NULL,
+  primary_dataset text NOT NULL,
+  secondary_dataset text,
+  view_config text NOT NULL,
+  -- A dataset exposes each view type at most once, so (view_type, primary_dataset)
+  -- is the logical key. Enforce it in the DB so a dataset can never get two rows of
```
IamJeffG:
```suggestion
  -- is the logical key. Enforce it in the DB so a primary_dataset can never get two rows of
```

composables/useViewType.ts
```diff
@@ -0,0 +1,56 @@
+import type { ViewType } from "@/types";
+
+const VIEW_TYPE_BY_SEGMENT: Record<string, ViewType> = {
+  map: "map",
+  gallery: "gallery",
+  alerts: "alerts",
+};
+
+/**
```
IamJeffG, line 26: This docstring is very hard to understand.

It over-explains the system architecture. My suggestion will sound minor but I think it'll be powerful: Can you move the wider context and system architecture to a module-level docstring in this same file (why view types exist, how the server uses them, the secondary-table failure mode), and then reduce this function's docstring to only what it does?  The test suite comment does a good job of capturing what this fn does at the right level: "decides whether a data request (to the server) should carry a view_type, and which one"

The function doesn't know anything about server behavior ("the server falls back to its deterministic default (oldest view by `view_id`)"); those are true facts about the system so can go at the module level.  Likewise explaining what the caller does with `undefined`.

composables/useViewType.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * 2. Whitelist the segment instead of trusting it. We look the first segment up
+ *    in `VIEW_TYPE_BY_SEGMENT` rather than returning it raw. This is NOT about
+ *    URL typos (those 404 before any fetch); it is about routes whose first
+ *    segment is a real word but not a view type — most importantly
+ *    `/dataset/:table`, where guard #1 passes (the table matches) but the
+ *    segment is `"dataset"`. Returning that raw would send `view_type=dataset`
+ *    and trigger a missing-config error. The lookup maps any non-view segment to
+ *    undefined, and it also keeps the return type `ViewType | undefined` without
+ *    an unsafe cast.
```
IamJeffG: way too verbose, to the point that I think it's misleading. 


```suggestion
 * 2. Confirm that `route` is a known, well-formed server route.
```

Yes, it uses a whitelist but only to ensure it's a known route.  It's a stretch to say that's about security because I would hope the server is implementing its own security.  And although you say "this is not about typos" ... a typo is one very possible reason that this might not be a well-formed server route.

composables/useRecordCache.ts
```diff
@@ -69,7 +72,13 @@ export const useRecordCache = () => {
       return pending.get(cacheKey)!;
     }
 
-    const request = $fetch<DataEntry>(`/api/${table}/${recordId}`)
+    const viewType = resolveViewTypeForTable(route, table);
+    const url = `/api/${table}/${recordId}`;
+    const request = (
+      viewType
```
IamJeffG, line 80: Will viewtype ever be falsey? If so, it's worth documenting why. If not can this be reduced?

types/index.ts
```diff
@@ -66,15 +69,30 @@ export interface Views {
...
+  view_config: string;
+};
+
+export type ViewConfigRow = {
+  viewId?: number | null;
+  viewName?: string | null;
+  viewType: ViewType;
+  primaryDataset: string;
+  secondaryDataset?: string | null;
+  viewConfig: ViewConfig;
```
IamJeffG, line 95: This duplication is strange and unexpected. Does drizzle have a way to alias the snake-cased database column names as camel-cased Typescript vars?

Or is this about something else? In which case it needs to be documented.

IamJeffG (commented):
Looking a lot better!


What will follow-up on the `TODO(single-source-of-truth)` look like? Ideally its own PR; do you want to add a new sub-issue or even just a new TODO checkbox item to the issue this is about?

Likewise #472 ought to include the write-then-read-back DB tests that we're unable to add here.

components/ConfigDashboard.vue
conservationtimothy: Decided that CI seed rows now use fixed IDs 1–5; `view_id` is now just always a number.

components/ConfigDashboard.vue
conservationtimothy: Yes you are right, fixed this.

composables/useViewType.ts
conservationtimothy: Yep -- `resolveViewTypeForTable` is called with the secondary Mapeo table on alerts pages. In `AlertsDashboard.vue`:
```ts
    const fetchTable = isMapeoFeature ? props.mapeoTable! : props.table;
    ...
    const fullRecord = await fetchRecord(fetchTable, recordId);
```

On `/alerts/fake_alerts`, when the user selects a Mapeo feature, `fetchTable`is `mapeo_data` (or whatever `MAPEO_TABLE` is) — not `fake_alerts`. Guard #1 returns `undefined`, so we don't send `view_type=alerts` for a table that has no alerts view row.

Same pattern for Mapeo exports via `useTableExportDownload` when `exportTableName` overrides the table.

composables/useViewType.ts
conservationtimothy: I've implemented this feedback and hopefully, it is clearer

composables/useRecordCache.ts
conservationtimothy: Yes, which I touched on [here](https://github.com/ConservationMetrics/gc-explorer/pull/470/changes#r3390610254)

IamJeffG (approved):
Yay!!

Please make sure to heed my comments from my previous review about what will happen after merging this:

> What will follow-up on the TODO(single-source-of-truth) look like? Ideally its own PR; do you want to add a new sub-issue or even just a new TODO checkbox item to the issue this is about?
> 
> Likewise https://github.com/ConservationMetrics/gc-explorer/issues/472 ought to include the write-then-read-back DB tests that we're unable to add here.

_Originally posted by @IamJeffG in https://github.com/ConservationMetrics/gc-explorer/pull/470#pullrequestreview-4468966903_

conservationtimothy:
> Yay!!
> 
> Please make sure to heed my comments from my previous review about what will happen after merging this:
> 
> > What will follow-up on the TODO(single-source-of-truth) look like? Ideally its own PR; do you want to add a new sub-issue or even just a new TODO checkbox item to the issue this is about?
> > Likewise #472 ought to include the write-then-read-back DB tests that we're unable to add here.
> 
> _Originally posted by @IamJeffG in [#470 (review)](https://github.com/ConservationMetrics/gc-explorer/pull/470#pullrequestreview-4468966903)_

https://github.com/ConservationMetrics/gc-explorer/issues/486 Created a subissue

Issue #426, opened 2026-04-29
193.5: Streamline DB operations to fetch primary & secondary datasets in a single request


conservationtimothy:
## Feature Request

Currently the database operations code makes separate requests when fetching primary and secondary dataset data. Refactor so that both are fetched in a single joined DB request, accepting `secondary_dataset` as an optional parameter.

### User Stories & Acceptance Criteria

- As a developer, there is a single DB request that retrieves both primary and secondary dataset data when both are configured
- The refactored code accepts an optional `secondary_dataset` parameter — if null, only the primary dataset is fetched
- No regression in data returned for existing alerts, gallery, or map views
- Subsequent gallery and map API work (193.6, 193.7) can build on this without further DB layer changes

### UI Sketch (or Screenshot)

N/A — DB/backend only.

## Implementation Plan

- Identify the DB operations code handling data set fetching
- Modify to accept optional `secondary_dataset` param and perform a join / simultaneous fetch when present
- Rework relevant endpoints to pass through the secondary param
- Confirm no duplicate requests in the network layer after change

PR #471, opened 2026-05-27
use views dataset fields for alerts primary and secondary data

conservationtimothy:
## Goal

Re-implement the Alerts multi-dataset API/client flow on top of the new single-table `views` data model.

## Screenshots

Screenshot 2026-05-27 at 10 28 06

## What I changed and why

This PR updates the alerts endpoint to resolve dataset linkage from the new `views` table fields. The endpoint now reads `primary_dataset` and optional `secondary_dataset`, uses `primary_dataset` for alerts data, and uses `secondary_dataset` for Mapeo data when configured. Strict alerts projection and required-column validation remain in place.

The alerts page and dashboard now consume `primary_dataset` and `secondary_dataset` from the API response. `AlertsDashboard` and `useIncidents` use those configured dataset names for record fetching, incident source tables, and export table naming, while still handling a missing secondary dataset gracefully.

UI renders and all alerts related functionality works fine: filtering, downloading, selecting an alert on map.

## What I'm not doing here

- Not implementing Gallery or Map on the new model yet.
- Not changing alert UI behavior beyond dataset wiring.

## LLM use disclosure

GPT 5.5 and Osa


components/AlertsDashboard.vue
```diff
@@ -211,15 +212,17 @@ watch(
     }
     if (!feature) return;
 
-    const isMapeoFeature = isMapeo.value && props.mapeoTable;
+    const isMapeoFeature = isMapeo.value && props.secondaryDataset;
```
IamJeffG: is worth a code comment that:

    // when primaryDataset is alerts, the secondaryDataset (if set) is Mapeo data


(I got this from your PR description, but i think it's good to live here too)

IamJeffG: Another idea to consider -- optionally -- is to just set a `const mapeoTable = props.secondaryDataset` neat the top of this function and then the variable continues to be self-describing (as it was before this PR)

server/api/[table]/alerts.ts
```diff
@@ -179,6 +197,8 @@ export default defineEventHandler(async (event: H3Event) => {
       mapboxStyle: defaultMapboxStyle,
       mapboxBasemaps: basemaps,
       mapboxZoom: Number(tableConfig.MAPBOX_ZOOM),
+      primary_dataset: primaryDataset,
+      secondary_dataset: secondaryDataset,
       mapeoTable,
```
Comment (IamJeffG) [inline server/api/[table]/alerts.ts:202]: Am I reading right that here `secondary_dataset` and `mapeoTable` are equivalent i.e. duplicates?

This is in the same line of thought as my other comment: I feel like the tension is that the database (column name) calls it "secondary_dataset" but all the app code wants it called "mapeoTable". I think you could pick ONE place to rename what we call this thing -- either in fetchViewDatasets or right here -- and just commit to those names upstream and downstream.

server/database/dbOperations.ts
```diff
@@ -575,6 +591,110 @@ export const fetchTableConfig = async (table: string): Promise<ViewConfig> => {
...
+    console.error(
+      `Error fetching ${viewType} datasets for table "${table}":`,
+      error,
+    );
+    throw error;
+  }
+};
+
+/**
+ * Fetches primary and optional secondary datasets in one DB-layer operation.
```
IamJeffG, line 663: A couple nitpicks about having combined these into this function:

* It seems a little misleading: I read this as "we only go to the database once", which is not correct: there are definitely two queries happening inside here.
* it means the pages (the vue components) have to give up a little control as to whether to even make the database fetch for the second dataset: I'm saying this because a real example is that the old code would not fetch the 2nd dataset unless `mapeoCategoryIds` is truthy, but in this new code it does.

Maybe you have a sense for what the following couple PRs will be, and in particular if there will truly be a lot of advantage by having this function -- as opposed to all the vue components making 2 DB calls (as the old code was).

I could easily be swayed to keep this function that combines the 2 DB accesses, if the docstring were more subdued, and give some thought as to whether there really is a need to make the secondaryFetch conditional based on something provided by the caller (secondaryOptions?)

tests/unit/server/alertsEndpointDatasets.test.ts
```diff
@@ -0,0 +1,172 @@
...
+  },
+}));
+
+describe("alerts endpoint dataset config", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+
+    mockFetchViewDatasets.mockResolvedValue({
+      primaryDataset: "alerts_confidential",
+      secondaryDatasets: ["mapeo_data"],
```
IamJeffG, line 86: Is there a similar test where secondaryDataset is null?  (i.e. no mapeo data)

server/api/[table]/alerts.ts
Comment (conservationtimothy) [reply server/api/[table]/alerts.ts:202]: Equivalence is only true for Alerts. For Maps, secondary dataset can be truly ANY dataset that supports maps too. it is a value that we use for 2/3 views we support but only in one view do we know ahead of time what it is, i.e. it'll always be equivalent to whatever `mapeoTable` value.

server/database/dbOperations.ts
conservationtimothy: It is based off this issue: https://github.com/ConservationMetrics/gc-explorer/issues/426

Which is based off of a call Rudo and i had.

tests/unit/server/alertsEndpointDatasets.test.ts
conservationtimothy: Added

PR #473, opened 2026-05-30
Gallery: componentize grid tiles (4 desktop / 2 mobile) — rounded chrome for all media

### LINKED ISSUE #447: Gallery: componentize grid tiles (4 desktop / 2 mobile) — rounded chrome for all media (image + audio)
## Feature Request

Grid tiles must treat **all primary media** the same visually: **images and audio** (and video if in scope) each live inside the same rounded, media-first frame (consistent radius, overflow clipping, min height/aspect rules per design). This is not an “image-only” grid — **audio controls** are first-class in the tile, not an afterthought.

- … (keep 2-col mobile / 4-col desktop, gutters, component extraction)
- **Audio** in a tile uses the same outer rounding and card chrome as **images** (no square unstyled `<audio>` bleeding past corners unless design says otherwise).
- Tile component is the single place that defines “what a gallery cell looks like” for all supported media types.
## Implementation Plan

conservationtimothy:
## Goal

This PR componentizse grid tiles (4 desktop / 2 mobile)for all media and adds rounded chrome. Closes #447 

## Screenshots

Screenshot 2026-05-30 at 14 33 48
Screenshot 2026-05-30 at 14 33 34

## What I changed and why

- Replaced inline `DataFeature` card rendering in the gallery grid with new `GalleryGrid` and `GalleryTile` components, making the layout media-first
- Gallery now uses a consistent 2-column mobile / 4-column desktop grid layout
- `MediaFile` gains a gallery variant that clips images to a square aspect ratio and wraps audio and video in the same rounded violet frame used elsewhere in the app
- Lightbox link is intentionally omitted on grid images in the gallery variant

The old gallery inlined full `DataFeature` cards, so every entry read like a form with metadata and media mixed together which was the intended design before but no longer works for the media-first layout we want: the grid should be a wall of tiles, not a stack of detail cards. Splitting GalleryGrid (layout only) from GalleryTile (one entry, one preview) keeps pagination, filters, and record caching in GalleryView while giving us a single place to evolve tile chrome later (overlay, carousel) without touching page-level logic again.

Using the first attachment only in the grid matches how we already resolve paths via `getFilePathsWithExtension` and avoids cramming multiple files into one cell before the carousel PR exists. A gallery variant on `MediaFile` instead of forking image/audio markup in the tile reuses lazy loading and extension detection, leaves map/alerts on the default variant unchanged, and drops the lightbox wrapper only where the grid will eventually open a detail panel instead of jQuery lightbox. Square aspect-square tiles give a uniform grid on 2- and 4-column breakpoints without the old 1→2→3→4 column ladder, which was inconsistent with the design spec.

## What I'm not doing here

No hover overlay, detail panel, carousel, filter relocation, or e2e lightbox migration. Map and alerts `MediaFile` behavior is unchanged.

## LLM use disclosure

Mostly for unit stuff. Design was all me.

### DIFF
--- a/components/gallery/GalleryGrid.vue
+++ b/components/gallery/GalleryGrid.vue
@@ -1,0 +1,7 @@
+<script setup lang="ts"></script>
+
+<template>
+  <div class="grid grid-cols-2 lg:grid-cols-4 gap-4" data-testid="gallery-grid">
+    <slot></slot>
+  </div>
+</template>
--- a/components/shared/MediaFile.vue
+++ b/components/shared/MediaFile.vue
@@ -92,7 +92,7 @@ const handleImageLoad = () => {
 
 const imageContainerClass = computed(() => {
   if (isGalleryVariant.value) {
-    return "w-full aspect-square overflow-hidden rounded-2xl bg-gray-100 dark:bg-gray-800";
+    return "w-full h-full overflow-hidden rounded-2xl bg-gray-100 dark:bg-gray-800";
   }
   return "w-full aspect-video rounded-lg bg-gray-100 dark:bg-gray-800";
 });
@@ -106,11 +106,14 @@ const imageClass = computed(() => {
 </script>
 
 <template>
-  <div>
+  <div :class="{ 'h-full w-full': isGalleryVariant }">
     <div
       v-if="isImage"
       ref="imageContainer"
-      :class="isGalleryVariant ? '' : 'mb-4'"
+      :class="[
+        isGalleryVariant ? 'h-full w-full' : '',
+        isGalleryVariant ? '' : 'mb-4',
+      ]"
     >
       <!-- Error state: Show red X icon when image fails to load (404) -->
       <div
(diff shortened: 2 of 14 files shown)


conservationtimothy:
p.s this is tested on staging locally.

components/gallery/GalleryGrid.vue
```diff
@@ -0,0 +1,7 @@
+<script setup lang="ts"></script>
+
+<template>
+  <div class="grid grid-cols-2 lg:grid-cols-4 gap-4" data-testid="gallery-grid">
```
rudokemper, line 4: Now here's a nice use of Tailwind that convinces me that it can be a more elegant solution than plain ol' CSS 😉

components/shared/MediaFile.vue
```diff
@@ -4,11 +4,19 @@ import type { AllowedFileExtensions } from "@/types";
...
+  defineProps<{
+    allowedFileExtensions: AllowedFileExtensions;
+    filePath: string;
+    mediaBasePath: string;
+    variant?: "default" | "gallery";
+  }>(),
+  {
+    variant: "default",
+  },
+);
```
rudokemper, line 17: Great, I didn't actually know about this Vue helper.

i18n/locales/en.json
```diff
@@ -91,6 +91,7 @@
   "gallery": "Gallery",
   "galleryEmpty": "There are no items to show in this gallery yet.",
   "galleryNoFilterResults": "No items match the current filters.",
+  "galleryNoMedia": "No media available",
```
rudokemper: This is redundant with `galleryEmpty`. I would use that translation.

rudokemper (approved):
Very good. Only one request on using an existing i18n string. Please make that change before merging.

PR #474, opened 2026-05-31
Gallery: hover/focus overlay — dim tile and “Click to view details”

### LINKED ISSUE #448: Gallery: hover/focus overlay — dim tile and “Click to view details”
## Feature Request

Each gallery tile should invite opening the detail experience: on **hover** (desktop) and a sensible **focus-visible** / tap affordance where needed, the media dims slightly and copy such as **“Click to view details”** appears with a **smooth CSS transition** (keyframes optional for polish).

- Hovering a tile applies a clear but subtle dim overlay; leaving the tile reverses it smoothly.
- Overlay text is readable (contrast, typography) and localized if the app uses `vue-i18n` for similar strings.
- Keyboard users can perceive focus and open details without relying on hover-only behavior.
- Does not block opening details (click target remains the whole tile or a clear primary control).

## Implementation Plan

conservationtimothy:
## Goal

Adds hover/focus overlay when a user hovers over a grid item. Closes #448 

## Screenshots

Screenshot 2026-05-30 at 15 06 11

## What I changed and why

- Added a dim overlay with "Click to view details" text on hover and keyboard focus inside `GalleryTile`, with no changes to `GalleryView` or the grid layout
- Overlay uses `pointer-events-none` so audio and video controls underneath remain usable
- Tile click handling already ignores clicks on media controls, keeping the path clear for open-detail behavior in a follow-on PR

In my head overlay belongs on the tile, not `GalleryView`, because it is per-item chrome tied to the same rounded media frame https://github.com/ConservationMetrics/gc-explorer/pull/473 introduced.Also, `Pointer-events-none` on the overlay avoids the common trap of blocking audio play buttons while still showing the invite-to-detail copy on hover; 

## What I'm not doing here

No detail panel, no `@open` emit wired in `GalleryView`, no carousel.

## LLM use disclosure
Styling largely me, Opus helped with stuff like i18n

### DIFF
--- a/components/shared/MediaFile.vue
+++ b/components/shared/MediaFile.vue
@@ -92,7 +92,7 @@ const handleImageLoad = () => {
 
 const imageContainerClass = computed(() => {
   if (isGalleryVariant.value) {
-    return "w-full aspect-square overflow-hidden rounded-2xl bg-gray-100 dark:bg-gray-800";
+    return "w-full h-full overflow-hidden rounded-2xl bg-gray-100 dark:bg-gray-800";
   }
   return "w-full aspect-video rounded-lg bg-gray-100 dark:bg-gray-800";
 });
@@ -106,11 +106,14 @@ const imageClass = computed(() => {
 </script>
 
 <template>
-  <div>
+  <div :class="{ 'h-full w-full': isGalleryVariant }">
     <div
       v-if="isImage"
       ref="imageContainer"
-      :class="isGalleryVariant ? '' : 'mb-4'"
+      :class="[
+        isGalleryVariant ? 'h-full w-full' : '',
+        isGalleryVariant ? '' : 'mb-4',
+      ]"
     >
       <!-- Error state: Show red X icon when image fails to load (404) -->
       <div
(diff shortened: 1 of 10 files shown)


components/shared/MediaFile.vue
```diff
@@ -92,7 +92,7 @@ const handleImageLoad = () => {
 
 const imageContainerClass = computed(() => {
   if (isGalleryVariant.value) {
-    return "w-full aspect-square overflow-hidden rounded-2xl bg-gray-100 dark:bg-gray-800";
+    return "w-full h-full overflow-hidden rounded-2xl bg-gray-100 dark:bg-gray-800";
```
rudokemper, line 95: Just curious why the change?

rudokemper (approved):
Looks good.

One minor bug that will probably be solved when you work on the next steps is that when you click on a media item, the tile overlay effect persists, even as you hover the mouse over other media items.

Image

components/shared/MediaFile.vue
conservationtimothy: Because I moved the square shape to `GalleryTile` (`aspect-square` on the tile wrapper) and made gallery `MediaFile` fill that box with `h-full` instead.

PR #479, opened 2026-06-03
Use `SlidersHorizontal` Lucide icon for config

### LINKED ISSUE #478: Better icon for config
## Feature Request

@nicopace or @IamJeffG brought up on the GC monthly product call that the current Lucide `database` icon doesn't really represent well the screen to customize views config. I agree.

Some options:

Sliders
Image

`columns-3-cog`
Image

Or search for your own choice: https://lucide.dev/icons/

I personally think the second or fourth slider icon would work. The fourth one seems to imply that it's about customizing a document, or a page, so perhaps more specific than a generic "options" slider, but could also make the user think it's about printing. So maybe a generic horizontal slider is best.

Thoughts welcome - @conservationtimothy feel free to chime in too!

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-explorer/issues/478

## Screenshots

image

### DIFF
--- a/components/shared/AppHeader.vue
+++ b/components/shared/AppHeader.vue
@@ -3,7 +3,7 @@ import type { User } from "@/types";
 import { Role } from "@/types";
 import GlobeLanguagePicker from "@/components/shared/GlobeLanguagePicker.vue";
 import HeaderBrand from "@/components/shared/HeaderBrand.vue";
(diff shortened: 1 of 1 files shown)


nicopace (approved):
Looks good from the code review.

PR #482, opened 2026-06-08
Gallery: per-entry media carousel in grid tiles and in detail view

### LINKED ISSUE #450: Gallery: per-entry media carousel in grid tiles and in detail view
## Feature Request

A single entry may include **multiple** attachments (e.g. several photos **and** audio). Users should be able to **switch between media within the gallery grid tile** (carousel / stepper / dots — match design) **and** see the same multi-media navigation **in the detail view** after opening. “Click to view details” refers to opening the **split detail experience**, not to discovering hidden media.

- In the **grid**, each tile shows one “current” medium at a time; user can move to other media for that entry without opening the detail panel.
- In the **detail** view, the large media region supports the same set of attachments (carousel or equivalent), including **audio** with appropriate controls.
- Opening details (separate from carousel navigation) remains obvious: e.g. overlay copy, dedicated control, or full-tile click excluding carousel controls — **document chosen pattern** to avoid accidental detail opens when swiping arrows.
- Lazy loading / performance: avoid loading every attachment at full resolution for every tile on first paint 
- Keyboard / screen reader: carousel controls are focusable and labeled; detail open action is keyboard-accessible.

## Implementation Plan

conservationtimothy:
## Goal

Closes #450 -- multi-attachment entries get prev/next + dots in grid tiles and in the inline detail panel, without opening detail when using carousel controls.

## Screenshots

Screenshot 2026-06-01 at 17 49 42
Screenshot 2026-06-01 at 17 49 37

Screenshot 2026-06-01 at 17 39 12

## What I changed and why

- Added a shared `GalleryMediaCarousel` component used in both `GalleryTile` and `GalleryDetailPanel` for stepping through multiple attachments
- Only the active slide mounts `MediaFile` (via `:key` on file path) to avoid loading every attachment at full resolution on first paint
- Carousel controls use `@click.stop` and are excluded from the tile's open handler so arrows/dots do not accidentally trigger the detail view
- Fixed Rudo's sticky overlay bug: overlay visibility is now driven by hover and tile-root keyboard focus, carousel controls blur after use, and mouse-leave blurs any focused control inside the tile so "Click to view details" does not persist after carousel interaction

#450 explicitly separates browsing attachments on the tile from opening detail. One shared component keeps grid and detail behavior aligned and avoids duplicating navigation logic. Lazy single-slide rendering matches the acceptance criterion on performance. The overlay fix targets the actual failure mode: carousel buttons took focus and `focus-within` kept the hint visible on the wrong tile -- hover-only overlay plus blur-on-leave matches expected grid UX without breaking keyboard open on the tile itself. Unit tests cover carousel navigation, detail panel wiring, and overlay show/hide on mouseenter/leave.

## What I'm not doing here

- Lightbox vendor removal (separate follow-up issue)

## LLM use disclosure

For tests mostly


rudokemper (approved):
Code and functionality looks good.

Bonus points for a small sub-issue to add key bindings for left and right arrow key, if you feel up for it.

How do you plan to handle e2e tests for this epic?

conservationtimothy:
> How do you plan to handle e2e tests for this epic?

Leaning towards adding a final issue to add E2E tests with the whole picture

PR #493, opened 2026-06-30
Compute coordinates and provide Google link for alert features

### LINKED ISSUE #487: Google Maps link doesn't appear for an alert on the Alerts Dashboard view
On one of our partner's gc-explorer's instance, they can see the alerts but they can't see the google maps link to the alert.

I can see that the code is in the repository:
https://github.com/ConservationMetrics/gc-explorer/blob/main/components/shared/DataFeature.vue#L161

And that there are no conditionals there... so I don't know is making it not appear on this screen.

Image

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-explorer/issues/487

## Screenshots

image

## What I changed and why

Utilize `calculateCentroidFromParsedCoords` and `tryParseDataEntryGeoCoordinates` to calculate `geographicCentroid` for alert features.

## LLM use disclosure

None

### DIFF
--- a/tests/unit/utils/dataTransformers.test.ts
+++ b/tests/unit/utils/dataTransformers.test.ts
@@ -195,6 +195,13 @@ describe("transformAlertEntry", () => {
     expect(result).not.toHaveProperty("g__coordinates");
   });
 
+  it("should compute geographicCentroid for Google Maps link", () => {
+    const result = transformAlertEntry(rawAlert, table);
+    const coordPattern = /^-?\d{1,3}\.\d+, -?\d{1,3}\.\d+$/;
+
+    expect(result.geographicCentroid).toMatch(coordPattern);
+  });
+
   it("should handle GFW records without proprietary fields", () => {
     const gfwAlert = {
       ...rawAlert,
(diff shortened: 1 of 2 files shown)


rudokemper:
fyi @nicopace

tests/unit/utils/dataTransformers.test.ts
```diff
@@ -208,5 +215,6 @@ describe("transformAlertEntry", () => {
     expect(result).not.toHaveProperty("territory");
     expect(result).not.toHaveProperty("t0_url");
     expect(result).not.toHaveProperty("alertAreaHectares");
+    expect(result.geographicCentroid).toMatch(/^-?\d{1,3}\.\d+, -?\d{1,3}\.\d+$/);
```
conservationtimothy: Let's hoist `coordPattern` out of the individual test bodies into a module-level const, and reuse it in both assertions instead of redefining the regex literal in the second `it` block.

conservationtimothy (changes requested):

conservationtimothy (approved):

conservationtimothy (approved):
We should fix the linting before we merge, though.

PR #494, opened 2026-07-01
Switch to database-first DB schema management and test setup runs migrations

### LINKED ISSUE #472: Switch to database-first DB schema management and test setup runs migrations
> I don't understand why we have this function (`ensureViewConfigTableExists`) to create a table. The migration creates the table. How would it not exist by the time our app is running?
>
> I asked Claude and she tells me that in CI, `migrate-and-start.sh` explicitly skips migrations when `CI=true`, and so when tests hit the database ther's no `views` table yet. (I did not check this claim.)
>
> If that's true, it seems to me a better practice is to just run the migrations before starting the app -- every time: when running tests, when running the app locally, and when running in production. That would avoid having to maintain code like this and would also avoid the possibility of divergence between the table created here versus by the migration. IMO the migration is the source of truth, full stop.

 _Originally posted by @IamJeffG in [#470](https://github.com/ConservationMetrics/gc-explorer/pull/470/changes#r3314920394)_

conservationtimothy:
## Goal

Move `gc-exp` to having migrations are the authoritative schema history, they run on every Nuxt/Nitro startup (local dev, CI, production), and tests read real database tables instead of `CI=true` hardcoded bypasses. Closes #472 

## Screenshots

N/A 

## What I changed and why

 Migration `0007` is now safe on fresh databases. `schemas/` was baselined from `drizzle-kit pull` against staging (excluding `metrics` and legacy `view_config`). A trimmed `0008` delta aligns prod drift (`views.view_type` nullable, `views_view_id_idx` with `IF NOT EXISTS` for DBs that already have the index). Migrations run from `server/plugins/migrate.ts` on every startup; Docker no longer runs `drizzle-kit migrate` or skips migrations when `CI=true`. Runtime and test bypasses (`buildCiViewConfigRows`, `ensureViewConfigTableExists`, hardcoded table lists) are removed. Test fixtures live in data-only `guardianconnector.sql`, applied after `migrate()` in CI inside the same plugin so seed never races ahead of DDL.

| SHA | Commit | What changed | Why |
|-----|--------|--------------|-----|
| [`bd6e19a`](https://github.com/ConservationMetrics/gc-explorer/commit/bd6e19af60d640350ed97309cc9a0514d2475121) | fix(db): guard view_config backfill for fresh database migrations | `0007_add_view_config_view_columns.sql` | Fresh test DBs have no `view_config`; unconditional INSERT crashed the migration chain |
| [`b32e73d`](https://github.com/ConservationMetrics/gc-explorer/commit/b32e73d64b9e81a3a4a0c451b56a1a41255b780b) | chore(db): baseline drizzle schema from current database | `viewConfig.ts`, `annotatedCollections.ts`, `package.json` (`db:pull`) | Align Drizzle schema with staging introspection; omitted `metrics` and legacy `view_config` |
| [`3450f0d`](https://github.com/ConservationMetrics/gc-explorer/commit/3450f0d55cbf177e577be268723d50f2448a98de) | chore(db): generate drizzle migration delta | `0008_blue_vanisher.sql`, `meta/0008_snapshot.json`, `_journal.json` | Capture prod drift after baseline; SQL trimmed to `ALTER view_type DROP NOT NULL` + `CREATE UNIQUE INDEX IF NOT EXISTS views_view_id_idx` (generate produced full `CREATE TABLE` without historical snapshots; `IF NOT EXISTS` for staging DBs that already had the index) |
| [`a405d2f`](https://github.com/ConservationMetrics/gc-explorer/commit/a405d2f0ee256f12ada2a5930589f838e083dc70) | refactor(db): run drizzle migrations from nitro startup | `server/plugins/migrate.ts`, `migrate-and-start.sh`, `Dockerfile`, `README.md` | App owns migrations everywhere; removed `drizzle-kit` from prod image and CI migration skip |
| [`242e903`](https://github.com/ConservationMetrics/gc-explorer/commit/242e90349a56028c17984ac0ccdd7ffd7f32396e) | refactor(db): use migrated tables in tests and runtime | `dbOperations.ts`, `server/utils/index.ts`, `config/index.get.ts` | Remove `CI=true` DB bypasses; tests and runtime use real migrated tables |
| [`440d791`](https://github.com/ConservationMetrics/gc-explorer/commit/440d791f56f8c936326ff9d1c522dc6002c32db8) | test(db): seed view fixtures after startup migrations | `guardianconnector.sql`, `migrate.ts`, `dbConnection.ts`, `docker-compose.tests.yml` | Data-only fixtures (from old `buildCiViewConfigRows` JSON); seed after migrate in same plugin; warehouse-only Postgres init |

I ran a fresh local `docker compose -f docker-compose.tests.yml up` after building the image: all nine migrations applied (`0000`–`0008`), five `views` rows and two `public_views` rows matched the former CI fixtures, warehouse tables (`bcmform_responses`, `fake_alerts`, `seed_survey_data`) were present, the app returned HTTP 200, and all 410 unit tests passed. That confirmed migrate → seed ordering works. The `0007` guard is safe on empty DBs and unchanged on prod where `view_config` exists. `0008` is a small delta; `IF NOT EXISTS` on the index lets it apply cleanly against staging where `views_view_id_idx` already existed.

Also ran a manual check in dev to ensure the app starts up, applies migrations and works correctly. 

## What I'm not doing here

- `metrics` table — exists on shared DB, not relevant to GC Explorer
- `gc_settings` — owned by gc-landing-experiment; defer until Explorer needs it

## LLM use disclosure

 Claude (via Cursor). All generated SQL and code was reviewed before commit. `drizzle-kit pull` against staging was run manually by me (Osa).


server/plugins/migrate.ts
```diff
@@ -0,0 +1,20 @@
...
+  });
+
+  if (process.env.CI !== "true") return;
+
+  const seedPath = resolve("tests/db-seed/guardianconnector.sql");
+  if (!existsSync(seedPath)) return;
+
+  const seedSql = readFileSync(seedPath, "utf-8");
+  await configPg.unsafe(seedSql);
+});
```
IamJeffG: Strongly suggest keeping the CI-only stuff all nested together inside a context:




```suggestion
  if (process.env.CI === "true") {

    const seedPath = resolve("tests/db-seed/guardianconnector.sql");
    if (!existsSync(seedPath)) return;

    const seedSql = readFileSync(seedPath, "utf-8");
    await configPg.unsafe(seedSql);
  }
});
```

However I still have some concerns:

1. I believe the CI variable could be a number of different things besides the exact string "true".
2. I typically see test data being inserted as part of the test harness, not the migration scaffolding.
    - test seed data is not a migration
    - what if I wanted to run tests outside of CI? I should be able to run tests outside of CI, and when I do that, it should not set CI=true (as `docker-compose.tests.yml` currently does)

IamJeffG (commented):
This looks pretty good!  I worry about the if/else logic in migrations, but I think I understand why you did it. It just gives me pause because we don't really know if this will apply cleanly to all our deployments.

I don't accept test running being coupled to CI=true (which means I don't like seed data for testing being coupled to CI=true); and I don't like seed data for testing being part of the migrations scaffolding.

Might also be worth one last look at the README to make sure we are communicating correctly and clearly.  Maybe it is, but `test:unit` is described in a way that excludes a running DB, whereas  `test:e2e` seems to imply more than that.

.github/workflows/docker-publish.yml
```diff
@@ -78,21 +78,14 @@ jobs:
...
-          BASE_URL=http://localhost:8080
           MAPBOX_ACCESS_TOKEN=${{ secrets.MAPBOX_ACCESS_TOKEN }}
           MEDIA_BASE_PATH=${{ secrets.MEDIA_BASE_PATH }}
           PLANET_API_KEY=${{ secrets.PLANET_API_KEY }}
-          # Database connection details (matching docker-compose)
-          NUXT_DB_HOST=database
-          NUXT_DB_PORT=5432
-          NUXT_DB_USER=testuser
-          NUXT_DB_PASSWORD=testpassword
-          NUXT_DATABASE=test_warehouse
```
conservationtimothy, line 95: This was an intentional change to simply this file.

server/plugins/migrate.ts
conservationtimothy: Thanks for the review and I agree with ur thoughts. 

Went ahead and pushed some follow-ups. `migrate.ts` now only migrates - Drizzle runs at Nitro startup, nothing else. `NUXT_TEST` replaces `CI` for the test stack so docker-compose no longer sets `CI=true`; local docker E2E is `NUXT_TEST=true pnpm test:e2e` and `CI` stays for what GHA sets automatically (Playwright skips `pnpm dev`). Fixtures live in the test harness now too - `guardianconnector.sql` is seeded by Playwright `globalSetup` after the backend is healthy, not at migration time. The `NUXT_TEST` parsing shared helper accepts `true` case-insensitively and throws on unrecognized values so a typo doesn't silently half-enable test mode. README is also updated with a "what runs where" table.

Can be found https://github.com/ConservationMetrics/gc-explorer/pull/494/commits/ffeb7f9ad8f6e7aabcbbad973c2739809c56770e

IamJeffG (approved):
One concern is I see some warnings in the test logs -- were those already there before this PR (i.e. on `main`?)  https://github.com/ConservationMetrics/gc-explorer/actions/runs/28583439643/job/84749209122?pr=494

But LGTM.  Will you be able to babysit deployment to all our VMs once the time is right for that? (perhaps not til after further work on multi-dataset views epic)

conservationtimothy:
> One concern is I see some warnings in the test logs -- were those already there before this PR (i.e. on `main`?) https://github.com/ConservationMetrics/gc-explorer/actions/runs/28583439643/job/84749209122?pr=494
> 

They predate this PR, yes. [See](https://github.com/ConservationMetrics/gc-explorer/actions/runs/28358389327/job/84006945804?pr=490)

conservationtimothy:
> . Will you be able to babysit deployment to all our VMs once the time is right for that? (perhaps not til after further work on multi-dataset views epic)

Yeah, I'd be more than happy to.

Issue #486, opened 2026-06-11
Single source of truth: MAPEO_TABLE → secondary_dataset column


conservationtimothy:
## Feature Request

Make `views.secondary_dataset` the sole store for an alerts view's Mapeo table name. Today the same value lives in two places: the `secondary_dataset` column (derived on write via `deriveSecondaryDataset`) and `MAPEO_TABLE` inside the `view_config` JSON. They cannot drift, but every reader still uses the JSON field.

**Related:** per-view config normalization (PR #470 / `feature/single-table-view-config-model`). Out of scope for that PR by design; tracked here.

### User Stories & Acceptance Criteria

- [ ] **Write path:** On config create/update, persist the Mapeo table name only in `views.secondary_dataset` for alerts views. Strip `MAPEO_TABLE` from stored `view_config` JSON (same pattern as stripping `VIEWS`).
- [ ] **Read path (server):** Config/data endpoints that need the Mapeo table read `secondary_dataset` from the view row, not `MAPEO_TABLE` from JSON. Reconstruct `MAPEO_TABLE` on the in-memory `ViewConfig` object returned to clients if backward compatibility is still needed during transition.
- [ ] **Read path (client):** Config edit UI and alerts flow (`server/api/[table]/alerts.ts`, `ConfigAlerts.vue`) consume the Mapeo table from the normalized field once the API exposes it (may overlap with exposing `primary_dataset` / `secondary_dataset` on config API responses).
- [ ] **Migration:** New migration strips `MAPEO_TABLE` from existing `view_config` JSON where `secondary_dataset` is already populated (alerts rows only).
- [ ] **Types/docs:** Remove or deprecate `MAPEO_TABLE` on `ViewConfig` once all readers are migrated. Remove `TODO(single-source-of-truth)` comments at:
  - `server/database/dbOperations.ts` (`deriveSecondaryDataset`)
  - `server/api/[table]/alerts.ts`
  - `types/index.ts`
  - `server/database/migrations/0007_...sql` (historical comment only; new migration does the strip)

### UI Sketch (or Screenshot)

N/A — backend/data-model change; config UI may need to bind Mapeo field to `secondary_dataset` once exposed.

## Implementation Plan

1. Expose `secondary_dataset` on config read API responses (`ViewConfigRow` already has it; wire through endpoints the client uses).
2. Update alerts endpoint and config edit UI to read from column (or reconstructed field).
[...]

PR #495, opened 2026-07-06
use views dataset fields for alerts primary and secondary data

conservationtimothy:
## Goal

Re-implement exposing the alerts view primary and secondary datasets from the typed `views` table columns, and make the alerts client consume those API-returned dataset names before continuing the stacked dataset API work.

## Screenshots

Screenshot 2026-07-06 at 11 42 31
Screenshot 2026-07-06 at 11 41 29

## What I changed and why

- Added typed dataset helpers for view rows
- Alerts endpoint now resolves `primary_dataset` and `secondary_dataset` from the `views` table and includes both fields in the API response, removing the runtime dependency on `MAPEO_TABLE` in the config JSON
- Alerts endpoint uses `primary_dataset` for alert data and `secondary_dataset` for Mapeo data
- Alerts page and dashboard thread the returned `primary_dataset` and `secondary_dataset` through to record fetching, export naming, and incidents source-table selection
- Added unit coverage for the new dataset helpers and updated the dashboard component test fixture to match the new props

The implementation follows the existing database model: route `[table]` remains the primary dataset identifier, and `fetchViewDatasets(table, "alerts")` keys the lookup on `(view_type, primary_dataset)`. The endpoint still loads the existing alerts config for permissions and display settings, but the dataset tables used for reads now come from the typed columns that were added for this purpose.

I verified the new helper behavior covers primary/secondary hits, null secondary datasets, missing view rows, primary-only reads, and primary-plus-secondary reads. I also reran the alerts dashboard component tests to make sure the new props did not regress dashboard mounting or incident UI behavior.

## What I'm not doing here

This PR does not remove `MAPEO_TABLE` from stored config JSON (the to-do) or change the config editing UI. That cleanup belongs in the later stacked PR after gallery and map have been implemented

## LLM use disclosure

GPT 5.5 and Osa


server/api/[table]/alerts.ts
```diff
@@ -110,28 +133,15 @@ export default defineEventHandler(async (event: H3Event) => {
       ),
     };
 
-    // MAPEO_TABLE is read from the config JSON, which is also mirrored into the
-    // views.secondary_dataset column on save. This reader is the reason that column
-    // is not yet the single source of truth. TODO(single-source-of-truth): once the
-    // API returns secondary_dataset (and MAPEO_TABLE is stripped from the JSON),
-    // read it from the column instead. See deriveSecondaryDataset in dbOperations.
```
Comment (conservationtimothy) [inline server/api/[table]/alerts.ts:117]: Comment is removed because further down the line (within a day or 2), as part of this work a PR will be introduced to fix this and have a [single source of truth](https://github.com/ConservationMetrics/gc-explorer/issues/486).

IamJeffG:
> The implementation follows the existing database model: route [table] remains the primary dataset identifier, and fetchViewDatasets(table, "alerts") keys the lookup on (view_type, primary_dataset)

At some point we should revisit this decision.  We went along with it because it was the initially written plan, but at this point the real noun has become [view].

For example think of the use case where you want 2 views, both with the same primary dataset, but different secondary datasets.

server/database/dbOperations.ts
```diff
@@ -357,6 +358,76 @@ export const fetchData = async (
   return { mainData, columnsData, metadata };
 };
 
+type FetchedDatasetData = Awaited<ReturnType<typeof fetchData>>;
+
+/**
+ * Fetches the typed dataset columns for one configured view.
```
IamJeffG: The whole PR -- and the crux of it is here -- makes things hard to follow, since it tends to use the terms "view", "dataset" and "table" interchangeably, at least some of the time.

In fact, if we are going to use the word "dataset" at all, then I'd agree there's a 1:1 relationship between "table" (the string name) and "dataset" (the rows), but we don't use it that way. This function's name is an example: the actually fetches the table names, not data.  One solution could be either name this "fetchViewTables" and the other function "fetchViewData"; another solution would be to just stop using the word "table" (or "dataset"  - pick one).

This conflation of terms might have existed already and therefore is outside the scope of this PR. I didn't check. In that case, we can leave it.

IamJeffG: Line 366: "route identifier" is introducing yet another concept _that I don't think this function actually knows about._

server/api/[table]/alerts.ts
```diff
@@ -59,34 +60,56 @@ export default defineEventHandler(async (event: H3Event) => {
...
+      primaryDataset,
+      {
+        secondaryDataset: shouldFetchMapeoData ? secondaryDataset : null,
+        primaryOptions: {
+          limit,
+          mainColumns: alertsMainProjection,
+          includeMetadata: alertsMetadataProjection.length > 0,
+          metadataColumns: alertsMetadataProjection,
+        },
+        secondaryOptions: {
```
Comment (IamJeffG) [inline server/api/[table]/alerts.ts]: is this an unlimited fetch, i.e. all rows in the table?

rudokemper:
Turned that into an issue: https://github.com/ConservationMetrics/gc-explorer/issues/498

IamJeffG (approved):
Approval as long as we aren't doing full-table retrieval (I'm looking at the secondaryDataset, see my comment).  If we are, and you think that's necessary, let's talk.

server/api/[table]/alerts.ts
Comment (conservationtimothy) [reply server/api/[table]/alerts.ts]: No, it is not unlimited. `fetchData()` applies `options.limit ?? runtimeConfig.public.rowLimit`, so omitting `limit` on `secondaryOptions` still uses the configured row limit.

server/database/dbOperations.ts
conservationtimothy: I agree this is hard to follow. Some of the table/dataset conflation already exists in the repo: routes use `[table]`, the config helpers are named `fetchTableConfig`, but the new schema columns are `..._dataset`. I will rename to (hopefully) be more explicit.

PR #496, opened 2026-07-06
use views dataset fields for gallery primary data

conservationtimothy:
## Goal

Route the gallery view through the typed `views.primary_dataset` column and expose that dataset in the gallery API response as the next stacked step after the alerts dataset API work.

## Screenshots
Screenshot 2026-07-06 at 11 56 53

## What I changed and why

- Gallery endpoint now resolves its configured dataset via `fetchViewDatasets(table, "gallery")`, reads records from the returned primary dataset, and includes `primary_dataset` in the API response
- No secondary dataset logic added, keeping the change focused on primary dataset plumbing
- Gallery page now passes the API-returned primary dataset into `GalleryView` so full-record cache lookups use the typed dataset column rather than assuming the route param is the fetch target
- Added server-side test coverage for the gallery endpoint dataset behavior

The change follows the same route-key invariant as the alerts [PR](https://github.com/ConservationMetrics/gc-explorer/pull/495): `[table]` identifies the configured view row, and the actual warehouse table used for reads comes from `views.primary_dataset`. The endpoint still uses the existing gallery config for permissions and display/filter settings, but the data source now comes from the typed column that the view model owns.

## What I'm not doing here

This PR does not touch alerts, map, or anything beyond what is necessary

## LLM use disclosure

GPT 5.5 and Osa

### DIFF
--- a/tests/unit/server/galleryEndpointDatasets.test.ts
+++ b/tests/unit/server/galleryEndpointDatasets.test.ts
@@ -1,0 +1,111 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import galleryHandler from "@/server/api/[table]/gallery";
+
+type GalleryRouteEvent = {
+  context: { params: { table: string } };
+};
+
+type GalleryRouteResponse = Record<string, unknown>;
+
+type GalleryRouteHandler = (
+  event: GalleryRouteEvent,
+) => Promise<GalleryRouteResponse>;
+
+const handleGalleryRequest = galleryHandler as unknown as GalleryRouteHandler;
+
+const hoisted = vi.hoisted(() => {
+  Object.assign(globalThis, {
+    defineEventHandler: (handler: unknown) => handler,
+    useRuntimeConfig: () => ({
+      public: {
+        allowedFileExtensions: {
+          audio: [],
+          image: ["jpg"],
+          video: [],
+        },
+      },
+    }),
+  });
+
+  return {
+    fetchData: vi.fn(),
+    fetchTableConfig: vi.fn(),
+    fetchTableSqlColumns: vi.fn(),
+    fetchViewDatasets: vi.fn(),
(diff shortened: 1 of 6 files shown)


IamJeffG (approved):

tests/unit/server/galleryEndpointDatasets.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  filterUnwantedKeys: hoisted.filterUnwantedKeys,
+}));
+
+vi.mock("@/server/utils/dbHelpers", () => ({
+  parseAndValidateLimit: hoisted.parseAndValidateLimit,
+}));
+
+vi.mock("@/utils/accessControls", () => ({
+  validatePermissions: hoisted.validatePermissions,
+}));
```
IamJeffG, line 63: At some point, we mock so much that the test is not testing very much at all.

Have you considered the pros and cons of (and what the test suite might look like if) this suite doesnt mock all this, but instead just inserts some test rows in the DB and hits the DB?

conservationtimothy: You’re right. Looking at the existing Playwright setup, this would fit better as an API-level test than a mock-heavy unit test.

The test could use the existing seeded `views` rows in `guardianconnector` and warehouse tables from `tests/db-seed/warehouse.sql`, then call something like `/api/seed_survey_data/gallery` with `page.request.get(...)`. The assertions would be on behavior: the response includes `primary_dataset`, `table` matches that resolved primary dataset, and the returned data shape comes from the configured view table.

That would avoid mocking `fetchViewTables`, `fetchData`, and the filters, and would prove the real config DB + warehouse DB path. I think that’s a better shape for this coverage than the current interaction test.

conservationtimothy: Created an [issue](https://github.com/ConservationMetrics/gc-explorer/issues/502).

PR #503, opened 2026-07-08
Single source of truth: MAPEO_TABLE → secondary_dataset column  

### LINKED ISSUE #486: Single source of truth: MAPEO_TABLE → secondary_dataset column

conservationtimothy:
## Goal

Make `views.secondary_dataset` the stored source of truth for an alerts view’s Mapeo table name, while keeping the existing config UI working during the transition. Closes #486 

## Screenshots

N/A

## What I changed and why

- Config writes now accept `MAPEO_TABLE` from the alerts config form, derive `views.secondary_dataset` from it, then strip `MAPEO_TABLE` before serializing the JSON config
- Config reads reconstruct `MAPEO_TABLE` from `secondaryDataset` for existing API/UI compatibility, avoiding a larger form rewrite in this PR
- Added migration `0009_strip_mapeo_table_from_view_config` that removes `MAPEO_TABLE` from stored alert `view_config` JSON where `secondary_dataset` is already populated, without touching the actual Mapeo table value
- `MAPEO_TABLE` is no longer persisted inside `views.view_config`

I convinced myself this is right by checking the flow from both sides. When an admin saves the alerts config, the Mapeo table name is still accepted from the existing form, but it is saved only to `secondary_dataset`. When the config is loaded again, the UI still receives the value it expects, but it is rebuilt from `secondary_dataset` instead of being read from duplicated JSON.

## What I'm not doing here

This PR does not redesign the config UI which we've spoken on regularly as a blocker. 

## LLM use disclosure

GPT 5.5

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+server/database/migrations/meta/*_snapshot.json linguist-generated
--- a/server/database/dbOperations.ts
+++ b/server/database/dbOperations.ts
@@ -71,31 +71,49 @@ const createDuplicateViewError = (table: string, viewType: ViewType) => {
 };
 
 /**
- * Derives a view's secondary dataset (its Mapeo table) from its config. This is
- * the single source for the `MAPEO_TABLE` → `secondary_dataset` mapping, shared by
- * every place that builds a view row (live writes and the test fixture seed) so the column
- * can never drift from the config field it mirrors: only alerts views have a
- * secondary dataset, and a blank value normalizes to null.
- *
- * TODO(single-source-of-truth): `MAPEO_TABLE` and `secondary_dataset` currently hold
- * the SAME fact in two places — the value stays in the view_config JSON AND is copied
- * into the secondary_dataset column here. They cannot drift (the column is always
- * derived from `MAPEO_TABLE` via this one helper, and nothing writes the column
- * independently), but today the column is effectively a write-only mirror: every
- * reader (alerts.ts, the config edit UI) still reads `MAPEO_TABLE` from the JSON.
- * The follow-up that returns primary/secondary_dataset from the API should make the
- * column the only source: stop persisting `MAPEO_TABLE` in the JSON (strip it like
- * `VIEWS`) and reconstruct it from the column on read for backward compatibility.
+ * Derives the secondary_dataset column from an incoming alerts config payload.
  *
  * @param {ViewType} viewType - The view's type.
- * @param {ViewConfig} config - The view's settings JSON.
+ * @param {ViewConfig} config - The submitted view settings.
  * @returns {string | null} The secondary dataset, or null when not applicable.
  */
-const deriveSecondaryDataset = (
+const deriveSecondaryDatasetFromConfig = (
   viewType: ViewType,
   config: ViewConfig,
 ): string | null =>
   viewType === "alerts" ? config.MAPEO_TABLE?.trim() || null : null;
+
+/**
+ * Removes fields that are stored in typed view columns rather than view_config.
+ *
+ * @param {ViewConfig} config - Submitted view settings.
+ * @returns {ViewConfig} Settings safe to serialize into views.view_config.
+ */
+const toStoredViewConfig = (config: ViewConfig): ViewConfig => {
+  const storedConfig = { ...config } as ViewConfig & { VIEWS?: unknown };
+  delete storedConfig.MAPEO_TABLE;
+  delete storedConfig.VIEWS;
+  return storedConfig;
+};
+
+type ViewConfigSourceRow = Pick<
+  typeof viewConfig.$inferSelect,
+  "secondaryDataset" | "viewConfig" | "viewType"
+>;
+
+/**
+ * Parses stored config and reconstructs transient client fields from view columns.
+ *
+ * @param {ViewConfigSourceRow} row - View row from the config database.
+ * @returns {ViewConfig} Config object used by existing clients.
+ */
+const parseViewConfigRowConfig = (row: ViewConfigSourceRow): ViewConfig => {
+  const parsedConfig = JSON.parse(row.viewConfig) as ViewConfig;
+  if (row.viewType === "alerts" && row.secondaryDataset) {
+    return { ...parsedConfig, MAPEO_TABLE: row.secondaryDataset };
+  }
+  return parsedConfig;
+};
 
 /**
  * Checks whether a given table exists in the warehouse schema.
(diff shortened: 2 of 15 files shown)


types/index.ts
```diff
@@ -45,10 +45,7 @@ export interface ViewConfig {
   MAPBOX_BASEMAPS?: string; // JSON string of BasemapConfig[]
   MAPBOX_ZOOM?: number;
   MAPEO_CATEGORY_IDS?: string;
-  // Mirrored into the views.secondary_dataset column on save (see
-  // deriveSecondaryDataset). Not yet a single source of truth; readers still use
-  // this field. TODO(single-source-of-truth): drop once the API returns/consumes
-  // secondary_dataset.
+  // Config API/UI field backed by views.secondary_dataset.
   MAPEO_TABLE?: string;
```
IamJeffG: should this line be deleted?

IamJeffG: Oh maybe not. This isn't the type of the JSONB, it's the type of something passed around at a higher level, right?

For example a request payload from the frontend?

server/database/migrations/meta/0009_snapshot.json
IamJeffG, line 1: I have not seen snapshot files getting committed for previous migrations. Admittedly I've not used DrizzleORM. Is this supposed to live in Git?  I don't immediately see why it would be necessary.

server/database/dbOperations.ts
```diff
@@ -71,32 +71,49 @@ const createDuplicateViewError = (table: string, viewType: ViewType) => {
...
+ * @returns {ViewConfig} Config object used by existing clients.
+ */
+const parseViewConfigRowConfig = (row: ViewConfigSourceRow): ViewConfig => {
+  const parsedConfig = JSON.parse(row.viewConfig) as ViewConfig;
+  if (row.viewType === "alerts" && row.secondaryDataset) {
+    return { ...parsedConfig, MAPEO_TABLE: row.secondaryDataset };
+  }
+  return parsedConfig;
+};
+
```
IamJeffG: This is still a lot of helpers for what I think of as supporting an old data model.  What is still producing `row`s in this format? Can we fix it there too?

server/database/migrations/meta/0009_snapshot.json
conservationtimothy: 0008....json is an example. It is the newer way Drizzle does snapshots. We should ideally commit from this forum: https://www.answeroverflow.com/m/1394856908242026566

> Don't put anything from the drizzle folder into gitignore.
The migration files, the snapshots and the journal are all necessary to apply the changes to the DB.
The migrations are typically applied in the production start script by calling migrate before starting the server with node.
The migration history is stored in each DB, it decides which migrations will be applied.
Push the migration changes along with the schema changes.

types/index.ts
conservationtimothy: Yes, it is passed from the client -> server. It maps to the mapeo table input field to be specific.

server/database/migrations/meta/0009_snapshot.json
conservationtimothy: There is an issue on this, lots of upvotes: https://github.com/drizzle-team/drizzle-orm/issues/5528 
https://github.com/drizzle-team/drizzle-orm/issues/5635#issuecomment-4640145933

server/database/dbOperations.ts
conservationtimothy: Config UI no longer sends `MAPEO_TABLE`. It still labels the field as the "Mapeo table" because that is what admins are editing, but it now submits that value separately as `secondaryDataset`

That let us delete the extra dbOperations helpers entirely

server/database/migrations/meta/0009_snapshot.json
IamJeffG: Makes sense, thanks for the pointers. I can get behind committing snapshot + journal + SQL together.

Do you know anything about [adding `linguist-generated=true` metadata](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github) so these don't explode PR diffs?  I think it would be adding  `server/database/migrations/meta/*.json` to .gitattributes.  Could be worth exploring (I'm even ok with it being in this PR if you want and if you think it's correct and helpful).  Not urgent.

IamJeffG (approved):

server/database/migrations/meta/0009_snapshot.json
conservationtimothy: I don't but I am happy to acquaint myself and add it in this PR. Next opportunity to see it in action might be weeks away so worth tackling now.

conservationtimothy: Added the `.gitattributes` rule.

.gitattributes
```diff
@@ -0,0 +1 @@
+server/database/migrations/meta/*_snapshot.json linguist-generated
```
IamJeffG, line 1: Probably the entire directory is generated (in particular, journal.json)

But yeah, this really improves the "files changed" tab of viewing a PR! Success!

Issue #500, opened 2026-07-08
Add search and view type filter to config page


rudokemper:
## Feature Request

As we now disentangle view types from primary datasets, it would be good to be able to search and filter views on the `/config` page.

Let's use the same search bar and filter assets that we provide on the index page:

<img width="1094" height="189" alt="Image" src="https://github.com/user-attachments/assets/8ced1e3a-a417-4f95-ac3d-297734b51ce9" />

If these are not already shared components (e.g. provided in `/components/shared/` for reuse), let's turn them into one.

## Implementation Plan
(may be added later, before work begins)

Issue #499, opened 2026-07-08
Replace dataset-centric view creation workflow


rudokemper:
## Feature Request

Parent: [https://github.com/ConservationMetrics/gc-explorer/issues/437](https://github.com/ConservationMetrics/gc-explorer/issues/437)

The previous flow for adding a dataset view was/is:

<img width="864" height="220" alt="Image" src="https://github.com/user-attachments/assets/ada49c83-7f23-487a-89d1-faa9c3672af0" />

```text

Add Dataset: select table

 └── Configure

      ☑ Map

      ☑ Gallery

      ☑ Alerts

```

This worked because there was a 1:1 relationship between `(view_type, dataset)`.

Since [https://github.com/ConservationMetrics/gc-explorer/issues/193](https://github.com/ConservationMetrics/gc-explorer/issues/193) introduced secondary datasets, a view is now its own entity. A map is no longer simply "the map for dataset X." It is a map that uses dataset X as its primary dataset, optionally dataset Y as a secondary dataset, along with its own styling and configuration.

As [https://github.com/ConservationMetrics/gc-explorer/issues/498](https://github.com/ConservationMetrics/gc-explorer/issues/498) points out, the database still enforces a unique `(view_type, primary_dataset)` combination. That means we cannot yet create multiple map views for the same primary dataset.

This proposal updates the UI to match the new view-centric model while keeping the current database restriction. Users choose a view type first, then configure the view. When #498 is implemented, the duplicate restriction can be removed without changing the overall flow.

## User Stories & Acceptance Criteria

* User selects a view type first, then is taken to the create form for that type.

* The create URL supports pre-filling the view type and optionally the primary dataset.

* A new view is only created when the user clicks **Save**.

* On the create form, an AJAX check runs whenever the primary dataset changes. If a view with the same `(view_type, primary_dataset)` already exists, show a warning with a link to the existing view.

* The duplicate check only exists on the create form, so deep links are handled correctly even if no primary dataset was selected beforehand.

* Cancel leaves the database unchanged.

* Edit uses the same form on a different route (identified by view slug). Immutable fields such as the primary dataset are disabled with a short explanation.

## UI Sketch

```text

[+ Add view]

  → Page 1: Select view type (+ optional primary)

  → Page 2: /views/new?view_type=map&primary=...

      ↓

     Save

      ↓

Create DB row + slug

```

On Page 2, when the primary dataset changes:

```text

Primary dataset: CoMapeo_data



[...]

PR #508, opened 2026-07-13
Move view metadata (including view type) outside the configuration panel

### LINKED ISSUE #501: Move view metadata (including view type) outside the configuration panel
## Feature Request

The scope of this sub-issue bleeds a bit into https://github.com/ConservationMetrics/gc-explorer/issues/435, but I think it best housed under this "UI changes for multiple view" epic for now.

The way we currently show what kind of View a dataset when configuring is a little deceiving:

Image

With the top-right arrow toggle behavior, and being included in the overall frame of the config params component, it makes the user feel that it is something to be configured, especially since this was the hitherto behavior. (Here, a user could use checkboxes to select the type of view it is.) Now, a view type is static and immutable after view creation.

I think we should do away with this `ConfigViews` component altogether and specify the type of view that it is outside of the `ConfigCard` component.

Instead, the page header (or a small metadata section immediately beneath it) should communicate immutable information about the view, for example:

- **View**: Gallery
- **Name**: Biodiversity Gallery
- **Primary** dataset: bcmform_responses
- **Secondary** datasets: None

This better reflects the new mental model where a View is the primary resource and the configuration page is editing the properties of that specific view, not deciding what kind of view it should be.

The `ConfigCard` should then contain only settings that are actually configurable for that view. Removing the `ConfigViews` component also eliminates the visual implication that the view type can be changed after creation.

## Implementation Plan

conservationtimothy:
## Goal

Make immutable view identity (type, name, primary/secondary datasets) visible as read-only page metadata on `/config/[dataset]`, instead of burying view type inside a collapsible “Views” section that looks configurable. Closes #501 

## Screenshots
Screenshot 2026-07-12 at 22 36 49
Screenshot 2026-07-12 at 22 36 39

## What I changed and why

- Moved view type out of the editable `ConfigCard` form and into a read-only metadata block under the page header, making it clear that view type is fixed after creation
- Metadata block displays view type (icon + localized label), display name (`viewName`), primary dataset, and secondary dataset (or "None")
- Name and description remain editable inside `ConfigCard` as before; the metadata block only mirrors saved identity fields and refreshes the display name after a successful save
- Deleted `ConfigViews.vue` since the Views collapsible section is no longer needed
- Updated e2e expectations to assert the header metadata block instead of the Views collapsible
- Added i18n keys for the new metadata labels across all six locales

 The page already had the resolved view row (including `viewName` and `secondaryDataset`); the metadata block reads those values and keeps the same `config-view-type-display` test id so the type badge stays findable.

## What I'm not doing here

- Search / view-type filter on `/config` (#500)
- Type-first create flow (`/config/new`, #499 )
- Changing primary/secondary semantics or edit route shape

## LLM use disclosure

Cursor Grok 4.5 helped with translations


rudokemper:
Thanks @conservationtimothy, before doing code review I checked out the UI in development and had some thoughts:

image

(1) can be the **Dataset Display Name** (e.g. "BCM Form Responses - Gallery" in this example) instead of the primary dataset repeated
(2) In that case, we don't need this "Name" metadata field
(3) Instead of repeating the primary dataset here, can we provide a friendly name like "Configuration Options" (localized)?
(4) Kind-of out of scope, but I wouldn't mind you fixing this here: Can the copy instead say "Copy Config from Another View"

Lastly, can the **page title** be `GuardianConnector Explorer: Configuration - {dataset_name}` instead of `GuardianConnector Explorer: Configuration - {primary-dataset}`?

conservationtimothy:
@rudokemper , sure ! But hold off on code review till I've implemented your suggestions. 

On 1 and 5: Yes but will fallback to primary dataset *if* no display name has been configured. 

Everything else i agree with.

rudokemper (changes requested):
image

I think there is a lot of unnecessary spacing between these 3 metadata items.

Can it be 3 cols that collapse responsively, and also not so much space between them?

conservationtimothy:
Screenshot 2026-07-16 at 17 03 35
Screenshot 2026-07-16 at 17 04 00

@rudokemper Done

rudokemper (approved):

PR #512, opened 2026-07-15
Add search and view type filter to config page

### LINKED ISSUE #500: Add search and view type filter to config page

conservationtimothy:
## Goal

Reuse the home page search bar and view-type filter on `/config` so dataset views are easier to find as the list grows. Closes #500 

## Screenshots

Screenshot 2026-07-15 at 10 59 40
Screenshot 2026-07-15 at 10 59 33
Screenshot 2026-07-15 at 10 59 30
Screenshot 2026-07-15 at 10 59 23
Screenshot 2026-07-15 at 10 59 18
Screenshot 2026-07-15 at 10 59 10

## What I changed and why

- Extracted `SearchBar` and `ViewTypeFilter` into shared components backed by a small pure filter helper, eliminating duplicated markup between the home and config dashboard pages
- Home page behavior is unchanged: URL sync stays in the page and search still matches display name and description
- Config dashboard now uses the same controls to filter cards by view type and search `viewName` and `primaryDataset`, with matching query-param deep links, and the Add button sits in the filter row
- Added unit tests for the filter helpers and e2e coverage asserting search, empty results, view-type filtering, and URL persistence on the config dashboard
- 

The shared components only expose `v-model` / props, so the filtering and URL sync remain in the page/dashboard where they already lived on home and config mirrors that pattern. The unit tests execute the match helpers with empty queries, case-insensitive substrings, and single vs multi type lists, which is the same logic both UIs call. The new config e2e tests exercise the end behavior a user sees: typing a card name reduces the grid, gibberish shows the no-results copy, Map filter leaves only map-tagged cards, and reloading `/config?q=` / `?view=` restores the controls.

## What I'm not doing here

- Type-first create flow (#499)

## LLM use disclosure

Cursor Grok 4.5 helped with tests

### DIFF
--- a/components/ConfigDashboard.vue
+++ b/components/ConfigDashboard.vue
@@ -2,6 +2,9 @@
 import { useI18n } from "vue-i18n";
 import type { ViewConfigRow, ViewType } from "@/types";
 import { CONFIG_LIMITS } from "@/utils";
+import SearchBar from "@/components/shared/SearchBar.vue";
+import ViewTypeFilter from "@/components/shared/ViewTypeFilter.vue";
+import { matchesSearchQuery, matchesViewTypeFilter } from "@/utils/viewFilters";
 import {
   ChevronLeft,
   Images,
@@ -18,6 +21,7 @@ const props = defineProps<{
 
 const { t } = useI18n();
 const route = useRoute();
+const router = useRouter();
 
 const emit = defineEmits([
   "addTableToConfig",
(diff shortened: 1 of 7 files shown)


rudokemper (approved):
Great.

PR #517, opened 2026-07-15
Fix config fields rendering stale values 

### LINKED ISSUE #513: coordinates and zoom level dissapear when choosing projection
One of our partners were attepting to do their first view using `gc-explorer`.
while inputing the data, the user entered the zoom level, then the lat and lon, and then when selecting the projection, the lat, long and zoom dissappeared.

Image

haven't yet tried to replicate, want to document first.

### LINKED ISSUE #514: Media config fields don't populate from saved config (and can wipe it on save)

The Media section on the config edit page never shows saved values (Filebrowser
share URL/hash, media column), even when `MEDIA_BASE_PATH` etc. are saved.

Worse: saving without touching the Media section submits empty media values over
the saved ones, silently wiping media config.

Cause is a mount-order race:

- `ConfigCard` copies `viewConfig` into `localConfig` in its `onMounted`
- `ConfigMedia` reads `props.config` into local state in its own `onMounted`,
  and children mount before parents  so it always reads the still-empty `{}`
- Collapsible sections use `v-show`, so `ConfigMedia` mounts with the page,
  and it has no watcher on `props.config` to recover

Only Media is affected: the other sections (Map, Filters, Dataset) bind config
reactively in their templates, so they self-heal once `localConfig` fills.

#### Steps to Reproduce

1. Configure a view with a Filebrowser share URL saved in `MEDIA_BASE_PATH`
2. Open `/config/{dataset}?view_type={type}` and expand the Media section
3. Share URL/hash input is empty despite the saved value
4. Change any unrelated field and Save → media config is overwritten with empty values

## Expected Behavior

- Saved media values populate the Media section on load
- Saving without touching Media leaves saved media config unchanged

conservationtimothy:
## Goal

Fix the config regression where saved fields render empty and newly entered zoom, latitude, and longitude values visually disappear after changing map projection. Closes #513. Closes #514 

## What I changed and why

- Fixed config sections mounting before `ConfigCard.onMounted()` populated `localConfig`, causing components that captured config during setup or mount to retain empty values
- `ConfigMap` was most affected: numeric and 3D fields read from a stale cached reference to the original empty object, while projection read from the live prop — changing projection triggered a render that restored numeric inputs from the stale reference, making entered values appear to vanish even
- 

Database diagnostics confirmed the affected values were persisted correctly and returned with the expected numeric types, ruling out migration or API data loss. Git history showed that #470 changed section visibility from a delayed `availableViews` check to an immediately available `viewType`, exposing setup-time snapshots that had previously been masked by the old mount order. The new map regression test failed before this fix because zoom rendered as an empty string, then passed after the config was initialized synchronously and the stale alias was removed.

## What I'm not doing here

This does not change database storage, migrations, API payloads, view routing, or the type-first create flow. It only corrects config state initialization and rendering.

## LLM use disclosure

GPT Sol helped


rudokemper:
It seems that this solves the problem for the specific field in `ConfigMedia`, but I think there is a broader regression at play. This never used to be an issue. And it is not unique to `ConfigMedia`:

Look at `ConfigMap`. Zoom level, latitude, longitude never used to be empty. 

And there is an even easily reproducible bug c.f https://github.com/ConservationMetrics/gc-explorer/issues/513:

1. Set values for zoom level, latitude, and longitude
2. Now change the map projection
3. See the values disappear

Can you investigate the nature of this regression _globally_ across config, before implementing anything new? And if you do find that new solution like watcher is needed, implement it in a DRY way - but I would love to see a theory as to why this regression happened first.

I can say with pretty strong confidence is that this was never the case before we merged the Multiple Views API epic. Maybe worth spinning up a local GC Explorer docker container based on last image before that commit?

I think `communityfirst/guardianconnector-explorer:20260603-823` was our last stable image we were working with.

---

As an aside, in tests, avoid using `demo` since that is not a sandbox / staging deployment anymore, and in fact, avoid using our real production URLs at all.

rudokemper:
I can confirm the regression.

I deployed `communityfirst/guardianconnector-explorer:20260603-823` to our staging server.

Please check out: `https://explorer-test.{BASE_URL}/config/bcmform_responses` (you know what the base URL is for our staging server).

You will see that neither https://github.com/ConservationMetrics/gc-explorer/issues/513 nor https://github.com/ConservationMetrics/gc-explorer/issues/514 are an issue there.

utils/viewFilters 2.ts
```diff
@@ -0,0 +1,43 @@
+import type { ViewType } from "@/types";
```
rudokemper: A number of files here end with ` 2`

rudokemper: Actually, all of these files should not be in the PR it would seem - they are from the other work to add search bar and filters to `/config`?

rudokemper (changes requested):

utils/viewFilters 2.ts
conservationtimothy: this was as a result of local filesystem conflict copy

rudokemper (approved):

PR #490, opened 2026-06-24
Playwright tests failing on timeout exceeded (again)

conservationtimothy:
## Goal

Fix the Playwright failures, especially alerts dashboard map-readiness timeouts and RBAC/public-view CI behavior. Closes #441 .

## Screenshots

Not applicable.

## What I changed and why

I removed CI’s dependency on remote Mapbox style loading by having seed view configs use a minimal local Mapbox style, and updated map components to accept typed Mapbox style objects in addition to style URLs. The alerts dashboard now skips adding symbol label layers when the active style has no glyphs, keeping the local CI style usable while still rendering alert and Mapeo circle layers for tests.

I also made the map readiness signal more explicit. The concrete issue I was addressing was that our E2E readiness signal depended on an extra Mapbox `idle` event after the alert/Mapeo layers had already been added, which made test synchronization harder to reason about. Since `prepareMapCanvasContent()` already awaits alert/Mapeo layer setup and then calls `prepareMapLegendContent()` last, the legend can be built directly from the layers that were just added. I moved the test-only `mapReady` signal to the end of `prepareMapCanvasContent()` so it marks the actual readiness boundary: map data layers and legend content have been prepared.

Finally, I fixed an RBAC/public-view bug where `/api/config/public_views` read the `public_views` table directly instead of the seeded view rows used elsewhere in CI. That allowed middleware to deny a public dataset before the seeded `ROUTE_LEVEL_PERMISSION: "anyone"` config was applied. `fetchPublicViewTableNames()` now returns public datasets from the CI seed rows, with a unit regression test covering this behavior.

## How I convinced myself this is right

The alert failures were all waiting for map UI that is created after Mapbox’s `load` path, so the most plausible common failure was not 15 independent Playwright flakes but the dashboard never reaching its loaded map state in CI. A minimal local style removes the remote style API from that path while still exercising the application’s own layer/source setup. The glyph guard is necessary because the minimal style intentionally has no glyph URL, and Mapbox symbol layers that render text require glyphs.

The final RBAC failure showed the guest public-dataset precondition returning `denied`, which contradicted the CI seed config. Following the middleware path showed it checks `/api/config/public_views` before route-level config, and that API did not have the CI fallback used elsewhere. The new unit test executes that behavior directly by setting `CI=true` and asserting the seeded public datasets are exposed.

## What I'm not doing here

I improved the local Playwright workflow during investigation, but that is out of scope for this PR and was removed from this branch.

## LLM use disclosure

[...]

PR #524, opened 2026-07-16
Fix oauth middleware to read views-shaped `/api/config`

conservationtimothy:
## Goal

Restore role-based access checks on authenticated dataset routes after the views migration.

## Screenshots

N/A

## What I changed and why

`/api/config` now returns `{ views, availableTables }`, but `middleware/oauth.global.ts` still treated the response as the old `[tableConfig, tables]` tuple.

That made `const [tableConfig] = response` throw. The middleware `catch` only forces login for unauthenticated users, so for logged-in users the permission switch never ran and RBAC was silently skipped.

This PR reads permission from `response.views` by `primaryDataset`.

- Tested manually setting to different roles in Auth0 ensuring say a `SignedIn` user cannot access an Admin route etc.
- Confirmed `server/api/config/index.get.ts` returns `{ views, availableTables }`.
- Confirmed the old middleware destructuring throws on that object shape and falls into the catch path.
- After the change, permission resolves from `viewEntry.viewConfig.ROUTE_LEVEL_PERMISSION` with the same `"member"` default as before.

## What I'm not doing here

- No E2E / Playwright / seed changes (those stay on https://github.com/ConservationMetrics/gc-explorer/pull/490).
- No changes to public_views unauthenticated access.
- No broader auth or config API refactor.

## LLM use disclosure

None

### DIFF
--- a/middleware/oauth.global.ts
+++ b/middleware/oauth.global.ts
@@ -47,16 +47,18 @@ export default defineNuxtRouteMiddleware(async (to) => {
         return;
       }
 
-      const response =
-        await $fetch<
-          [
-            Record<string, { ROUTE_LEVEL_PERMISSION?: RouteLevelPermission }>,
-            string[],
-          ]
-        >("/api/config");
-      const [tableConfig] = response;
+      // /api/config returns { views, availableTables } after the views migration.
+      const response = await $fetch<{
+        views: Array<{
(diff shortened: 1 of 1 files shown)


conservationtimothy:
I set every single role actually.

rudokemper (approved):

PR #526, opened 2026-07-17
Hide `BasemapSelector` component if only one basemap available

rudokemper:
## Goal

https://github.com/ConservationMetrics/gc-explorer/commit/16d133857d9211e9e7403d6b4579f29ff61dd89e removed hard-coded basemap options, which means there is a possible state where the available basemaps == 1. And in that case, a basemap selector control doesn't make any sense.

## What I changed and why

- Conditionally render `BasemapSelector` component if `mapboxBasemaps.length` PLUS Planet basemap (if available) > 1.
- Tests to check for this behavior.

I clicked on a Basemap Selector control on a map with only one basemap, and thought it looked very odd to only have one radio button choice.

## LLM use disclosure

None

### DIFF
--- a/components/AlertsDashboard.vue
+++ b/components/AlertsDashboard.vue
@@ -1757,7 +1757,10 @@ onBeforeUnmount(() => {
       @toggle-layer-visibility="toggleLayerVisibility"
     />
     <BasemapSelector
-      v-if="showBasemapSelector"
+      v-if="
+        showBasemapSelector &&
+        ((mapboxBasemaps?.length ?? 0) > 1 || !!planetApiKey)
+      "
       :has-ruler-control="hasRulerControl"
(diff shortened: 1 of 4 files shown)


components/AlertsDashboard.vue
```diff
@@ -1757,7 +1757,10 @@ onBeforeUnmount(() => {
       @toggle-layer-visibility="toggleLayerVisibility"
     />
     <BasemapSelector
-      v-if="showBasemapSelector"
+      v-if="
+        showBasemapSelector &&
+        ((mapboxBasemaps?.length ?? 0) > 1 || !!planetApiKey)
```
conservationtimothy, line 1762: I like this code.

conservationtimothy (approved):

PR #529, opened 2026-07-20
 Make "Copy config from dataset" view-type aware

### LINKED ISSUE #510: Make "Copy config from dataset" view-type aware
## Feature Request
`useCopyConfig` on the config edit page offers every other dataset as a copy source, regardless of view type; copying map styling into a gallery config makes no sense.

There is also a latent bug now that datasets can have multiple views:
`pages/config/[dataset].vue` builds the copy-source lookup keyed by `primaryDataset` only, so when a dataset has several views, the last row wins and the wrong view's config can be offered/copied.

- Copy sources are limited to views of the **same view type** as the one being edited
- With multiple views on one dataset, the correct view's config is copied (keyed by view, not dataset)
- If no same-type views exist, the copy button is hidden (current behavior for empty list)

## Implementation Plan

conservationtimothy:
## Goal

Make “Copy Config from Another View” view-type aware so a map edit cannot pull gallery settings (or vice versa), and so a dataset with multiple views copies the matching type’s config instead of whichever row last won a dataset-keyed lookup. Closes #510 

## Screenshots

Screenshot 2026-07-20 at 10 40 11
Screenshot 2026-07-20 at 10 39 55

## What I changed and why

- Fixed `useCopyConfig` building a map keyed only by `primaryDataset`, which caused last-write-wins behavior when a table had two views, potentially offering the wrong sibling or a mismatched view type in the copy modal
- Composable now reads `ViewConfigRow[]`, filters to rows matching the current `viewType`, excludes the current primary dataset, and looks up the chosen source by `primaryDataset::viewType`
- Config edit page drops the dataset-keyed reduce and passes `viewRows` and `resolvedViewType` into the composable directly
- Modal options render source labels from the filtered rows

Tested it myself and the e2e path opens the seeded bcmform gallery card (which has a gallery peer) and checks the modal lists seed_survey_data, so the happy path still exercises a real same-type copy in CI.

## What I'm not doing here

- anything extrenous

## LLM use disclosure

Cursor Grok 4.5 Helped

### DIFF
--- a/composables/useCopyConfig.ts
+++ b/composables/useCopyConfig.ts
@@ -1,29 +1,60 @@
-import type { Views, ViewConfig } from "@/types";
+import type { ViewConfig, ViewConfigRow, ViewType } from "@/types";
+
+export type CopyConfigSource = {
+  key: string;
+  label: string;
+  viewConfig: ViewConfig;
+};
 
 /**
- * Composable for copying configuration from one dataset to another.
- * Manages modal state, dataset selection, and config cloning.
+ * Builds a stable lookup key for a view copy source.
  *
- * @param {Ref<Views>} viewsConfig - Reactive reference to all dataset configs.
- * @param {string} currentDataset - The current dataset key to exclude from the list.
+ * @param {string} primaryDataset - Primary dataset table name.
+ * @param {ViewType} viewType - View type of the source row.
+ * @returns {string} Key unique per (dataset, view type).
+ */
+export const copySourceKey = (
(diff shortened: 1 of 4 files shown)


rudokemper (approved):
Good. Let's also make some copy changes.

- Select a **view** to copy its configuration values into the current form. You can modify the values before saving.
- Select a **view...**

Image

conservationtimothy:
Done
Screenshot 2026-07-20 at 15 01 08

PR #536, opened 2026-07-21
For alerts, select Mapeo data table as secondary dataset; not specify in ConfigAlerts

### LINKED ISSUE #531: For alerts, select Mapeo data table as secondary dataset; not specify in ConfigAlerts
## Feature Request

Image

Image

We need to do some work to have Mapeo data be the selected **secondary dataset** for alerts views. 

## Implementation Plan

conservationtimothy:
## Goal

“Name of Mapeo data table” field is the sole source of Secondary dataset. Closes #531 

## What I changed and why

By default the behaviour we want ALREADY exists so this PR just Added a unit test that `buildViewConfigColumns` for alerts leaves `secondaryDataset` null when the value is omitted or whitespace-only. The editable Mapeo table field in ConfigAlerts already writes that column on save (so the top Secondary dataset metadata matches after save); this test guards against reintroducing a hardcoded default. 

- `pnpm test:unit tests/unit/server/viewConfigModel.test.ts` passes, including the new blank/null case and the existing case that a typed secondary (e.g. a partner table name) is preserved.

## LLM use disclosure

Used Cursor to add the regression test

### DIFF
--- a/tests/unit/server/viewConfigModel.test.ts
+++ b/tests/unit/server/viewConfigModel.test.ts
@@ -70,6 +70,21 @@ describe("buildViewConfigColumns", () => {
     expect(galleryColumns.secondaryDataset).toBeNull();
   });
 
+  it("keeps alerts secondaryDataset null when omitted or blank", () => {
+    const config: ViewConfig = {
+      DATASET_TABLE: "Fake Alerts",
+    };
+
+    expect(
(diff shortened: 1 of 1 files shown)


rudokemper (approved):
Great, didn't realize the desired behavior was already there! (I tested it in runtime and can confirm too)

Issue #523, opened 2026-07-16
Remove /config dashboard; move "Add new dataset view" to index


conservationtimothy:
## Feature Request

With per-view cards + admin gear on the index, the /config dashboard is a
duplicate listing. Remove it:
- "Manage Datasets" nav becomes "Add new dataset view", linking to the create
  flow entry page
- `/config` (dashboard) route removed; per-view edit routes
  (`/config/{dataset}?view_type=`) stay
- On the config edit page, "View dataset" links directly to the view route
  (e.g. `/alerts/fake_alerts`) instead of the removed `/dataset/{dataset}`

### User Stories & Acceptance Criteria

- No /config dashboard; adding a view starts from the index
- Editing reached via card gear; "View dataset" on the edit page opens the view

## Implementation Plan
(may be added later, before work begins)

conservationtimothy:
"We should have middleware redirects from /config and /dataset back to /, with a toast to please update your bookmarks.

In case anyone had /config or /dataset/{...} bookmarked"

rudokemper:
> "We should have middleware redirects from /config and /dataset back to /, with a toast to please update your bookmarks.
> 
> In case anyone had /config or /dataset/{...} bookmarked"

I think this should be a separate issue to keep scope.

PR #538, opened 2026-07-22
Replace dataset-centric view creation workflow

### LINKED ISSUE #499: Replace dataset-centric view creation workflow

conservationtimothy:
## Goal

Replace the dataset-first “add view” modal with a type-first create flow (`/config/new` → `/config/new/{view_type}`), including primary/secondary selection, same-type copy-config, and duplicate detection before save. Closes #499 

## Screenshots
Screenshot 2026-07-22 at 12 14 18
Screenshot 2026-07-22 at 12 01 50
Screenshot 2026-07-22 at 12 00 37
Screenshot 2026-07-22 at 12 00 30

Screenshot 2026-07-21 at 22 20 19

## What I changed and why

- Admins now choose a view type first, then configure primary (and optional secondary for map/alerts) on a dedicated create page, replacing the old modal that only listed unconfigured tables
- `availableTables` now returns all warehouse tables so a dataset with an existing map view can still get gallery or alerts views added
- Duplicate `(view_type, primary)` pairs are checked client-side and block Save, with a link to the existing edit page
- `POST new_table` now accepts a full config body including secondary dataset
- Create form supports copying from another same-type view, including that source's `secondary_dataset`, which seeds both the secondary dropdown and the alerts Mapeo text field
- On edit, secondary dataset remains visible in metadata and the alerts "Name of Mapeo data table" field stays editable so users can change the companion table after creation
- For new creations, we can copy an existing dataset view which is also filtered on `view_type`. So for a new alerts view, we can copy only alerts views. 

- Manual: create map/alerts with secondary; copy from an alerts view that has a Mapeo table and confirm dropdown + Mapeo text fill; edit alerts still shows Mapeo field.
- Unit: `useDuplicateViewCheck`, `useCopyConfig` (including secondary on copy + primary), `buildViewConfigColumns` for map secondary vs gallery null.
- E2E: create via type-first flow, cancel without DB write, deep-link `?primary=`, duplicate warning disables Save.

## LLM use disclosure

Cursor Grok 4.5 helped with heavy edits by me


conservationtimothy:
A question that came up, but is out of scope for this PR, is whether we want to allow editing secondary dataset fields where applicable. For example, `Name of mapeo data`, which is currently set to `mapeo_data`, is effectively the secondary dataset. If a user changes this value, that change should be reflected in Alerts. Similarly, for Maps, a valid workflow might be editing only the secondary dataset -- for example, changing it from `comapeo_cop` to `bcmform_labels` without modifying anything else.

rudokemper:
Haven't reviewed yet but just to speak to this:

> A question that came up, but is out of scope for this PR, is whether we want to allow editing secondary dataset fields where applicable. For example, `Name of mapeo data`, which is currently set to `mapeo_data`, is effectively the secondary dataset. If a user changes this value, that change should be reflected in Alerts. Similarly, for Maps, a valid workflow might be editing only the secondary dataset -- for example, changing it from `comapeo_cop` to `bcmform_labels` without modifying anything else.

Yes. I think so.

User story:

"Mapeo data" is old data, from the legacy Mapeo app, and I now want to overlay my CoMapeo dataset from the next gen app.

I should be able to change this value and have it be reflected everywhere for that Alert view (config, the actual view, etc.)

conservationtimothy:
Created an issue: https://github.com/ConservationMetrics/gc-explorer/issues/541

pages/config/new/[view_type].vue
```diff
@@ -0,0 +1,345 @@
...
+      <ChevronLeft class="w-5 h-5" />
+      {{ $t("addNewTable") }}
+    </NuxtLink>
+
+    <h1 class="text-2xl sm:text-3xl font-bold text-gray-900 mb-2">
+      {{ $t("configureNewView") }}
+    </h1>
+    <p class="text-gray-600 mb-6">
+      {{ $t(viewType) }} — {{ $t("primaryDatasetLabel") }}
+    </p>
```
Comment (rudokemper) [inline pages/config/new/[view_type].vue]: Instead of this, can we use the same View Pill that we use on existing config routes when the view has already been written to db?

Image

pages/config/new/[view_type].vue
```diff
@@ -0,0 +1,345 @@
...
+
+          <SecondaryDatasetFields
+            v-if="showsSecondaryDataset"
+            table-name="create-view"
+            :secondary-dataset="secondaryDataset"
+            :available-tables="availableTables"
+            :primary-dataset="primaryDataset"
+            @update-secondary-dataset="handleSecondaryDatasetUpdate"
+          />
+        </div>
```
Comment (rudokemper) [inline pages/config/new/[view_type].vue:209]: Could you create a component `SelectDatasetField` and use for both?

Furthermore,

Image

Few fixes here:

`Primary dataset (required)`

`Secondary dataset (optional)`

Instead of "None", "Select a secondary dataset..."

pages/config/new/[view_type].vue
```diff
@@ -0,0 +1,345 @@
...
+        >
+          <div class="mb-4">
+            <CheckCircle2 class="w-16 h-16 mx-auto text-green-500" />
+          </div>
+          <h2 class="text-2xl font-bold text-gray-900 mb-2">Saved!</h2>
+          <p class="text-gray-600">
+            Configuration has been saved successfully.
+          </p>
+        </div>
+      </div>
```
Comment (rudokemper) [inline pages/config/new/[view_type].vue]: This is now duplicated with behavior on the `/config/` route, right? Let's turn this into a component.

pages/config/new/[view_type].vue
```diff
@@ -0,0 +1,345 @@
...
+            type="button"
+            data-testid="copy-config-button"
+            class="inline-flex items-center gap-2 px-4 py-2 text-sm font-medium text-gray-700 bg-white border border-gray-300 rounded-lg hover:bg-gray-50 transition-colors"
+            @click="handleOpenCopyModal"
+          >
+            <Copy class="w-4 h-4" />
+            {{ $t("copyConfigFromDataset") }}
+          </button>
+        </div>
+      </div>
```
Comment (rudokemper) [inline pages/config/new/[view_type].vue]: See below -- turn into component to be able to use here and `/config/<dataset>`

pages/config/new/[view_type].vue
```diff
@@ -0,0 +1,345 @@
...
+      errorMessage.value = t("errorCouldNotAddDataset");
+    }
+    showErrorToast(t("errorCouldNotAddDataset"));
+  } finally {
+    isSaving.value = false;
+  }
+};
+
+useHead({
+  title: "GuardianConnector Explorer: " + t("addNewTable"),
```
Comment (rudokemper) [inline pages/config/new/[view_type].vue:152]: Not in this PR, but we should make an effort to rename all of our i18n keys.

Like, this should be `addNewDatasetView`. And there are a few other examples in the lines above where we are still dataset-centric in our naming of keys. 

Can you make a standalone issue for this?

pages/config/index.vue
```diff
@@ -1,9 +1,8 @@
 <script setup lang="ts">
```
rudokemper, line 1: I don't totally follow all the changes on this page, but I'm also not concerned because in https://github.com/ConservationMetrics/gc-explorer/issues/523 we are getting rid of this config dashboard altogether.

pages/config/[dataset].vue
```diff
@@ -18,10 +18,15 @@ const datasetRaw = route.params.dataset;
 const dataset = Array.isArray(datasetRaw)
   ? datasetRaw.join("/")
   : String(datasetRaw || "");
+
+// Static /config/new* routes win in Nuxt; this guards odd matches if routing changes.
+if (dataset === "new") {
+  await navigateTo("/config/new");
+}
+
```
Comment (rudokemper) [inline pages/config/[dataset].vue]: I don't follow.

In the rare edge case that my primary dataset is called `new`, and I've created a view for it, then I should be able to configure that at `/config/new?view_type=`... right?

And we can differentiate route that from `/config/new/*`?

components/config/ConfigAlerts.vue
```diff
@@ -7,13 +7,21 @@ import { updateTags } from "@/composables/useTags";
...
-  views: Array<string>;
-  keys: Array<string>;
-}>();
+const props = withDefaults(
+  defineProps<{
+    tableName: string;
+    config: ViewConfig;
+    secondaryDataset?: string | null;
+    /** Create flow: show Name of Mapeo data table (synced with secondary). */
+    showMapeoTableField?: boolean;
```
rudokemper, line 16: I think we need a follow-up issue to generalize `secondaryDataset` for the Alerts Dashboard.

There is no really no need to specify "Mapeo table". That's a hard-coded relic of how we did this in the past.

There is nothing unique about "Mapeo table" or Mapeo data for the alerts dashboard. We can be data source agnostic here, and allow the user to pick _any_ secondary dataset to render on the alerts dashboard, so long as it's geospatial data.

Can you make an issue to follow up on this?

server/database/dbOperations.ts
```diff
@@ -117,8 +118,9 @@ export const buildViewConfigColumns = (
     viewName: config.DATASET_TABLE?.trim() || primaryDataset,
     viewType,
     primaryDataset,
+    // Map and alerts may reference a companion warehouse table; gallery does not.
     secondaryDataset:
-      viewType === "alerts" ? secondaryDataset?.trim() || null : null,
+      viewType === "alerts" || viewType === "map" ? trimmedSecondary : null,
```
rudokemper: I could have sworn that somewhere, we explicitly declared which datasets can have a secondary / companion table, and which ones can not. But perusing the codebase I can't find it just now.

I would love to use that here. That way we don't have to hard code the behavior in multiple places.

If, one day, we decide that gallery **can** have a secondary dataset, then it's just one place where we make that update.

If we didn't already do this, could we make it part of this PR so as to leverage it everywhere?

rudokemper (changes requested):
On the right track! Behavior looks good but made comments to improve code maintainability, and a few UI changes.

pages/config/new/[view_type].vue
Comment (conservationtimothy) [reply pages/config/new/[view_type].vue:152]: https://github.com/ConservationMetrics/gc-explorer/issues/543

components/config/ConfigAlerts.vue
conservationtimothy: https://github.com/ConservationMetrics/gc-explorer/issues/544.

pages/config/new/[view_type].vue
Comment (conservationtimothy) [reply pages/config/new/[view_type].vue]: Screenshot 2026-07-23 at 22 19 19
Screenshot 2026-07-23 at 22 20 14


Done. extracted ViewTypePill and reused it on create and edit routes

pages/config/new/[view_type].vue
Comment (conservationtimothy) [reply pages/config/new/[view_type].vue:209]: Screenshot 2026-07-23 at 22 21 30
Done

pages/config/new/[view_type].vue
Comment (conservationtimothy) [reply pages/config/new/[view_type].vue]: extracted the duplicated saved confirmation into `SavedModal`

pages/config/new/[view_type].vue
Comment (conservationtimothy) [reply pages/config/new/[view_type].vue]: Extracted the shared copy button and modal into `CopyConfigControl`

pages/config/[dataset].vue
Comment (conservationtimothy) [reply pages/config/[dataset].vue]: Good catch. I overthought the route precedence here. The static route wins, so this guard is unreachable. Removed.

components/shared/ViewTypePill.vue
```diff
@@ -0,0 +1,24 @@
+<script setup lang="ts">
```
rudokemper, line 1: I'd move this to `components/shared`. It could be reused outside of config.

utils/viewTypes.ts
```diff
@@ -0,0 +1,19 @@
...
+  "alerts",
+  "map",
+] as const satisfies readonly ViewType[];
+
+/** Whether a view type can reference a companion warehouse table. */
+export const supportsSecondaryDataset = (
+  viewType: ViewType | undefined,
+): boolean =>
+  viewType !== undefined &&
+  SECONDARY_DATASET_VIEW_TYPES.some((type) => type === viewType);
```
rudokemper: Move to `types/index.ts`. There is already some View / View type content in there.

rudokemper (approved):
Resolve comments before merging. Thanks!

PR #558, opened 2026-07-29
Decode encoded dataset table names for Unicode warehouse tables

### LINKED ISSUE #554: Dataset table routes (/`view_type/*`, `/config/*`) with non-Latin chars don't work
## Feature Request

Following up on https://github.com/ConservationMetrics/gc-scripts-hub/issues/262, where for a dataset with Thai characters,

> ❌ Explorer: I could add a dataset view, but once set up, neither opening the /map/<name> nor /config/<name> worked. (Map view: "We couldn't load this data. Config: redirected back to /).

@IamJeffG:

> No matter what characters you throw in the database, we could make certain that Explorer is always properly calling `encodeURI` on database names to convert into URLs, and `decodeURI` on URL components to restore special characters.

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-explorer/issues/554.

## What I changed and why

- https://github.com/ConservationMetrics/gc-explorer/commit/1f628ec4c14e4b345a6fdcb26ae5f92a02d7117d -- Add URL encode/decode helpers for dataset table names: `encodeDatasetNameForUrl` / `decodeDatasetNameFromUrl` / `normalizeTableName`, with Thai round-trip tests
- https://github.com/ConservationMetrics/gc-explorer/commit/41ea6e132becd34d83778a91a55c4e8997881125 -- Normalize table names before warehouse and config lookups (decode before `to_regclass` / `sql.identifier`, matching legacy encoded `primary_dataset` rows
- https://github.com/ConservationMetrics/gc-explorer/commit/45d6831b01996775af80fe10a462121cd250ac4c -- Decode `[table]` path params in API handlers (`getTableParam` at entry for `[table]` + config routes)
- https://github.com/ConservationMetrics/gc-explorer/commit/fdb7612c61513ece754bf15531f5c917a4bb0d48 -- Encode and decode dataset names in client routes and links (Unicode in app state; encode path segments only; decode in pages, oauth middleware, composables)
  - Keep the real Unicode table name in component/composable state
  - `encodeURIComponent` only when building URL path segments (`/map/...`, `/api/...`, `/config/...`)
  - `decodeURIComponent` when reading route params in map/gallery/alerts/config pages, oauth middleware, and related composables

- Unit tests
- Extensively trying it in runtime with real datasets.

## LLM use disclosure

Largely prompt engineered by various flavors of Claude in Cursor. But I vetted the code commit-by-commit and made minor tweaks when necessary.


server/database/dbOperations.ts
```diff
@@ -13,10 +13,26 @@ import {
...
 import { viewConfig, publicViews } from "./schema";
 import { configDb, warehouseDb } from "./dbConnection";
 
+export { normalizeTableName };
+
+/**
+ * Keys to match against views.primary_dataset when the DB may still hold a
+ * percent-encoded legacy value for the same table.
+ */
+const datasetNameLookupKeys = (table: string): string[] => {
```
conservationtimothy: `datasetNameLookupKeys()` can't fire. Every `[table]` handler goes through `getTableParam()`, which already decodes, so `new Set([normalized, raw])` collapses to one key. I verified by mocking `inArray` and tracing the real path; the production flow only ever queries `["สำรวจใหม่"]`.

The real question is existing data. Pre-PR rows would have been stored encoded: the old handler passed `event.context.params.table` straight into `addNewTableToConfig`, and although the client never called `encodeURIComponent`, URL serialization percent-encodes non-ASCII path segments anyway. If such rows exist this fallback doesn't rescue them, because `/config` saves run `UPDATE … WHERE primary_dataset IN ('<decoded>')`, match 0 rows, and still show success -- nothing checks the rowcount. Same for deleting a view.

Could you run this on a deployment with Thai datasets?

```sql
SELECT view_id, view_type, primary_dataset, secondary_dataset FROM views
WHERE primary_dataset ~ '%[0-9A-Fa-f]{2}' OR secondary_dataset ~ '%[0-9A-Fa-f]{2}';

SELECT table_name FROM public_views
WHERE table_name ~ '%[0-9A-Fa-f]{2}';
```

If there are no rows → drop the helper and keep the simpler `eq()` lookups. Rows → a one-off migration decoding those columns, since the read-side fallback can't fix the write paths anyway. Either way the helper's comment describes something unreachable.

server/api/[table]/[recordId].get.ts
```diff
@@ -1,12 +1,13 @@
...
 import { validatePermissions } from "@/utils/accessControls";
 
 import type { H3Event } from "h3";
 import type { ViewType } from "@/types";
 
 export default defineEventHandler(async (event: H3Event) => {
-  const { table, recordId } = event.context.params as {
-    table: string;
+  const table = getTableParam(event);
+  const { recordId } = event.context.params as {
```
Comment (conservationtimothy) [inline server/api/[table]/[recordId].get.ts]: I think this reintroduces, for `recordId`, the exact issue the PR fixes for table names. The [client](https://github.com/ConservationMetrics/gc-explorer/pull/558/changes#diff-b3ac041d8d1edc044b2cad1ff614a196f24927c7ad796b476731d99c332af881R79) now sends `encodeURIComponent(recordId)`, but the server reads the param directly without decoding.

Route params really aren't decoded that's the premise `getTableParam()` is built on (I believe). h3 decodes only when you ask (`if (opts.decode)`, `h3/dist/index.mjs:254`), and Nitro doesn't. I ran a request through the same h3 version with `/api/:table/:recordId`: `event.context.params` came back as `{"recordId":"record%201%2F%E0%B8%81"}`, and only `getRouterParams(event, { decode: true })` gave `record 1/ก`.

 `fetchRecord` binds the id as a value, so there's no SQL error, just zero matches and returns 404 `Record not found`. Affects https://github.com/ConservationMetrics/gc-explorer/blob/2a5747cfdbb5be5cdb1891409b8869e91aa4856e/components/MapView.vue#L357 and https://github.com/ConservationMetrics/gc-explorer/blob/2a5747cfdbb5be5cdb1891409b8869e91aa4856e/components/AlertsDashboard.vue#L232.

Caveat: plain hex/UUID `_id`s are unaffected, so this won't show up in testing. But for those the new `encodeURIComponent` is also a no-op, so either it needs a matching decode here or it shouldn't be there. Maybe a `getRecordIdParam()` alongside `getTableParam()` doing the guarded decode, with the existing `trim() === ""` check moved after it? The other two paths already look fine, batch ids go in the POST body, and the export route takes `recordId` as a query param, which h3 decodes.

conservationtimothy (changes requested):
Looks good generally just some changes / feedback

server/api/[table]/[recordId].get.ts
Comment (rudokemper) [reply server/api/[table]/[recordId].get.ts]: Really good catch, this is the same encode-without-decode hole as was happening with table names. 

I like your suggestion to add a parallel `getRecordIdParam()` for `recordId` and implemented it here:

https://github.com/ConservationMetrics/gc-explorer/pull/558/commits/e3ae3ff36dd036c18f2cf8a6f87159acec6043f3

server/database/dbOperations.ts
rudokemper: Also a good catch, thanks! 

I don't think we need to sweat the existing data. The only two data sources for which this applies can easily be reconstructed from upstream data. So I opted to drop `datasetNameLookupKeys` and switched lookups to `eq()`.

https://github.com/ConservationMetrics/gc-explorer/pull/558/commits/264d3b499d9316b4c67d10669a40011beee9050f

rudokemper:
https://github.com/ConservationMetrics/gc-explorer/actions/runs/30572455468/job/90973085917 Playwright test failure seems unrelated.

server/api/[table]/[recordId].get.ts
Comment (conservationtimothy) [reply server/api/[table]/[recordId].get.ts]: Thank you!

conservationtimothy (approved):
Thanks for this and incorporating my feedback. LGTM

PR #559, opened 2026-07-30
Support authStrategy=none with a local Admin session (quality-of-life DX improvement)

rudokemper:
## Goal

Before, we supported `NUXT_PUBLIC_AUTH_STRATEGY=none`, but I think this broke when we introduced RBAC.

Having spent some time working on GC Explorer dev recently, I find it annoying to have to hop through auth0 every single time localhost runtime starts up. 

This PR restores full `authStrategy=none` support by seeding a local Admin session and skipping Auth0 route/API gates, so local dev no longer requires an Auth0 login hop. (But you can of course set `=auth0` if you _are_ working on RBAC related things.)

### DIFF
--- a/server/middleware/local-auth.ts
+++ b/server/middleware/local-auth.ts
@@ -3,10 +3,12 @@ import { Role } from "@/types";
 
 /**
  * When NUXT_PUBLIC_AUTH_STRATEGY=none, ensure every request has a local
- * Admin session so route/API RBAC works without Auth0 (local/dev/CI).
+ * Admin session so route/API RBAC works without Auth0.
  */
 export default defineEventHandler(async (event) => {
-  const { public: { authStrategy } } = useRuntimeConfig();
+  const {
+    public: { authStrategy },
+  } = useRuntimeConfig();
   if (authStrategy !== "none") return;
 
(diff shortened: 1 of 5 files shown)


server/middleware/local-auth.ts
```diff
@@ -0,0 +1,32 @@
...
+          id: "local-admin",
+          name: "Admin",
+          description: "Local development admin (authStrategy=none)",
+        },
+      ],
+      userRole: Role.Admin,
+    },
+    loggedInAt: Date.now(),
+  });
+});
```
conservationtimothy, line 32: Have no doubt this works but curious *how* it works? How does it intercept when "none" to set the session

conservationtimothy (approved):

server/middleware/local-auth.ts
rudokemper: It’s Nitro server middleware: anything in `server/middleware/` is auto-registered and runs on every incoming HTTP request before API handlers / page SSR.

https://nitro.build/examples/middleware#defining-middleware

conservationtimothy: Ah very cool

Issue #435, opened 2026-05-08
[EPIC] UI improvements for Config dashboard


rudokemper:
## Feature Request

Placeholder for now.

### User Stories & Acceptance Criteria

- Many fields that target selecting a table column should be dropdowns / typeaheads, NOT string fields

### UI Sketch (or Screenshot)

See both Mobile and Desktop pages from Transistir Social Tech here: https://drive.google.com/drive/u/1/folders/1mD9WMLmmnbm_m-Ay3j4v2wQp5JY_Tt_5

## Implementation Plan
(may be added later, before work begins)

rudokemper:
Save button should be more visible on the giant kitchen sink config form; or, maybe auto-saving on each field save?

PR #563, opened 2026-07-31
Generalize alerts secondary datasets beyond Mapeo

### LINKED ISSUE #544: Generalize alerts secondary datasets beyond Mapeo
## Feature Request

Make secondary datasets on the Alerts Dashboard data source agnostic. Remove Mapeo-specific naming and allow an alerts view to use any compatible geospatial secondary dataset.

The current `showMapeoTableField`, `mapeoTable`, and related wording are relics of the original implementation. Mapeo data does not require unique handling at the configuration boundary.

- As an administrator, I can select any compatible table as the secondary dataset for an alerts view.
- Alerts configuration uses generic "secondary dataset" terminology instead of "Mapeo table".
- The Alerts Dashboard renders geospatial features from the configured secondary dataset.
- Existing alerts views configured with Mapeo data continue to work.
- Non-geospatial tables are excluded or produce clear validation feedback.
- Mapeo-specific props, translation keys, and internal configuration language are removed where they no longer represent unique behavior.

### UI Sketch (or Screenshot)

Reuse the existing secondary dataset selector with the label "Secondary dataset (optional)".

## Implementation Plan

conservationtimothy:
## Goal

Make the Alerts Dashboard secondary layer data-source agnostic: any compatible geospatial table can be configured and rendered, with generic “secondary dataset” terminology instead of Mapeo-specific naming, while existing Mapeo-backed alerts views keep working. Closes #544 

## Screenshots
Screenshot 2026-07-31 at 14 35 44
Screenshot 2026-07-31 at 14 32 03
Screenshot 2026-07-31 at 14 28 06
Screenshot 2026-07-31 at 14 27 31

## What I changed and why

- Renamed the Mapeo-specific alerts surface to generic secondary naming throughout: `mapeoData` → `secondaryData`, `MAPEO_CATEGORY_IDS` → `SECONDARY_CATEGORY_IDS`, `mapeoDocId` → `secondaryDocId`, layer `mapeo-data` → `secondary-data`, `feature_type: "mapeo"` → `feature_type: "secondary"`
- Treats Mapeo as one possible secondary source rather than a hardcoded special case at the configuration boundary
- Category filtering is now optional: data is fetched whenever a secondary table is set, and the allowlist is only applied when category IDs are present
- Hex document IDs are auto-detected for Mapbox hashing, replacing the `isMapeoData` flag
- Secondary dataset selector is now limited to geospatial tables
- Legacy `mapeoDocId`, `MAPEO_CATEGORY_IDS`, and `feature_type: "mapeo"` are accepted via migration or silent URL rewrite so existing views and deep links continue to work

I confirmed the generic flow works for any compatible geospatial table while existing Mapeo-backed views remain unchanged through backward compatibility and targeted tests.

## What I'm not doing here

- Rendering secondary datasets on the Map view (config allows it; map API still primary-only)
- Toast when rewriting legacy `mapeoDocId` URLs
- Renaming warehouse table `mapeo_data` or Mapeo schema docs/fixtures

Basically nothing extrenous
## LLM use disclosure

Cursor Grok 4.5 with me driving

### DIFF
--- a/components/config/ConfigAlerts.vue
+++ b/components/config/ConfigAlerts.vue
@@ -20,9 +20,14 @@ const emit = defineEmits<{
 
 type Tag = { text: string };
 
+const resolveCategoryIds = (config: ViewConfig): string | undefined =>
+  config.SECONDARY_CATEGORY_IDS || config.MAPEO_CATEGORY_IDS;
+
+const initialCategoryIds = resolveCategoryIds(props.config);
+
 const initialTags: Record<string, Tag[]> = {
-  MAPEO_CATEGORY_IDS: props.config.MAPEO_CATEGORY_IDS
-    ? props.config.MAPEO_CATEGORY_IDS.split(",").map((tag) => ({ text: tag }))
+  SECONDARY_CATEGORY_IDS: initialCategoryIds
+    ? initialCategoryIds.split(",").map((tag) => ({ text: tag }))
     : [],
 };
 
@@ -34,7 +39,10 @@ const { tags, handleTagsChanged: rawHandleTagsChanged } = updateTags(
 const handleTagsChanged = (key: string, newTags: Tag[]): void => {
   rawHandleTagsChanged(key, newTags);
   const values = newTags.map((tag) => tag.text).join(",");
-  emit("updateConfig", { [key]: values });
+  emit("updateConfig", {
+    SECONDARY_CATEGORY_IDS: values,
+    MAPEO_CATEGORY_IDS: undefined,
+  });
 };
 </script>
 
@@ -47,7 +55,7 @@ const handleTagsChanged = (key: string, newTags: Tag[]): void => {
       >
         {{ $t(toCamelCase(key)) }}
       </label>
-      <template v-if="key === 'MAPEO_CATEGORY_IDS'">
+      <template v-if="key === 'SECONDARY_CATEGORY_IDS'">
         <VueTagsInput
           class="tag-field"
           :tags="tags[key]"
--- a/components/shared/MapLegend.vue
+++ b/components/shared/MapLegend.vue
@@ -95,13 +95,11 @@ watch(
             </div>
             <span>
               {{
-                item.name === "Secondary data"
-                  ? $t("secondaryData")
-                  : item.name === "Most recent alerts"
-                    ? $t("mostRecentAlerts")
-                    : item.name === "Previous alerts"
-                      ? $t("previousAlerts")
-                      : item.name
+                item.name === "Most recent alerts"
+                  ? $t("mostRecentAlerts")
+                  : item.name === "Previous alerts"
+                    ? $t("previousAlerts")
+                    : item.name
               }}
             </span>
           </label>
--- a/components/shared/MapLegend.vue
+++ b/components/shared/MapLegend.vue
@@ -95,8 +95,8 @@ watch(
             </div>
             <span>
               {{
-                item.name === "Mapeo data"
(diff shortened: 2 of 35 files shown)


server/utils/index.ts
```diff
@@ -63,3 +66,35 @@ export const getFilteredTableNames = async () => {
 
   return tableNames;
 };
+
+/**
+ * Warehouse tables that expose both `g__type` and `g__coordinates` (plottable secondary datasets).
+ */
+export const getGeospatialTableNames = async (): Promise<string[]> => {
```
rudokemper, line 73: The addition of this could have been its own PR. Think of how easier it would be to review this overall large PR, AND to follow the commit history of this project. I strongly recommend thinking about this for future PRs.

components/config/ConfigAlerts.vue
```diff
@@ -34,7 +39,10 @@ const { tags, handleTagsChanged: rawHandleTagsChanged } = updateTags(
 const handleTagsChanged = (key: string, newTags: Tag[]): void => {
   rawHandleTagsChanged(key, newTags);
   const values = newTags.map((tag) => tag.text).join(",");
-  emit("updateConfig", { [key]: values });
+  emit("updateConfig", {
+    SECONDARY_CATEGORY_IDS: values,
+    MAPEO_CATEGORY_IDS: undefined,
+  });
```
rudokemper, line 45: The concept of a "category" is unique to Mapeo.

If we are generalizing to any secondary dataset, then we need more general language and handling.

Currently, for Mapeo data, we hardcode a `category` field somewhere, and then in config, allow the user to pick the categories for which they want data to show on the map:

Image

Now we actually need **two** fields:

1. "Filter data by column" (to replaced hard-coded `category` field for Mapeo)
2. "Pick the values to show on the map" (the current equivalent of "Mapeo Category IDs to show on the alerts map")

We actually already _have_ a "Filter data by column" field - we just don't use it for the alerts dashboard.

Confusingly, we also have a "Which values to filter out from the column" which does the **opposite** of what we need for (2). 

But we can reuse the existing "Filter data by column" field, and then turn "Category IDs to show from the secondary dataset on the alerts map" into that?

Maybe this should be a follow-on PR, though.

rudokemper (changes requested):
This accomplishes the glorified "Find & Replace All" for "Mapeo" -> "secondary". But it is missing an important piece of the puzzle which I described in one of my comments. I think we should have another PR (targeting this branch) handle the required changes to config to allow users to pick their filter column, and specify values to **include** on the alerts dashboard.

Once the dust on all this has settled, we can rethink overall config UX in https://github.com/ConservationMetrics/gc-explorer/issues/435

components/shared/MapLegend.vue
```diff
@@ -95,8 +95,8 @@ watch(
             </div>
             <span>
               {{
-                item.name === "Mapeo data"
-                  ? $t("mapeoData")
+                item.name === "Secondary data"
+                  ? $t("secondaryData")
```
rudokemper: Image

Whatever the name of the secondary dataset table is should show here, instead of "Secondary Data" which is not helpful to the end user.

conservationtimothy: Screenshot 2026-08-03 at 19 19 29
Done

components/config/ConfigAlerts.vue
conservationtimothy: Opened a PR https://github.com/ConservationMetrics/gc-explorer/pull/566

rudokemper (approved):

PR #565, opened 2026-07-31
Alerts secondary feature detail fetch 404s: companion table has no view config

### LINKED ISSUE #564: Alerts secondary feature detail fetch 404s: companion table has no view config

While working on https://github.com/ConservationMetrics/gc-explorer/pull/563, I discovered that on the Alerts Dashboard, clicking a secondary-dataset feature (e.g. points from a companion table such as `mapeo_data`) triggers a full-record hydrate for the sidebar via `GET /api/:table/:recordId` where `:table` is the **secondary** warehouse table.

That route always calls `fetchTableConfig(table)` to resolve `ROUTE_LEVEL_PERMISSION`. Companion / secondary tables are stored only as `secondary_dataset` on an alerts view -- they are not `primary_dataset` for any row in `views`. The lookup fails with:

`No view configuration found for table "<secondary_table>"` (HTTP 404)

The alerts map can still render secondary GeoJSON (loaded through the alerts view API), but sidebar detail hydration for those features fails. 

`useViewType` already omits `view_type` for cross-table reads so the server does not look up `(secondary_table, alerts)`, but the record endpoint still requires *some* view config row for the path table as primary.

#### Steps to Reproduce

1. Open an alerts view that has a geospatial secondary dataset configured (e.g. primary `fake_alerts`, secondary `mapeo_data`).
2. Wait for the map to load secondary points.
3. Click a secondary feature (not an alert polygon/point).
4. Observe the server log / network tab: `GET /api/<secondary_table>/<id>` returns 404 with “No view configuration found for table …”.

## Expected Behavior

Selecting a secondary feature should load full record details for the sidebar (or a clear, intentional fallback to the minimal GeoJSON properties already on the map) without requiring the secondary table to be registered as its own view primary. Permission checks should follow the **parent alerts view** (or another defined policy for companion tables), not a missing `views` row for the secondary table name.

conservationtimothy:
## Goal

Stop the alerts sidebar from 404ing when you click a secondary-dataset point (for example `mapeo_data`). Closes #564 

## Screenshots
Screenshot 2026-07-31 at 16 20 11
Screenshot 2026-07-31 at 16 20 01

## What I changed and why

- Fixed full-row lookups for secondary points failing because `/api/<secondary_table>/<id>` tried to resolve view config for the secondary table itself, which is never configured as its own view
- Client now sends the alerts view context (`permission_table` + `view_type`) alongside secondary record requests
- Server uses the alerts view's permissions, verifies the secondary table belongs to that view, then loads the row from the secondary table
- Batch `/records` calls follow the same pattern

The map already had the secondary points; only the follow-up detail request failed, so the data was fine and the permission lookup was wrong. Using the parent alerts view matches how the page is opened. Checking that the secondary table is actually that view’s secondary stops someone from pointing `permission_table` at a public view to read an unrelated table. Same-table fetches still only send `view_type`. Tests cover both the own-table and secondary-table query shapes.

## What I'm not doing here

- Changing how the alerts map loads secondary GeoJSON
- Turning secondary tables into their own views
- Updating other APIs (export, data) the same way

## LLM use disclosure

Cursor Grok 4.5 with me driving

### DIFF
--- a/composables/useRecordCache.ts
+++ b/composables/useRecordCache.ts
@@ -1,4 +1,4 @@
-import { resolveRecordPermissionQuery } from "@/composables/useViewType";
+import { resolveRecordFetchQuery } from "@/composables/useViewType";
 import { encodeDatasetNameForUrl } from "@/utils/identifierUtils";
 
 import type { DataEntry } from "@/types";
@@ -73,13 +73,11 @@ export const useRecordCache = () => {
       return pending.get(cacheKey)!;
     }
 
-    // Same-table view reads send view_type; companion reads send permission_table +
-    // view_type so the parent alerts/map view authorizes the secondary table.
-    const permissionQuery = resolveRecordPermissionQuery(route, table);
+    const query = resolveRecordFetchQuery(route, table);
     const url = `/api/${encodeDatasetNameForUrl(table)}/${encodeURIComponent(recordId)}`;
     const request = (
-      Object.keys(permissionQuery).length > 0
-        ? $fetch<DataEntry>(url, { query: permissionQuery })
+      Object.keys(query).length > 0
+        ? $fetch<DataEntry>(url, { query })
         : $fetch<DataEntry>(url)
     )
       .then((record) => {
@@ -139,14 +137,12 @@ export const useRecordCache = () => {
     }
 
     try {
-      const permissionQuery = resolveRecordPermissionQuery(route, table);
+      const query = resolveRecordFetchQuery(route, table);
       const batchPromises = batches.map((batch) =>
         $fetch<DataEntry[]>(`/api/${encodeDatasetNameForUrl(table)}/records`, {
           method: "POST",
           body: { ids: batch },
-          ...(Object.keys(permissionQuery).length > 0
-            ? { query: permissionQuery }
-            : {}),
+          ...(Object.keys(query).length > 0 ? { query } : {}),
         }),
       );
       const batchResults = await Promise.all(batchPromises);
--- a/composables/useRecordCache.ts
+++ b/composables/useRecordCache.ts
@@ -1,4 +1,4 @@
-import { resolveViewTypeForTable } from "@/composables/useViewType";
+import { resolveRecordPermissionQuery } from "@/composables/useViewType";
 import { encodeDatasetNameForUrl } from "@/utils/identifierUtils";
 
 import type { DataEntry } from "@/types";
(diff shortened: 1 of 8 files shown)


server/database/dbOperations.ts
```diff
@@ -611,6 +611,70 @@ export const fetchTableConfig = async (
...
+  } = {},
+): Promise<ViewConfig> => {
+  const normalizedDataTable = normalizeTableName(dataTable);
+  const permissionTable = options.permissionTable?.trim()
+    ? normalizeTableName(options.permissionTable)
+    : null;
+
+  if (!permissionTable) {
+    return fetchTableConfig(normalizedDataTable, options.viewType);
+  }
```
rudokemper: I find "permission table" to be an unclear term for what you are trying to implement across this PR.

What if you called it `isSecondaryTable`. That would make this block so much clearer, plus what comes next e.g. about fetching `primaryTable` and `secondaryTable`.

composables/useRecordCache.ts
```diff
@@ -73,13 +73,13 @@ export const useRecordCache = () => {
       return pending.get(cacheKey)!;
     }
 
-    // resolveViewTypeForTable returns undefined on purpose (cross-table read, non-view
-    // route, missing :tablename) — callers omit view_type and the server uses its default.
-    const viewType = resolveViewTypeForTable(route, table);
+    // Same-table view reads send view_type; companion reads send permission_table +
+    // view_type so the parent alerts/map view authorizes the secondary table.
```
rudokemper: I have read this three times and I don't understand what this is saying at all, and there is semantic overload going on here.

- What is a "same-table view read"?
- What is a "companion read"? (Is this redundant with "secondary table"?)
- Here is a good example of what i meant in my other comment that the term "permission table" really does not clarify anything.
- "so the parent alerts/map view authorizes the secondary table" -- I have no idea what this means. Views don't have "parents" and they don't authorize tables.

Can you make a strong effort to improve the readability of the code AND comments throughout this entire PR?

rudokemper (changes requested):

composables/useRecordCache.ts
conservationtimothy: Yes, absolutely

server/database/dbOperations.ts
conservationtimothy: I am going to replace `permissionTable` with `primaryDataset`, since we need the primary dataset and view type to find and validate the configured secondary dataset.

composables/useViewType.ts
```diff
@@ -24,6 +24,11 @@ const VIEW_TYPE_BY_SEGMENT: Record<string, ViewType> = {
   alerts: "alerts",
 };
 
+type RecordFetchQuery = {
+  view_type?: ViewType;
+  primary_dataset?: string;
+};
+
```
rudokemper: Remember that we store types in `types/index.ts`.

rudokemper (approved):

PR #566, opened 2026-08-03
Make alerts secondary filters dataset-agnostic

conservationtimothy:
## Goal

Allow alerts views to filter any compatible geospatial secondary dataset using an explicitly configured column and values to include, completing the data-source-agnostic behavior requested in #544.

## Screenshots

image

## What I changed and why

- Alerts config now reuses the existing filtering component and exposes `FRONT_END_FILTER_COLUMN` alongside a new generic `SECONDARY_FILTER_VALUES` field
- Alerts API applies filter values to the explicitly configured secondary dataset column via the existing inclusion filter, replacing the Mapeo-specific behavior that searched for category-named columns
- Added a migration that converts category-based config keys to the generic include-values key and supplies the implicit `p__categoryid` column used by existing Mapeo views
- Runtime compatibility remains in place for configs that have not yet been migrated

The config and API now use the same approach: choose a secondary-dataset column and the values to include. The endpoint test uses a `status` column and confirms that only matching records are returned, so filtering is no longer tied to categories.

I also verified that existing Mapeo configs still work after migration, mapping to `p__categoryid` with the same values. If the new fields aren't set, all secondary geospatial data is returned. All 471 unit tests pass, along with formatting, lint, and migration checks.

## What I'm not doing here

This PR does not change the existing exclusion filtering used by map and gallery views, nor does it alter secondary-dataset selection or geospatial compatibility checks. 

## LLM use disclosure

GPT Sol 5.6 helped


conservationtimothy:
`server/database/migrations/meta/0011_snapshot.json` is 353 LOC which is what blew up this PR same as https://github.com/ConservationMetrics/gc-explorer/pull/563

conservationtimothy:
Trying GitHub's stacked experience 😛

rudokemper (approved):

PR #571, opened 2026-08-05
Gallery: filter entries by media type (photo, audio)

### LINKED ISSUE #454: Gallery: filter entries by media type (photo, audio)
## Feature Request

Add filtering so users can narrow gallery entries by **media type** (e.g. photo, audio), composed with existing filters (category column + date range) as **AND** logic unless product specifies otherwise.

- User can select one or more media types and see only entries that include at least one attachment of those types (define rules clearly in implementation plan).
- Types align with MIME detection / file extensions already implied by `allowedFileExtensions` and `getFilePathsWithExtension`.
- Clearing media-type filter restores full set subject to other active filters.
- i18n strings for new UI labels.

## Implementation Plan

conservationtimothy:
## Goal

Let users narrow the gallery by media type (photo, audio, video), composed with existing filters as AND. Closes #454 

## Screenshots
Screenshot 2026-08-05 at 13 19 34
Screenshot 2026-08-05 at 13 19 27
Screenshot 2026-08-05 at 13 19 20

## What I changed and why

- Added `getMediaTypesForEntry` and `filterByMediaTypes` helpers using allowed extensions with case-insensitive matching, plus an optional media column
- Gallery list API now retains string fields carrying media extensions on each minimal list item when `MEDIA_COLUMN` is unset, so client-side type filtering and initial tile paths have data to classify — previously those fields were stripped, leaving the filter inert while tiles still showed media after full-record fetch
- Added media type checkboxes to the gallery filter panel, applied as AND after date and category filters
- "Clear all" now also clears media type filters

- Manually: open gallery with mixed photo/audio/video, toggle each type and clear all.

## What I'm not doing here

- Server-side media-type prefilter (still client-side after list payload includes media fields).

## LLM use disclosure

Cursor


rudokemper (changes requested):
Thanks @conservationtimothy. My main request is that we turn the "filter by media type" into a shared component. We may end up using this on the Map view, Although we have to think a little bit about available UI real estate there. But I can see this being a useful feature for views beyond the Gallery. Sorry that we didn't think about this while writing the specs, but it feels like an obvious thing seeing the filter screenshot.

Also, we should have a checkbox "None" for the use case where a View has data without media. Obviously that wouldn't come up for the Gallery! But let's add that to the shared component.

(And conditional logic to not render that checkbox, if there is no data that matches the condition, as should be the case for Gallery.)

conservationtimothy:
Done @rudokemper https://github.com/ConservationMetrics/gc-explorer/pull/571/changes/a15fee3bb2c1f14eba59b4a4d6ebfddb51ad2d42

utils/index.ts
```diff
@@ -36,6 +49,81 @@ export const getFilePathsWithExtension = (
...
+export type GalleryMediaType = "audio" | "image" | "video";
+
+/** Media type filter selections; `"none"` = entry has no classified media files. */
+export type MediaTypeFilterValue = GalleryMediaType | "none";
+
+type MediaExtensionConfig = {
+  audio: string[];
+  image: string[];
+  video: string[];
+};
```
rudokemper: Please remember that types have their own file.

utils/index.ts
```diff
@@ -36,6 +49,81 @@ export const getFilePathsWithExtension = (
...
+  if (!selectedTypes.length) return items;
+  return items.filter((item) => {
+    const present = getMediaTypesForEntry(item, allExtensions, mediaColumn);
+    return selectedTypes.some((type) => {
+      if (type === "none") return present.length === 0;
+      return present.includes(type);
+    });
+  });
+};
+
```
rudokemper: Let's turn these into their own mediaHelpers utils file.

tests/unit/components/shared/MediaTypeFilter.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import MediaTypeFilter from "@/components/shared/MediaTypeFilter.vue";
+
+Object.assign(globalThis, { ref, computed, watch });
+
+const extensions = {
+  audio: ["mp3"],
+  image: ["jpg"],
+  video: ["mp4"],
+};
```
rudokemper: Isn't this already defined somewhere?

tests/unit/utils/mediaTypeFilter.test.ts
```diff
@@ -0,0 +1,126 @@
...
+  filterByMediaTypes,
+  getMediaTypeFilterOptions,
+  getMediaTypesForEntry,
+} from "@/utils";
+
+const extensions = {
+  audio: ["mp3", "m4a"],
+  image: ["jpg", "png", "webp"],
+  video: ["mp4", "mov"],
+};
```
rudokemper: Isn't this already defined somewhere?

utils/index.ts
rudokemper: ...that file already exists. Add these there.

rudokemper (approved):
File / code organization requests, please resolve these before merging.

PR #577, opened 2026-08-10
Require view_type on delete_table endpoint

### LINKED ISSUE #511: Require view_type on delete_table endpoint
## Feature Request
`POST /api/config/delete_table/:table` treats `view_type` as optional. Without it, ALL views for the dataset are deleted (leftover from the 1:1 model).

No live UI path omits it today as the edit page always passes `view_type`. But there's dead remove-code in `ConfigDashboard.vue` / `pages/config/index.vue` that calls the endpoint without it

- Delete requests without `view_type` are rejected
- Deleting a view never affects other views on the same dataset
- Dead remove-code removed

## Implementation Plan

conservationtimothy:
## Goal

Require `view_type` on dataset-view deletion so only one `(primary_dataset, view_type)` row is removed. Closes #511 

## Screenshots

N/A — API and config only

## What I changed and why

- Delete-table route now validates `view_type` as exactly one of `alerts`, `map`, or `gallery`, rejecting missing, empty, repeated, and unknown values
- `removeTableFromConfig` always deletes by primary dataset and view type together
- Config edit page always sends the resolved view type on delete so the client matches the server contract

Because `view_type` is now required and validated before reaching `removeTableFromConfig`. Deleting one view type only removes that type, and `public_views` is kept until the last view is deleted. The validation is tested directly using the same `400` error format used elsewhere in dbHelpers, without mocking the database or Nitro handler.

## What I'm not doing here

No changes to `update_config` and no cleanup of unrelated config pages.

## LLM use disclosure

Cursor Grok for tests

### DIFF
--- a/server/utils/dbHelpers.ts
+++ b/server/utils/dbHelpers.ts
@@ -1,9 +1,37 @@
 import { useRuntimeConfig } from "#imports";
 import type { H3Event } from "h3";
+import { VIEW_TYPES, type ViewType } from "@/types";
 import {
   decodeDatasetNameFromUrl,
   normalizeTableName,
 } from "@/utils/identifierUtils";
+
+/**
+ * Validates a raw `view_type` query value as exactly one of alerts, map, or gallery.
+ * Repeated query keys arrive as arrays and are rejected.
+ *
+ * @param {unknown} raw - Raw `view_type` from `getQuery`.
+ * @returns {ViewType} The validated view type.
+ */
+export const parseRequiredViewType = (raw: unknown): ViewType => {
+  if (Array.isArray(raw) || typeof raw !== "string" || raw === "") {
+    throw Object.assign(
+      new Error(
+        "view_type is required and must be a single value: alerts, map, or gallery",
+      ),
+      { statusCode: 400 },
+    );
+  }
+
+  if (!(VIEW_TYPES as readonly string[]).includes(raw)) {
+    throw Object.assign(
+      new Error("view_type must be one of: alerts, map, or gallery"),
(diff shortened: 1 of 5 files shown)


server/utils/dbHelpers.ts
```diff
@@ -1,10 +1,38 @@
...
+      { statusCode: 400 },
+    );
+  }
+
+  if (!(VIEW_TYPES as readonly string[]).includes(raw)) {
+    throw Object.assign(
+      new Error("view_type must be one of: alerts, map, or gallery"),
+      { statusCode: 400 },
+    );
+  }
```
rudokemper, line 31: In the past, we've spoken about trying not to hard code the different view types. As we _might_ add another view type in the future, and then have to make 100s of code changes to reflect that.

To wit, we introduced a type on line 126 of `types/index.ts`:

```ts
export type ViewType = "alerts" | "map" | "gallery";
```

I am wondering how we can similarly define the values once, and use everywhere in places like this. Another example is your unit test that you added here.

I think the best way to do this is to define in `nuxt.config.ts`, e.g.

```ts
  appConfig: {
    viewTypes: ["alerts", "map", "gallery"],
  },
```

And then use that everywhere to output lists of view types, validate against, check types, etc.

Out of scope for this PR but can you make an issue to track this?

rudokemper (approved):

server/utils/dbHelpers.ts
conservationtimothy: https://github.com/ConservationMetrics/gc-explorer/issues/578

PR #579, opened 2026-08-11
Define view types once (e.g. Nuxt appConfig) and reuse everywhere

### LINKED ISSUE #578: Define view types once (e.g. Nuxt `appConfig`) and reuse everywhere
## Feature Request

`ViewType` and `VIEW_TYPES` already live in `types/index.ts`, but other code still hard-codes `"alerts"`, `"map"`, and `"gallery"` (validation messages, unit tests, etc.). Adding a new view type means updating many scattered strings. We should define the list once and reuse it everywhere we list, validate, or type-check view types.

- As a developer adding a new view type, I update one list (proposed: `nuxt.config.ts` → `appConfig.viewTypes`) and validation, tests, and UI lists pick it up without editing many hard-coded strings.
- Acceptance:
  - View type values are defined in one place and used for validation, lists, and related checks (including `parseRequiredViewType` and its unit test).
  - `ViewType` stays derived from that list (or otherwise stays in sync so TypeScript and runtime cannot drift).
  - Existing `alerts` / `map` / `gallery` behavior is unchanged.

### UI Sketch (or Screenshot)

N/A

## Implementation Plan

Proposed shape:

```ts
appConfig: {
  viewTypes: ["alerts", "map", "gallery"],
},
```

Wire validation helpers, tests, and lists to that source.

conservationtimothy:
## Goal

Define view types once (e.g. Nuxt appConfig) and reuse everywhere. Closes #578 

## Screenshots

N/A

## What I changed and why

- View types are now declared once as `viewTypes` in `nuxt.config.ts` and exposed via `useAppConfig().viewTypes`, removing the old `VIEW_TYPES` constant from `types/index.ts`
- Validation (`parseRequiredViewType` and its error text), route-segment matching, `ConfigCard` media/filters visibility, and the create-view type picker all read from the single config list instead of repeating `"alerts"` / `"map"` / `"gallery"`
- Create-view page still holds icon and description metadata per type but builds the option list by mapping over `viewTypes`
- `ViewType` remains importable from `@/types` as a type derived from the same const, for use in props and function signatures only
- Secondary dataset support stays as a separate subset, unchanged

`nuxt.config` is now the single source of truth for view types, with everything using `useAppConfig()` instead of a separate `VIEW_TYPES` list. Tests verify that valid types work, invalid ones list all configured types, and the mock matches the config. Existing alerts, map, and gallery values are unchanged, and all 562 unit tests pass.

## What I'm not doing here

- Not putting secondary-dataset view types on `appConfig`
- Not rewriting every place that intentionally uses a single view type (route handlers, fixtures, MediaFile `variant`, DOM `#map`)
- Not changing alerts / map / gallery behavior beyond reading the shared list

## LLM use disclosure

Used Cursor Grok 4.5 for implementation, after a `grill-me` design pass against #578.

### DIFF
--- a/tests/unit/helpers/importsMock.ts
+++ b/tests/unit/helpers/importsMock.ts
@@ -1,11 +1,17 @@
 import { vi } from "vitest";
 
+import { viewTypes } from "../../../nuxt.config";
+
 // Mock the composables used in LoginScreen component
 export const useUserSession = vi.fn(() => ({
   loggedIn: { value: false },
   session: { value: {} },
   user: { value: null },
   fetch: vi.fn(),
+}));
+
+export const useAppConfig = vi.fn(() => ({
+  viewTypes,
 }));
 
 export const useCustomAuth = vi.fn(() => ({
--- a/tests/unit/helpers/importsMock.ts
+++ b/tests/unit/helpers/importsMock.ts
@@ -1,6 +1,6 @@
 import { vi } from "vitest";
 
-import { viewTypes } from "../../../nuxt.config";
+import { viewTypes } from "@/nuxt.config";
 
 // Mock the composables used in LoginScreen component
 export const useUserSession = vi.fn(() => ({
--- a/tests/unit/utils/viewTypes.test.ts
+++ b/tests/unit/utils/viewTypes.test.ts
@@ -5,7 +5,7 @@ import {
   SECONDARY_DATASET_VIEW_TYPES,
   supportsSecondaryDataset,
 } from "@/types";
-import { viewTypes } from "../../../nuxt.config";
+import { viewTypes } from "@/nuxt.config";
 
 describe("view type capabilities", () => {
   it("exposes the nuxt.config viewTypes list via useAppConfig", () => {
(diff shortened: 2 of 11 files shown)


tests/unit/utils/viewTypes.test.ts
```diff
@@ -1,14 +1,16 @@
 import { describe, expect, it } from "vitest";
 
+import { useAppConfig } from "#imports";
 import {
   SECONDARY_DATASET_VIEW_TYPES,
   supportsSecondaryDataset,
-  VIEW_TYPES,
 } from "@/types";
+import { viewTypes } from "../../../nuxt.config";
```
rudokemper:
```suggestion
import { viewTypes } from "@/nuxt.config";
```

tests/unit/helpers/importsMock.ts
```diff
@@ -1,5 +1,7 @@
 import { vi } from "vitest";
 
+import { viewTypes } from "../../../nuxt.config";
```
rudokemper:
```suggestion
import { viewTypes } from "@/nuxt.config";
```

rudokemper (approved):
Awesome.

tests/unit/utils/viewTypes.test.ts
rudokemper: Unless this doesn't work for tests?

PR #583, opened 2026-08-12
Config: remove "Which values to filter out from the column" field

### LINKED ISSUE #465: Config: remove "Which values to filter out from the column" field
## Feature Request

Or `FILTER_OUT_VALUES_FROM_COLUMN`. I think this is more confusing than helpful. Akin to https://github.com/ConservationMetrics/gc-explorer/issues/460, let's rip it out.

## Implementation Plan

conservationtimothy:
## Goal

Closes #465 by removing the “Which values to filter out from the column” configuration and its associated row-filtering behavior.

## Screenshots

Not included. This removes an existing Config field without introducing new UI.

## What I changed and why

- Removed `FILTER_OUT_VALUES_FROM_COLUMN` and its hidden `FILTER_BY_COLUMN` companion from the config type, Config UI, translations, and documentation
- Removed the corresponding row-filter helper and its callsites in the map and gallery endpoints, including the now-unnecessary projected column
- Removed tests and mocks specific to the retired helper
- Normalized key ordering in locale files.

I traced both config keys through the map and gallery filtering flow and confirmed neither key nor the old helper is used anymore. The relevant tests passed, along with all 561 unit tests, confirming map and gallery still work correctly/

## What I'm not doing here

It does not change front-end filtering or unwanted-column filtering or anything extrenous

## LLM use disclosure

GPT Sol


rudokemper (commented):
Wait, is this also removing "Filter data by column"? Because that should still be available, and passed to the views to populate this component:

Image

We are just removing the ability to _filter out specific values_ from there.

conservationtimothy:
Screenshot 2026-08-12 at 16 15 36
@rudokemper , No.

rudokemper (approved):
Oh, I see, some of that `FILTER_BY_COLUMN` stuff was left-over / vestigial, and the name of the "filter by column" is now `FRONT_END_FILTER_COLUMN`. I think... the naming is a little confusing. (Not a problem to solve here.)

PR #584, opened 2026-08-12
Config: remove Unwanted Substrings field

### LINKED ISSUE #460: Config: remove Unwanted Substrings field
## Feature Request

Since we are doing https://github.com/ConservationMetrics/gc-explorer/issues/438, which will make it possible for the GC Explorer administrator to select which fields they want to not show on the front end view, I think we should do away with the "Unwanted Substrings" (`UNWANTED_SUBSTRINGS`) variable and its field in Config. In programmatic experience, this feature has only caused unexpected edge case problems.

## Implementation Plan

conservationtimothy:
## Goal

Closes #460 by removing the Unwanted Substrings configuration and cleaning existing stored configurations.

## Screenshots

Not included. This removes an existing Config field without introducing new UI.

## What I changed and why

- Removed `UNWANTED_SUBSTRINGS` from the Config UI, configuration type, translations, documentation, and gallery and alerts filtering paths
- Exact column-name filtering via `UNWANTED_COLUMNS` is unchanged
- Added migration `0012_strip_unwanted_substrings.sql` to remove the deprecated key from existing `views.view_config` JSON objects
- Normalized key ordering in locale files

I traced the setting from the Config editor through `ViewConfig` into gallery and alerts filtering and confirmed there are no active references left outside the cleanup migration. Existing tests still cover unwanted-column filtering, and the relevant endpoint tests plus all 562 unit tests pass.

## What I'm not doing here

This does not remove `UNWANTED_COLUMNS`, alter exact column-name filtering.

## LLM use disclosure

GPT Sol


rudokemper (approved):
I am surprised the i18n files alphabetization was so off.

Issue #438, opened 2026-05-11
Config: API returns columns for config views, UI uses them for single-select dropdowns


rudokemper:
### User Stories & Acceptance Criteria

- For `UNWANTED_COLUMNS`, some fields, like `_id`, should never be possible to exclude. Also, any of the field choices that are set for "Color column", "Column to filter values from", "Timestamp column", "Column to use for media files", "Icon column" -- if these are set, we should remove them from the possible list of typeaheads for `UNWANTED_COLUMNS` and have good validations in place to prevent a user from saving a config state where these columns are included IF they are set.

## Implementation Plan
(may be added later, before work begins)

PR #596, opened 2026-08-19
Config: return dataset column metadata

conservationtimothy:
## Goal

Return original and SQL column names for Config controls.

Part of #438.

> Stack: 1 of 4. This PR targets `feat/438-config-column-selectors-integration`.

## Screenshots

Not applicable

## What I changed and why

I added an admin-only endpoint that returns column metadata for one warehouse table. The database helper uses the columns metadata table when it is available. It uses `information_schema.columns` as a fallback and does not fetch dataset rows.

The endpoint uses the existing table-name normalization and Config permission checks. and the fallback preserves the SQL column name as the original name when no mapping table exists.

## What I'm not doing here

I am not adding save validation or changing Config controls in this PR.

## LLM use disclosure

### DIFF
--- a/server/api/config/columns/[table].get.ts
+++ b/server/api/config/columns/[table].get.ts
@@ -1,0 +1,17 @@
+import { fetchTableColumnEntries } from "@/server/database/dbOperations";
+import { getTableParam } from "@/server/utils/dbHelpers";
+import { validatePermissions } from "@/utils/accessControls";
+
+import type { H3Event } from "h3";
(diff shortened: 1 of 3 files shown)


rudokemper (approved):
Great, thanks.

I think it would be helpful to leave a comment about the fallback behavior via `fetchTableSqlColumns`; that the column names are derived from `information_schema.columns` via that fn, if no `__columns` table is found.

PR #597, opened 2026-08-19
Config: validate configured column selections

conservationtimothy:
## Goal

Reject unavailable, protected, or conflicting Config column values.

Part of #438.

> Stack: 2 of 4. This PR is based on `feat/438-column-metadata-api`.

## Screenshots

Not applicable. This PR does not change the interface.

## What I changed and why

I added shared column validation to Config create and update requests. The validation uses the correct primary or secondary dataset, protects unwanted columns, rejects conflicts with active fields, and rejects `_id` as a functional column. 

The validator resolves original names to SQL names before it checks conflicts. Alerts filter values use the selected secondary dataset, while other functional fields use the primary dataset. 

## What I'm not doing here

I am not adding dropdown components or changing Config forms in this PR.

## LLM use disclosure

### DIFF
--- a/server/api/config/update_config/[table].post.ts
+++ b/server/api/config/update_config/[table].post.ts
@@ -23,7 +23,11 @@ export default defineEventHandler(async (event: H3Event) => {
   } catch (error) {
     if (error instanceof Error) {
       console.error("Error updating config on API side:", error.message);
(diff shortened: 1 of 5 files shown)


rudokemper (approved):
Very good. I am "just approving" but I did read all of the code!

PR #600, opened 2026-08-20
Config: add dataset-backed column controls

conservationtimothy:
## Goal

Add reusable controls that load and display valid dataset columns.

Part of #438.

> Stack: 3 of 4. This PR is based on `feat/438-column-validation`.

## Screenshots

Not included. 

## What I changed and why

I added a reactive dataset-column loader and a shared column select component. The loader ignores stale responses when the selected dataset changes. The select shows original and SQL names, keeps unavailable saved values visible as errors, and does not offer `_id`.

Manual testing

## What I'm not doing here

I am not replacing existing Config fields in this PR.

## LLM use disclosure

### DIFF
--- a/composables/useDatasetColumns.ts
+++ b/composables/useDatasetColumns.ts
@@ -1,0 +1,58 @@
+import { encodeDatasetNameForUrl } from "@/utils/identifierUtils";
+
+import type { ColumnEntry } from "@/types";
+import type { Ref } from "vue";
+
+type DatasetColumnsResponse = {
+  columns: ColumnEntry[];
+  table: string;
+};
+
+/**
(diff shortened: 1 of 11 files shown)


composables/useDatasetColumns.ts
```diff
@@ -0,0 +1,58 @@
+import { encodeDatasetNameForUrl } from "@/utils/identifierUtils";
+
+import type { ColumnEntry } from "@/types";
+import type { Ref } from "vue";
+
+type DatasetColumnsResponse = {
+  columns: ColumnEntry[];
+  table: string;
+};
```
rudokemper: You already know what I would write.

rudokemper (approved):

PR #601, opened 2026-08-20
Config: use dataset columns in forms

conservationtimothy:
## Goal

Use dataset columns in Config dropdowns and the unwanted-column typeahead.

Part of #438.

> Stack: 4 of 4. This PR is based on `feat/438-column-controls`.

## Screenshots

Screenshot 2026-08-19 at 14 33 46
Screenshot 2026-08-19 at 14 32 57

## What I changed and why

I replaced free-text column fields with dataset-backed selects and added valid suggestions to the unwanted-column field. Color, icon, timestamp, and media fields use primary columns. The Alerts filter uses the selected secondary dataset when it is available. Save remains disabled while metadata loads or while configured columns are invalid.

Checking out the UI myself

## What I'm not doing here

I am not combining primary and secondary options or redesigning unrelated Config sections.

## LLM use disclosure

### DIFF
--- a/components/config/ConfigCard.vue
+++ b/components/config/ConfigCard.vue
@@ -2,10 +2,12 @@
 import { useAppConfig } from "#imports";
 import {
   supportsSecondaryDataset,
+  type ColumnEntry,
   type ViewConfig,
   type ViewType,
 } from "@/types";
 import { CONFIG_LIMITS } from "@/utils";
+import { validateViewConfigColumns } from "@/utils/viewConfigColumns";
 import ConfigPermissions from "./ConfigPermissions.vue";
 import ConfigCollapsibleSection from "./ConfigCollapsibleSection.vue";
 import { Check, Trash2 } from "lucide-vue-next";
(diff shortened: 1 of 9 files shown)


rudokemper (approved):
Great.

Image

Assuming this is a legacy screenshot? As in a previous PR you explicitly omitted `_id`.

Do we know what happens if there is currently an unavailable column set for the field? Like, if somebody accidentally wrote "catagory_id" but the column name is _category_id"

conservationtimothy:
> Great.
> 
> Image
> Assuming this is a legacy screenshot? As in a previous PR you explicitly omitted `_id`.
> 

Yeah it *is* a legacy screenshot. 

> Do we know what happens if there is currently an unavailable column set for the field? Like, if somebody accidentally wrote "catagory_id" but the column name is _category_id"

If a field is set to something like `catagory_id` and the real SQL column is `_category_id`:

1. Select still shows the bad value as a disabled option (so the select doesn’t go blank).
2. Red border + `columnNotAvailable` (“This column is not available in the selected dataset.”).
3. Save stays disabled -- `ConfigCard` requires `columnValidation.isValid`.
4. API would reject it anyway on create/update with something like `COLOR_COLUMN references unavailable column "catagory_id"`.

PR #604, opened 2026-08-20
[Config] Add a Button that takes the user to the edit screen on their Mapbox studio style editor

### LINKED ISSUE #485: [Config] Add a Button that takes the user to the edit screen on their Mapbox studio style editor
For the average user, getting into mapbox's infrastructure can be daunting, specially if they only need to do surgical edits and not get to learn the platform.

I want to document this ergonomic improvement, although there is no user that backs it, for us to weight it in.

In this screen:

Image

To add a button to the right of the style that could take you to the Mapbox studio style editor screen, by appending the style url to "https://console.mapbox.com/studio/" by just removing the "mapbox://" part of the url.

I tried it already and it works.

conservationtimothy:
## Goal

Close #485 -- show a control next to each Config Map basemap style field that opens that style in Mapbox Studio when the value starts with `mapbox://styles/`.

## Screenshots
Screenshot 2026-08-20 at 14 58 35

## What I changed and why

This change adds a small util that turns a `mapbox://styles/...` value into a Studio URL, and shows an external-link control beside each style input only when that util returns a URL. The link opens in a new tab. Other or empty style values keep the field as it was, with no link.

Manual testing

## What I'm not doing here

- Mapbox access token fields or token validation
- API changes

So basically -- nothing extrenous

## LLM use disclosure
Used Cursor Grok 4.6 for i18n.


utils/mapboxStyle.ts
```diff
@@ -0,0 +1,20 @@
+const MAPBOX_STYLE_PREFIX = "mapbox://styles/";
```
rudokemper: I know it's not _quite_ the same thing but I'd prefer this go into `mapGLHelpers` as it's helper code for working with Mapbox (GL). 

Otherwise, this util file will probably never have any other content in it.

rudokemper (approved):
Great. Just one piece of feedback to address before merging.

utils/mapboxStyle.ts
rudokemper: To clarify, when I say "this", I mean all the code in the file. Move to `mapGLHelpers`, remove `mapboxStyle.ts`

PR #608, opened 2026-08-23
Remove `UNWANTED_COLUMNS` from GC Explorer, add migration to update dbs

### LINKED ISSUE #607: [Config] Rip out UNWANTED_COLUMNS
## Feature Request

I think `UNWANTED_COLUMNS` is more confusing than it is helpful. The goal was to declutter the sidebar but I think this can instead be solved through https://github.com/ConservationMetrics/gc-explorer/issues/582.

## Implementation Plan

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-explorer/issues/607.

## What I changed and why

`UNWANTED_COLUMNS` was more confusing than useful, so I removed it end to end: config UI, types, save-time validation, gallery/alerts filtering (`filterUnwantedKeys`), docs, i18n, and tests.

Existing view configs still store the key in `view_config` JSON, so I added migration `0013_strip_unwanted_columns.sql` to drop it, same approach as `0012_strip_unwanted_substrings.sql`.

- Grepped the repo: `UNWANTED_COLUMNS` remains only in the new migration SQL.
- Unit tests.
- Followed `acfda4b` (#583) for ripping out a config field, and `f0ca24b` (#584) for the JSON-strip migration.

## LLM use disclosure

I had Cursor Grok 4.5 check the previous commits for removing `UNWANTED_SUBSTRINGS` and emulate that code to generate this.


conservationtimothy (approved):
Code is good but there are merge conflicts.

PR #611, opened 2026-08-23
Config: Harden API validations and error screen behavior

### LINKED ISSUE #464: Config: Harden API validations and error screen behavior
## Feature Request

https://github.com/ConservationMetrics/gc-explorer/issues/438 will do a lot of good to eliminate possible edge cases where users provide column headers that do not exist. But I think we should also make an effort to harden API behavior, and when we show error screens to the user.

Currently:

- If I provide a value for `FRONT_END_FILTER_COLUMN`, `COLOR_COLUMN`, or `ICON_COLUMN` that does not exist in the table,
  - `GalleryView` renders the data fine.
  - `MapView` throws an error in logs (note `fake_column` is my value for "Column to filter values from"): `ERROR  Error fetching data from table: Failed query: SELECT "_id", "g__type", "g__coordinates", "fake_column", "end", "_submission_time" FROM "bcmform_responses" LIMIT $1`
    - The front end _still_ loads, but shows "Total features: 0".
    - I would expect to see our "Error" screen

Why do this if we are already improving the Config UI? Imagine a use case where a Guardian Connector admin changes the column names -- like, I change my `color` field to `_color` -- GC Explorer is not going to know about that change, and the same error behavior would show up.

Also:

Currently, when some config fields (e.g. for Map) are incorrectly filled out, the front end simply shows an error "It was not possible to load this data" and the console simply reports:

```
Error fetching data: Yr: [GET] "/api/<dataset>/map?limit=10000": 404 
Caused by: FetchError: [GET] "/api/<dataset>/map?limit=10000": 404
```

It would be helpful to know more about what went wrong. For example, you get this error if you failed to provide a correct Mapbox Access Token.

conservationtimothy:
## Goal

When a Map or Gallery view config names a warehouse column that is not on the table, the view must throw an error. The shared Error screen must show a clear, localizable reason. Closes #464 

## Screenshots

Screenshot 2026-08-23 at 20 16 37

## What I changed and why

- Map and Gallery now validate configured columns against `information_schema.columns` before running SELECT, collecting all missing fields and returning a structured 422 with `errorCode`, field/column pairs, table name, and an English `statusMessage` — replacing a silent `fetchDataFromTable` failure that returned `[]` and a 200
- Unset optional columns remain valid; Gallery only runs the check when `MEDIA_COLUMN` is set since that is when those columns are projected
- Map, Gallery, and Alerts pages map the 422 payload to i18n templates: a known `errorCode` uses single- or multi-column error strings, other failures fall back to `statusMessage` then generic load-failure copy

I know it is right because he Error screen to show the specific error when available, then the server message, then the generic fallback and we stop swallowing errors

## What I'm not doing here

- A new Error screen layout
- Changes to Config column dropdowns
- A global change to the `fetchDataFromTable` empty-array swallowing error

## LLM use disclosure

Cursor Grok 4.6 with me driving


utils/errorMessages.ts
```diff
@@ -0,0 +1,68 @@
+import { VIEW_CONFIG_MISSING_COLUMNS_ERROR } from "@/types";
```
rudokemper, line 1: Let's rename to `errorMessages.ts` so it can be a more generally useful container.

rudokemper (approved):

PR #619, opened 2026-08-24
Show gear icon routing to config on views, for Admin users only

### LINKED ISSUE #618: [Config] For Admin role, add a gear to access config directly from the view.
## Feature Request

From the view itself e.g. `/alerts/<view>`, if the user is an Admin role, add a gear to access config directly so that they don't have to go back to the index to get there.

## Implementation Plan

rudokemper:
## Goal

Closes #618.

Most of the code line additions are in tests.

## Screenshots

image

## What I changed and why

- Moved the admin/CI/authStrategy check out of `pages/index.vue` into `composables/useCanManageConfig.ts` (with unit tests) so views can reuse it.
Added `AdminConfigGear.vue` next to titles in `GalleryView`, `MapIntroPanel`, and `AlertsIntroPanel`; hidden for non-admins, opens that view’s config in a new tab, and uses localized “Access Config” (also on `DatasetCard`).

Trying in runtime, and tests.

## LLM use disclosure

Cursor Auto did all the work based on my prompts.


conservationtimothy (approved):

conservationtimothy:
@rudokemper Just fix lints please. Code is good.

PR #622, opened 2026-08-25
Replace mock heavy tests with DB API tests

### LINKED ISSUE #502: Replace mock heavy tests with DB API tests
## Feature Request

Replace the mock-heavy view dataset endpoint unit tests with DB-backed API tests that use the existing Playwright test database setup.

As a reviewer, I want the gallery/map dataset API tests to exercise the real config DB and warehouse DB path, so the tests prove endpoint behavior instead of only proving mocked function calls.

Acceptance criteria:

- Add a Playwright API spec that calls the real gallery/map endpoints against seeded test data.
- Use existing `guardianconnector` and warehouse fixtures from the e2e setup.
- Assert that the endpoint returns `primary_dataset`.
- Assert that `table` matches the resolved `primary_dataset`.
- Assert that the response has the expected view payload shape.
- Remove or reduce mock-heavy endpoint tests that only assert `fetchViewTables` / `fetchData` interactions.

### UI Sketch (or Screenshot)

N/A. This is test coverage for API behavior.

## Implementation Plan

Use the existing Playwright setup. Add a spec like `tests/e2e/11-view-dataset-api.spec.ts` that calls endpoints with `page.request.get(...)`, using seeded rows from `tests/db-seed/guardianconnector.sql` and warehouse tables from `tests/db-seed/warehouse.sql`.

For gallery, call a seeded gallery endpoint such as `/api/seed_survey_data/gallery` and assert `primary_dataset`, `table`, and gallery payload shape. For map, call `/api/bcmform_responses/map` and assert the same dataset contract plus map payload shape. Keep helper-level unit tests only if they remain small and meaningful; move endpoint behavior coverage to Playwright.

conservationtimothy:
## Goal

Replace the mock-heavy view dataset endpoint unit tests with DB-backed API tests that use the existing Playwright test database setup. Closes #502 

## Screenshots

N/A. This change is test coverage for API behavior.

## What I changed and why

I added `tests/e2e/12-view-dataset-api.spec.ts`. The spec uses the existing admin Playwright session and calls `GET /api/seed_survey_data/gallery` and `GET /api/bcmform_responses/map`. It asserts that `primary_dataset` matches the seed table, that `table` equals `primary_dataset`, and that the payload has the gallery or map shape.

I removed `tests/unit/server/galleryEndpointDatasets.test.ts` and `tests/unit/server/mapEndpointDatasets.test.ts` since those files mocked the database.

By testing the new spec.

## What I'm not doing here

Nothing extrenous

## LLM use disclosure

Cursor Grok 4.6 helped


IamJeffG (approved):

IamJeffG (commented):
Actually before merging I have a question: Is Playwright actually needed for the assertions you're making here?  as I understand these tests are only testing backend API behavior. You don't need a browser for that, and I'd argue you don't need Playwright for it either.

At least this is my understanding of what this PR is doing and what Playwright is for.  Correct me if I'm wrong.

IamJeffG:
cc @Atmosfearful for visibility

conservationtimothy:
@IamJeffG , Playwright is used because it already boots the app, seeds both DBs, and (for member routes that require auth e.g. `/api/bcmform_responses/map`) already has a real session. 

Right now, `pnpm test:unit` never starts Nuxt and never opens Postgres. It runs in jsdom with mocked `#imports`. Hitting `/api/...` against seeded tables would need a new job: compose up, wait for the app, seed, then fetch.

IamJeffG:
I see your point.
- `pnpm test:unit` is just for unit tests that require no scaffolding
- your PR adds this repo's _first_ "integration tests" of the backend API.
- Playwright is already used for e2e integration tests of backend+frontend.

And all the "integration" style setup that's needed for the backend API is already coded up for use by Playwright.  Thank you, I understand better now.


A tweak I suggest you look into: Playwright also ships a browserless request fixture that can reuse the same admin.json cookie file without launching Chromium at all.

https://playwright.dev/docs/api-testing#reusing-authentication-state
https://playwright.dev/docs/api/class-fixtures#fixtures-request


So with what I'm suggesting you can still do globalSetup and use playwright to setup auth as you do already, but I would be thrilled if these particular tests just didn't have to call `browser.newContext()` and `context.newPage()`.

IamJeffG:
If you are able to go that route (and actually, even if you aren't) I'd add a small blurb to the "test" section of the README enumerating these 3 different classes of tests.

conservationtimothy:
Done in this commit https://github.com/ConservationMetrics/gc-explorer/pull/622/commits/b7b93853205de500392196f02bfe203fece39580

README.md
```diff
@@ -81,7 +81,11 @@ docker run --env-file=.env -it -p 8080:8080 guardianconnector-explorer:latest
 
 ## Testing
 
-GuardianConnector Explorer uses two testing frameworks:
+GuardianConnector Explorer has three classes of tests:
+
+1. **Unit and component tests (Vitest).** Isolated logic with mocked dependencies. No Nuxt server and no database. `pnpm test:unit`.
+2. **Backend API tests (Playwright `request`).** HTTP calls against the running app and seeded databases. Auth0 setup still uses a browser once (`auth.setup.ts`); the tests themselves reuse `admin.json` on the browserless `request` fixture and do not open a page. Same Docker stack as E2E (`NUXT_TEST=true`). Example: `tests/e2e/12-view-dataset-api.spec.ts`.
+3. **End-to-end tests (Playwright browser).** Full UI through backend and database.
```
Atmosfearful, line 88: Useful docs, nice.

tests/unit/server/mapEndpointDatasets.test.ts
```diff
@@ -1,270 +0,0 @@
...
-    expect(hoisted.fetchTableConfig).toHaveBeenCalledWith(thaiTable, "map");
-    expect(hoisted.fetchViewTables).toHaveBeenCalledWith(thaiTable, "map");
-    expect(hoisted.fetchTableSqlColumns).toHaveBeenCalledWith(thaiTable);
-    expect(hoisted.fetchData).toHaveBeenCalledWith(
-      thaiTable,
-      expect.objectContaining({ limit: 25 }),
-    );
-  });
-
-  it("returns 422 when a configured column is missing", async () => {
```
Atmosfearful, line 197: It looks like these 422 test cases were not ported over to the new tests. Intentional?

conservationtimothy: A 422 there would need a broken seed row, which we did not add / have.

IamJeffG: From experience, we will not be able to get far with a single shared set of test data for _all_ API tests.

The correct solution here is that the individual test (inside the `it(..)`, or inside a scoped `beforeAll(..)`) will set up the data that this particular test case needs.

So in this case, specifically the "returns 422" test would set up a broken role in its setup.  Most test runners will do all that in a transaction that gets rolledback after the test finished (i.e. it manages teardown for you) but I am not 100% certain that this is what Jest (or playwright) does -- if not then you also need an `afterAll`.

IamJeffG: I'd go further to suggest that _most_ data seeding happen in the test files themselves, not a global fixture.

Atmosfearful: Agreed, typically I rely on seed scripts just to populate QA and local dev environments (with light smoke tests to solidify the seed scripts), then for the unit and e2e tests I have them own all their own data entirely (via reuseable setup/teardown/fixture helpers)

conservationtimothy: Created an issue: https://github.com/ConservationMetrics/gc-explorer/issues/627

PR #626, opened 2026-08-26
Use background image from `gc_settings`

rudokemper:
## Goal

Use the same `gc_settings.background_image` value as GC Landing Page for Explorer's login background, instead of `NUXT_PUBLIC_BACKGROUND_IMAGE`.

## What I changed and why

Dropped the env-based background. Login now reads `background_image` from `gc_settings` through `GET /api/background-image`. If that row is empty or the table isn't there, it falls back to `/background.jpg` (same bundled image as Landing Page).

The fetch lives in `useLoginBackground`, same pattern as Landing Page's `useThemeSettings`, without bringing over theme admin/logo settings. The Drizzle mapping is read-only and not wired into Explorer migrations, since Landing Page owns `gc_settings`.

- Unit tests and checking it in Runtime.

## What I'm not doing here

- No theme admin UI, logo URL, or writes to `gc_settings`, nor a migration for `gc_settings`.

## LLM use disclosure

Cursor Auto

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -42,6 +42,3 @@ NUXT_PORT=8080
 # Public App Variables
 # ==========================================
 NUXT_PUBLIC_COMMUNITY_NAME=your_community_name
-
-# optional: Login page background image URL or path (falls back to public/background.jpg if not set)
-NUXT_PUBLIC_BACKGROUND_IMAGE=
--- a/components/LoginScreen.vue
+++ b/components/LoginScreen.vue
@@ -1,6 +1,5 @@
 <script lang="ts" setup>
-import { computed } from "vue";
-import { onMounted, useI18n, useRuntimeConfig, useUserSession } from "#imports";
+import { onMounted, useI18n, useUserSession } from "#imports";
 import GlassCard from "@/components/shared/GlassCard.vue";
 import GlobeLanguagePicker from "./shared/GlobeLanguagePicker.vue";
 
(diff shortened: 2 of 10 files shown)


rudokemper:
Hm, why would this PR trigger this CI build error?

```
Error: ERR_PNPM_IGNORED_BUILDS

  × installing dependencies
  ╰─▶ Ignored build scripts: @parcel/watcher@2.5.1, esbuild@0.18.20,
      esbuild@0.25.5, sharp@0.32.6, unrs-resolver@1.9.2, vue-demi@0.14.10
  help: Run "pnpm approve-builds" to pick which dependencies should be allowed
        to run scripts.
```

Is it because of a discrepancy with a very recent version of `pnpm` being pulled in to build?

@conservationtimothy any ideas?

conservationtimothy:
> Hm, why would this PR trigger this CI build error?
> 
> ```
> Error: ERR_PNPM_IGNORED_BUILDS
> 
>   × installing dependencies
>   ╰─▶ Ignored build scripts: @parcel/watcher@2.5.1, esbuild@0.18.20,
>       esbuild@0.25.5, sharp@0.32.6, unrs-resolver@1.9.2, vue-demi@0.14.10
>   help: Run "pnpm approve-builds" to pick which dependencies should be allowed
>         to run scripts.
> ```
> 
> Is it because of a discrepancy with a very recent version of `pnpm` being pulled in to build?
> 
> @conservationtimothy any ideas?

the reason is **CI was not pinned to a specific pnpm version**: `npm install -g pnpm` installed pnpm **10.34.5 [earlier today](https://github.com/ConservationMetrics/gc-explorer/actions/runs/32975716590/job/98199831814?pr=624)**, where these ignored build scripts were only a warning, but it now installs **pnpm 11**, where the same issue fails the install with `ERR_PNPM_IGNORED_BUILDS`. This lines up with pnpm’s official [[v10 → v11 migration guide](https://pnpm.io/migration)](https://pnpm.io/migration), which documents the breaking changes to dependency build-script permissions in v11. The immediate fix is to pin pnpm to `10.34.5`, then upgrade to v11 separately.

rudokemper:
Okay. So unrelated to the PR. Thanks. 

Can you file a separate issue to resolve?

conservationtimothy (approved):

PR #631, opened 2026-08-27
Hide incidents from users below Member

### LINKED ISSUE #593: [Alerts] Hide Incidents buttons when view is public

Investigating https://github.com/ConservationMetrics/gc-explorer/issues/586 I noticed that the Incidents link didn't work because I wasn't logged in. The view is public. But the incidents wouldn't load:

Image

I _think_ that is correct: we shouldn't expose any incidents functionality to `SignedIn` or `Guest` roles.

Depending on where we land, the solution to this issue should make sure

1. API endpoints for Incidents are guarded: Member role and higher only
2. Incidents UI buttons and any screens are hidden for the dashboard when its permission level is less than what we land on.

rudokemper:
## Goal

Closes #593 

Keep incidents off public alerts views for anyone below Member. The API and UI were still exposing the feature to SignedIn, Guest, and logged-out users, which showed a failed Saved Incidents panel.

## What I changed and why

- Guarded incidents and collections API routes with `validatePermissions(event, "member")` so listing, fetching, creating, updating, and deleting incidents is Member+ only.
- Added `useCanAccessIncidents` and hid the alerts dashboard incidents controls, sidebar, and remove modal unless the user is Member/Admin (or `authStrategy` is `none`).
- Skipped incident fetches and `?incidentId=` opens when the user cannot access incidents, so public views no longer request data they cannot see.

- Unit tests cover every incidents/collections route: Member is required, and handlers are not called when that check fails.
- Composable tests cover Member/Admin allowed and SignedIn, Guest, logged-out, missing user, and missing role denied.
- Dashboard tests confirm controls/modal stay hidden, incidents are not fetched, URL incident IDs are ignored without access, and Member-level access still loads incidents on map init.

## LLM use disclosure

Cursor Auto


composables/useHasRole.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    if (authStrategy === "auth0" && loggedIn.value && user.value) {
+      const typedUser = user.value as User | null;
+      const userRole = typedUser?.userRole ?? Role.SignedIn;
+      return userRole >= Role.Member;
+    }
+
+    return false;
+  });
+};
```
conservationtimothy, line 34: Feel like this can be more generic? I think beyond Incidents, we might, in the future, want to gate some functionality behind member nd above.

conservationtimothy (changes requested):

composables/useHasRole.ts
rudokemper: Good callout -- adjusted to `useHasRole` generic (with default "Member") in https://github.com/ConservationMetrics/gc-explorer/pull/631/changes/5f2a69745bb150888ae70fb243c1586e0cdeea27#diff-7d3a9c8cf369b215d6fe46030be76ec8fbcb09eb34cc84cdef8219a8d629fe0d

conservationtimothy (approved):

PR #635, opened 2026-08-27
Restore pulsing halo on most recent alerts

### LINKED ISSUE #587: [Alerts] Blinking of new mining activity
One of our partners uses GC-Explorer to expose mining activity to the mining ministry. 

It is relevant for them to they noticed that when the alerts were blinking the behavior that we used to have where a red dot had a aura that was moving that caught the attention of the mining ministry and made them react positively by acting on those alerts as it felt made them feel urgent and reactive to those alerts.

It would be good to evaluate whether this is a regression that needs to be retracted.

rudokemper:
## Goal

Closes #587.

Bring back the expanding red halo on most recent alerts (individual dots and clusters) so they read as urgent again. Clustering had dropped the old DOM pulse in favor of static red markers.

## Screenshots

image

## What I changed and why

- Added Mapbox circle halo layers under most-recent alert points, clusters, and centroids.
- Radius stays locked to the marker size (plus a small gap). Only stroke width and opacity animate over 2s, so the ring stays tight and does not jump or restart halfway.
- Halo layers follow the most-recent legend visibility toggle and are excluded from click/hover handling.
- Animation is stopped when the dashboard unmounts.

- Runtime and tests

## LLM use disclosure

Cursor Auto agent.

### DIFF
--- a/components/AlertsDashboard.vue
+++ b/components/AlertsDashboard.vue
@@ -18,6 +18,14 @@ import {
   toggleLayerVisibility as utilsToggleLayerVisibility,
   resolveTerrainExaggeration,
 } from "@/utils/mapGLHelpers";
+import {
+  addPulsingHaloLayers,
+  CENTROID_CLUSTER_HALO_RADIUS,
+  CENTROID_HALO_RADIUS,
+  POINT_CLUSTER_HALO_RADIUS,
+  POINT_HALO_RADIUS,
+  stopPulsingHalo,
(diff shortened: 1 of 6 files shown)


conservationtimothy:
@Atmosfearful , A good PR to get into Alerts 🙂

conservationtimothy (approved):

rudokemper:
Merging so we can get it in on time for the monthly product review, but happy to discuss anything in the code :slightly_smiling_face:

PR #636, opened 2026-08-27
Playwright API tests should insert and tear down their own data

### LINKED ISSUE #627: Playwright API tests should insert and tear down their own data
## Feature Request

Do not use one shared seed as the data for all backend API tests. Each API spec (inside the test, or in a scoped `beforeAll`) must insert the view config and warehouse rows that case needs. Put that seeding in the test file, not only in `tests/db-seed/guardianconnector.sql` / `globalSetup`.

Playwright does not wrap Postgres in a transaction that rolls back after the test so we should pair every insert with `afterAll` / `afterEach` teardown. Use unique view names so parallel tests do not clash.

This applies to current API specs (`tests/e2e/12-view-dataset-api.spec.ts`, `tests/e2e/13-alerts-dataset-api.spec.ts`) and to new API specs. One example is a 422 case that inserts a broken view config (a configured column missing on the warehouse table). Happy-path gallery, map, alerts, and record tests should insert their own valid rows the same way, instead of depending on the global fixture.

UI e2e can keep the global seed until a later change.

**As a** developer writing a backend API Playwright test  
**I want** that test to insert the data it needs and delete it when it finishes  
**So that** API cases do not share one fixture set and can cover both happy path and error path.

- **Given** the Docker test stack is up (`NUXT_TEST=true`)  
  **When** an API spec runs  
  **Then** it does not depend on a global `views` row that other API specs also own  
  **And** teardown deletes the rows that spec inserted.

### UI Sketch (or Screenshot)

N/A (test infrastructure).

## Implementation Plan

conservationtimothy:
## Goal

Make API test create and remove its own database records so test runs do not depend on shared view configurations. Closes #627 

## Screenshots

Not applicable. 

## What I changed and why

I added shared Docker test database settings and helpers that create a unique warehouse table and matching view row for each API test in `gc-exp/tests/e2e/12-view-dataset-api.spec.ts` (our only API test for now). The helpers delete both resources after each test group. I updated the gallery and map API tests to use these fixtures and added coverage for a view configuration that refers to a missing column.

I ran the updated API spec against the Docker Compose stack. All seven setup and API tests passed, including the new 422 case. After the run, I queried the config and warehouse databases and confirmed that no `api_test_` rows or tables remained. 

## What I'm not doing here

This change does not modify browser E2E fixtures or remove the global database seed. It does not change application API behavior or production database code.

## LLM use disclosure

GPT Sol 5.6 Helped

### DIFF
--- a/tests/e2e/helpers/apiTestData.ts
+++ b/tests/e2e/helpers/apiTestData.ts
@@ -1,0 +1,94 @@
+import { randomUUID } from "node:crypto";
+
+import type { ApiTestView, ApiTestViewInput } from "@/types";
+import {
+  createTestDatabaseClient,
+  TEST_CONFIG_DATABASE,
+  TEST_WAREHOUSE_DATABASE,
+} from "./testDatabase";
+
+const quoteIdentifier = (value: string): string =>
+  `"${value.replaceAll('"', '""')}"`;
+
+/**
+ * Creates a unique warehouse table and its matching view row for one API test.
+ *
+ * @param input - Source table and view fields needed by the test.
+ * @returns {Promise<ApiTestView>} Identity used by the route and teardown.
+ */
+export const createApiTestView = async (
+  input: ApiTestViewInput,
+): Promise<ApiTestView> => {
+  const primaryDataset = `api_test_${randomUUID().replaceAll("-", "").slice(0, 12)}`;
+  const configSql = createTestDatabaseClient(TEST_CONFIG_DATABASE);
+  const warehouseSql = createTestDatabaseClient(TEST_WAREHOUSE_DATABASE);
+
+  try {
+    await warehouseSql.unsafe(
+      `CREATE TABLE ${quoteIdentifier(primaryDataset)} AS TABLE ${quoteIdentifier(input.sourceTable)}`,
+    );
+    await configSql`
+      INSERT INTO views (
+        view_name,
+        view_type,
+        primary_dataset,
+        secondary_dataset,
+        view_config
+      )
+      VALUES (
+        ${primaryDataset},
+        ${input.viewType},
+        ${primaryDataset},
+        ${input.secondaryDataset ?? null},
+        ${JSON.stringify(input.viewConfig)}
+      )
+    `;
+
+    return {
+      primaryDataset,
+      secondaryDataset: input.secondaryDataset ?? null,
+      viewType: input.viewType,
+    };
+  } catch (error) {
+    await configSql`
+      DELETE FROM views
+      WHERE primary_dataset = ${primaryDataset}
+    `;
+    await warehouseSql.unsafe(
+      `DROP TABLE IF EXISTS ${quoteIdentifier(primaryDataset)}`,
+    );
+    throw error;
+  } finally {
+    await Promise.all([
+      configSql.end({ timeout: 5 }),
+      warehouseSql.end({ timeout: 5 }),
+    ]);
+  }
+};
+
+/**
+ * Deletes one test-owned view row and its copied warehouse table.
+ *
+ * @param view - Identity returned by createApiTestView.
+ * @returns {Promise<void>}
+ */
+export const deleteApiTestView = async (view: ApiTestView): Promise<void> => {
+  const configSql = createTestDatabaseClient(TEST_CONFIG_DATABASE);
+  const warehouseSql = createTestDatabaseClient(TEST_WAREHOUSE_DATABASE);
+
+  try {
+    await configSql`
+      DELETE FROM views
+      WHERE primary_dataset = ${view.primaryDataset}
+        AND view_type = ${view.viewType}
+    `;
+    await warehouseSql.unsafe(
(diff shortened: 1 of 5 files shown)


tests/e2e/helpers/apiTestData.ts
```diff
@@ -0,0 +1,94 @@
...
+        view_config
+      )
+      VALUES (
+        ${primaryDataset},
+        ${input.viewType},
+        ${primaryDataset},
+        ${input.secondaryDataset ?? null},
+        ${JSON.stringify(input.viewConfig)}
+      )
+    `;
```
IamJeffG: I agree DQL is fine here (i.e. outside of migrations), but would it be possible to use the drizzle ORM methods to do it (on `viewConfig` table object), as opposed to raw SQL?

Raw SQL isn't inherently bad, but you lose type checking and it's likely to go out-of-sync the next time we modify this table's schema.

tests/e2e/helpers/apiTestData.ts
```diff
@@ -0,0 +1,94 @@
...
+  input: ApiTestViewInput,
+): Promise<ApiTestView> => {
+  const primaryDataset = `api_test_${randomUUID().replaceAll("-", "").slice(0, 12)}`;
+  const configSql = createTestDatabaseClient(TEST_CONFIG_DATABASE);
+  const warehouseSql = createTestDatabaseClient(TEST_WAREHOUSE_DATABASE);
+
+  try {
+    await warehouseSql.unsafe(
+      `CREATE TABLE ${quoteIdentifier(primaryDataset)} AS TABLE ${quoteIdentifier(input.sourceTable)}`,
+    );
```
IamJeffG, line 34: FYI @Atmosfearful  --- this project touches tables that are not managed by this project's Drizzle ORM. Those are the "warehouse" tables, they live in a different logical database.

(At least I think that is what this code is doing here)

And since Drizzle doesn't manage them, it's a reasonable exception that we have to do DDL in the test.

tests/e2e/helpers/apiTestData.ts
IamJeffG, line 1: Putting these database setups into its own global-ish helper is premature abstraction, IMO.
`12-view-dataset-api.spec.ts` is our only API test for now.  My expectation as we add more is that each test spec file INSERT its own data rows that are specifically needed in that test.  I get that it's a bit fuzzier since these are just helper functions, so maybe they would be extracted for sharing with future test files. But would future test files also be testing these same `viewConfig` at all?

Food for thought. I don't have a very confident take on this one.

tests/e2e/helpers/apiTestData.ts
IamJeffG: One thing I don't love is this is just cloning a table that was defined in tests/db-seed/warehouse.sql.  My hope is to get tests FAR more independent from that file, so eventually that file goes away.  i.e. get entirely rid of global test data.

The motivating use case is the same: if a future test needs a specific row shape, it seems to me they currently have to add it in the global `warehouse.sql`, and this helper here hasn't actually bought us the data independence we need.

Thoughts?

tests/e2e/helpers/apiTestData.ts
conservationtimothy: There are a few PRs stacked on this one and they then re-use the same helper https://github.com/ConservationMetrics/gc-explorer/blob/618993878d9c7ffd7e24f5b5b0c3c5d638bb0ed9/tests/e2e/14-views-rest-api.spec.ts , https://github.com/ConservationMetrics/gc-explorer/blob/15d573d91b5c4ef0b665ed7b84475c3b0bec05f6/tests/e2e/13-public-views-api.spec.ts

tests/e2e/helpers/apiTestData.ts
conservationtimothy: Yeah, I agree with your sentiment; will update.

tests/e2e/helpers/apiTestData.ts
conservationtimothy: Given this some thought.

You are right that this does not make the test data independent.  `warehouse.sql` is still here because the browser e2e tests load those tables (`/gallery/seed_survey_data` and the rest). That is why this PR did not delete it. But that does not mean the API tests need to keep cloning data from it.

I opened a follow-up [issue](https://github.com/ConservationMetrics/gc-explorer/issues/642) for the end state: API and browser e2e tests create the warehouse tables and rows they need, then Docker stops loading `warehouse.sql`.

I also plan to update the README to reflect the new mental model once implemented.

tests/e2e/helpers/apiTestData.ts
conservationtimothy: Updated https://github.com/ConservationMetrics/gc-explorer/pull/636/commits/5b013127a5e0f87d50fc86caedba58eef16f83f4

Atmosfearful (approved):
I think it's passable, I will have stronger opinions about testing layout when I'm more familiar with the codebase.

Issue #552, opened 2026-07-27
Refactor view configuration CRUD into a RESTful `/api/views` API


conservationtimothy:
## Feature Request

View configuration CRUD currently uses controller-style routes under `/api/config`:

- `GET /api/config` lists all view rows and available warehouse tables.
- `GET /api/config/:table` lists the views for a `primary_dataset`.
- `POST /api/config/new_table/:table?view_type=...` creates a view.
- `POST /api/config/update_config/:table?view_type=...` updates a view.
- `POST /api/config/delete_table/:table?view_type=...` deletes a view.

These routes describe operations and datasets rather than the `views` resources they manage. Each view already has a stable `view_id` primary key, while mutations currently identify views indirectly using `(primary_dataset, view_type)`.

The terminology is also misleading: `delete_table` deletes a row from the Guardian Connector `views` table, not the underlying warehouse table.

Introduce resource-oriented view configuration endpoints, for example:

- `GET /api/views`
- `GET /api/views/:view_id`
- `POST /api/views`
- `PATCH /api/views/:view_id`
- `DELETE /api/views/:view_id`

> NOTE: This issue concerns view configuration CRUD. Migrating the dataset data routes currently exposed under `/api/:table/*` should be handled separately.

### User Stories & Acceptance Criteria

- API consumers can create, retrieve, update, and delete views through noun-based `/api/views` routes using appropriate HTTP methods.
- Operations on an existing view identify exactly one row by `view_id`.
- Updating or deleting one view never changes sibling views sharing its `primary_dataset`.
- Mutations retain the current admin authorization requirements.
- Existing UI views use the new endpoints.
- Existing controller-style mutation routes are removed or explicitly deprecated with a documented compatibility period.
- Access to the available warehouse tables currently returned by `GET /api/config` is preserved through a clearly named endpoint or response contract.
- `public_views` synchronization remains correct when sibling views exist.
- Tests cover CRUD behavior, authorization, missing views, and datasets with multiple view types.

### UI Sketch (or Screenshot)

Not applicable. This is an API refactor with no intended UI behavior change.

## Implementation Plan
(may be added later, before work begins)

PR #637, opened 2026-08-27
Make public access view-specific

conservationtimothy:
## Goal

> Part 1 of 3 for [#552: Refactor view configuration CRUD into a RESTful /api/views API](https://github.com/ConservationMetrics/gc-explorer/issues/552).

Make public access view-specific so views of the same dataset can have independent permissions.

## Screenshots

Not applicable 

## What I changed and why

- Changed `public_views` from dataset-name identity to `view_id`.
- Added an in-place migration that preserves the table and backfills public views from existing configuration.
- Added a cascading foreign key from `public_views.view_id` to `views.view_id`.
- Updated public-view synchronization to operate on individual views.
- Updated middleware to match both dataset and view type.
- Updated database fixtures for the new schema.
- Added Playwright coverage for sibling-view permission isolation.

A dataset can expose multiple view types with different permissions. Using only the dataset name could accidentally grant public access to a private sibling view.

ompose stack with the migration.
- Ran dataset API, public-view synchronization, and visibility Playwright tests against Docker.
- Verified public status changes affect only the selected view.

## What I'm not doing here

- Adding RESTful `/api/views` CRUD endpoints.
- Migrating configuration pages to the new API.
- Removing legacy `/api/config` controllers.

## LLM use disclosure

GPT Sol 5.6 helped

### DIFF
--- a/middleware/oauth.global.ts
+++ b/middleware/oauth.global.ts
@@ -1,5 +1,10 @@
 import { defineNuxtRouteMiddleware, useRuntimeConfig } from "#imports";
-import type { User, RouteLevelPermission } from "@/types";
+import type {
+  PublicViewRow,
+  RouteLevelPermission,
+  User,
+  ViewType,
+} from "@/types";
 import { Role } from "@/types";
 import { decodeDatasetNameFromUrl } from "@/utils/identifierUtils";
 
@@ -45,11 +50,19 @@ export default defineNuxtRouteMiddleware(async (to) => {
           ? to.params.tablename
           : to.path.split("/").pop()!;
       const tableName = decodeDatasetNameFromUrl(rawTableName);
-      const publicTableNames = await $fetch<string[]>(
+      const viewType = to.path.split("/")[1] as ViewType;
+      const publicViews = await $fetch<PublicViewRow[]>(
         "/api/config/public_views",
       );
       // Public access: no login needed
-      if (publicTableNames.includes(tableName)) return;
+      if (
+        publicViews.some(
+          (view) =>
+            view.primaryDataset === tableName && view.viewType === viewType,
+        )
+      ) {
+        return;
+      }
 
       // Require authentication
       if (!loggedIn.value) {
@@ -62,10 +75,12 @@ export default defineNuxtRouteMiddleware(async (to) => {
         views: Array<{
           primaryDataset: string;
           viewConfig: { ROUTE_LEVEL_PERMISSION?: RouteLevelPermission };
+          viewType: ViewType;
         }>;
       }>("/api/config");
       const viewEntry = response.views?.find(
-        (view) => view.primaryDataset === tableName,
+        (view) =>
+          view.primaryDataset === tableName && view.viewType === viewType,
       );
       const permission: RouteLevelPermission =
         viewEntry?.viewConfig?.ROUTE_LEVEL_PERMISSION ?? "member";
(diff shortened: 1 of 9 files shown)


tests/e2e/13-public-views-api.spec.ts
```diff
@@ -0,0 +1,98 @@
+import { test, expect } from "@/tests/e2e/fixtures/auth-storage";
+import {
+  createApiTestView,
+  deleteApiTestView,
+} from "@/tests/e2e/helpers/apiTestData";
+import type { PublicViewRow } from "@/types";
+
+test("public view synchronization isolates sibling view permissions", async ({
```
Atmosfearful: Since we don't have rigid / formal spec documents, I would use these e2e tests as our source of truth, and be more granular with our tests in this file.

In this case rather than a single overarching test, I try to define one for each acceptance criteria. Examples of more granular and test names might be:
```js
// ROUTE_LEVEL_PERMISSION: "member"
"MEMBER permission may be viewed by a team member"
"MEMBER permission may not be viewed by a member of a different team"
"MEMBER permission may not be viewed by a user with no team"
```

Good e2e tests would make it easy for newcomers (like me, and future LLMs) to understand the intention at the product level. It's the first place I look when doing a PR review :D

Atmosfearful (changes requested):

tests/e2e/13-public-views-api.spec.ts
conservationtimothy: Makes loads of sense!!

tests/e2e/13-public-views-api.spec.ts
```diff
@@ -1,47 +1,99 @@
+import type { APIRequestContext } from "@playwright/test";
+
 import { test, expect } from "@/tests/e2e/fixtures/auth-storage";
 import {
   createApiTestView,
   deleteApiTestView,
 } from "@/tests/e2e/helpers/apiTestData";
-import type { PublicViewRow } from "@/types";
+import type { ApiTestView, PublicViewRow } from "@/types";
```
conservationtimothy, line 8: @Atmosfearful , I hope this is much better? 🙂

tests/e2e/13-public-views-api.spec.ts
conservationtimothy: Done https://github.com/ConservationMetrics/gc-explorer/pull/637/commits/df3897cb08dfb2189e68c0528ba0d3e1f74ac7fa

tests/e2e/13-public-views-api.spec.ts
Atmosfearful: Yes the behavior is clearer to me now thanks!

middleware/oauth.global.ts
```diff
@@ -45,11 +50,19 @@ export default defineNuxtRouteMiddleware(async (to) => {
           ? to.params.tablename
           : to.path.split("/").pop()!;
       const tableName = decodeDatasetNameFromUrl(rawTableName);
-      const publicTableNames = await $fetch<string[]>(
+      const viewType = to.path.split("/")[1] as ViewType;
+      const publicViews = await $fetch<PublicViewRow[]>(
         "/api/config/public_views",
```
Atmosfearful, line 55: Not feedback, but more of a note-to-self to come back to and understand better;

- this check here for "exists in set of all public views" threw up a red flag to me, especially in middleware that is flagged as `global`.
- I would have expected getters and db queries handle the authz to protect specific resources like `view`, while middleware is reserved for route-level protection and handling redirects and such

Atmosfearful (commented):
This is taking me too long to digest, I think we are better off setting up a 1-on-1 to pair review. Then I can ask all the dumb questions about things that aren't even related to this PR, and will help me understand the changes.

PR #644, opened 2026-09-01
Migrate to pnpm 11 with committed allowBuilds

### LINKED ISSUE #629: Migrate to pnpm 11 with committed allowBuilds
## Feature Request

Move Explorer from pinned pnpm 10.34.5 to pnpm 11 in a controlled change. Commit an explicit `allowBuilds` policy in `pnpm-workspace.yaml` so CI allows only reviewed dependency build scripts.

pnpm 11 requires Node.js 22+. This repo still uses Node 20 in Actions and Docker.
 
* As a maintainer, I run `pnpm approve-builds` locally, review each package, and commit the resulting `allowBuilds` map in `pnpm-workspace.yaml`.
* As CI, I read that committed policy and allow only those packages to run install/build scripts. I do not prompt. I do not approve everything.
* Acceptance:
  * `packageManager` / CI / Docker use the same pnpm 11.x.
  * Node is 22+ wherever pnpm 11 runs.
  * `allowBuilds` lists each previously ignored package as `true` or `false` after review (`esbuild`, `sharp`, `@parcel/watcher`, `unrs-resolver` likely `true`; `vue-demi` likely `false` unless we prove we need its postinstall).
  * `pnpm install --frozen-lockfile` succeeds in CI with no `ERR_PNPM_IGNORED_BUILDS`.
  * We do not set `dangerouslyAllowAllBuilds: true`.

### UI Sketch (or Screenshot)

N/A (CI and package-manager policy)

## Implementation Plan

1. Bump Node to 22+ in Actions, Docker, and `engines` as needed.
2. Install pnpm 11 locally and run `pnpx codemod run pnpm-v10-to-v11` if config needs moving.
3. Run `pnpm approve-builds` (interactive). Review each package. Do not use `--all` unless every remaining entry is already reviewed.
4. Commit `allowBuilds` in `pnpm-workspace.yaml`.
5. Pin CI and Docker to that pnpm 11.x.
6. Confirm `pnpm install --frozen-lockfile` and existing lint/unit/e2e jobs.

conservationtimothy:
## Goal

Move GitHub Actions and Docker installs to pnpm 11 with a committed allowBuilds policy, so ignored dependency scripts do not fail CI. Closes #629 

## Screenshots

N/A

## What I changed and why

pnpm 11 treats unreviewed dependency build scripts as an error and requires Node 22. I pin the three GitHub Actions jobs that install dependencies on the runner to Node 22 and `pnpm@11.25.0`. I make the same Node and pnpm pins in both Dockerfile stages. I ran a pnpm 11 frozen install, then set `allowBuilds` in `pnpm-workspace.yaml` for the packages that install actually blocked. `@parcel/watcher`, `esbuild`, `sharp`, and `unrs-resolver` are allowed because they need native or platform binaries. `vue-demi` is denied because this app is Vue 3 and that script only switches Vue 2 versus Vue 3 entry points.

I also set `packageManager` to `pnpm@11.25.0` so local Corepack matches CI. The lockfile stayed valid under `--frozen-lockfile`.

Testing locally

## What I'm not doing here

Nothing extrenous

## LLM use disclosure

None

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -4,8 +4,8 @@ packages:
 allowBuilds:
   "@parcel/watcher": true
   esbuild: true
-  sharp: true
-  unrs-resolver: true
-  vue-demi: false
+  sharp: true # transitive dependency of turf
+  unrs-resolver: true # transitive dependency of @nuxt/eslint
+  vue-demi: false # transitive dependency of reka-ui
 
 minimumReleaseAge: 10080 # minutes == 7 days
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -2,7 +2,7 @@ packages:
   - .
 
 allowBuilds:
-  '@parcel/watcher': true
+  "@parcel/watcher": true
   esbuild: true
   sharp: true
   unrs-resolver: true
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,4 +1,11 @@
 packages:
   - .
 
+allowBuilds:
+  '@parcel/watcher': true
+  esbuild: true
+  sharp: true
+  unrs-resolver: true
+  vue-demi: false
+
 minimumReleaseAge: 10080 # minutes == 7 days
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -20,10 +20,10 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: "20"
+          node-version: "22"
       - name: Install dependencies
         run: |
-          npm install -g pnpm@10.34.5
+          npm install -g pnpm@11.25.0
           pnpm install
       - name: Run linting and formatting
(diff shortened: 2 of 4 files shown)


pnpm-workspace.yaml
```diff
@@ -1,4 +1,11 @@
 packages:
   - .
 
+allowBuilds:
+  "@parcel/watcher": true
+  esbuild: true
+  sharp: true
+  unrs-resolver: true
+  vue-demi: false
```
rudokemper: I had to track down why we pull in `vue-demi` to begin with.

It is a dependency of `floating-ui`, itself a dependency of `reka-ui`.

I wouldn't mind it if we commented this in for each dep like

```suggestion
allowBuilds:
  "@parcel/watcher": true
  esbuild: true
  sharp: true # transitive dependency of turf
  unrs-resolver: true # transitive dependency of @nuxt/eslint
  vue-demi: false # transitive dependency of reka-ui
```

That way we know easily to remove here, if we uninstall those deps.

rudokemper (approved):

PR #647, opened 2026-09-07
Migrate view-wide Mapbox tokens to per-basemap configuration

### LINKED ISSUE #545: Migrate view-wide Mapbox tokens to per-basemap configuration
## Feature Request

Move `MAPBOX_ACCESS_TOKEN` from the top level of each view’s `view_config` into the corresponding objects stored in `MAPBOX_BASEMAPS`.

Staging currently contains:

- 19 views with valid `view_config` JSON
- 16 views with a non-empty global Mapbox token
- 7 views with `MAPBOX_BASEMAPS`, containing 9 basemaps
- 0 basemaps with an existing per-basemap token
- 9 token-bearing views without `MAPBOX_BASEMAPS`, apparently using legacy `MAPBOX_STYLE`

Other partner VMs may contain a similar mix, but should be inventoried before deployment.

- As an administrator, I can configure a separate Mapbox access token for each background map.
- Existing basemaps receive the current view-wide token during migration.
- Views using legacy `MAPBOX_STYLE` are converted to a one-item `MAPBOX_BASEMAPS` configuration or otherwise migrated without losing their token.
- `MAPBOX_ACCESS_TOKEN` is removed only after the affected view is successfully converted.
- Unexpected or malformed affected configurations prevent the migration from completing rather than being silently skipped.
- Views continue to load their configured basemaps after deployment.
- The migration is tested with:
  - Multiple basemaps
  - Legacy `MAPBOX_STYLE`
  - Existing per-basemap tokens
  - Empty basemap arrays
  - Malformed nested basemap JSON

### UI Sketch (or Screenshot)

Not applicable. This primarily changes configuration storage and migration behavior.

## Implementation Plan

1. Update the basemap configuration type and UI to use `access_token` per basemap.
2. Add a Drizzle SQL data migration that:
   - Parses `view_config`
   - Copies the global token into each missing basemap token
   - Converts legacy `MAPBOX_STYLE` configurations where required
   - Preserves existing per-basemap tokens
   - Removes the global token only after successful conversion
3. Update server map and alerts configuration to consume the selected basemap token.
4. Add migration, unit, and end-to-end coverage.
5. Spot-check migrated views after deployment.

conservationtimothy:
## Goal

Move Mapbox access tokens from view-wide configuration into each individual basemap configuration. This allows every Mapbox basemap to use its own style and token when selected. Closes #545 

> [!NOTE]
> Most of the additions are tests covering parsing, runtime switching, configuration, persistence, and API behavior. Generated Drizzle migration metadata also accounts for a significant portion of the diff. The work is split into three focused commits so each layer can be reviewed independently. For example 353 LOC addition for `server/database/migrations/meta/0015_snapshot.json`

## Screenshots

CleanShot 2026-09-07 at 21 19 51@2x
CleanShot 2026-09-07 at 21 14 43@2x
CleanShot 2026-09-07 at 21 13 33@2x

### 1. Runtime model and API behavior

[`17aee96` — Model Mapbox tokens per basemap](https://github.com/ConservationMetrics/gc-explorer/commit/17aee96)

- Added `access_token` to each basemap.
- Made `MAPBOX_BASEMAPS` the runtime source of Mapbox styles and tokens.
- Updated map, alerts, and gallery APIs to use the default basemap token.
- Updated map and alerts views to switch `mapboxgl.accessToken` when users select another basemap.
- Added parser, API, map, and alerts tests.

This keeps each style and its required token together and removes reliance on the old view-wide token.

### 2. Configuration UI

[`51e25f0` — Configure tokens on each Mapbox basemap](https://github.com/ConservationMetrics/gc-explorer/commit/51e25f0)

- Added an access-token input to every basemap card.
- Added validation requiring every configured basemap to have a token.
- Removed the old standalone Mapbox token and style fields.
- Updated documentation and translations.
- Added component and E2E coverage for saving distinct tokens on multiple basemaps.

This makes the configuration UI match the new per-basemap runtime model.

### 3. Data migration

[`adc6cd2` — Migrate Mapbox tokens into basemap config](https://github.com/ConservationMetrics/gc-explorer/commit/adc6cd2)

- Added a Drizzle migration that copies existing view-wide tokens into basemap objects.
- Converts legacy `MAPBOX_STYLE` configurations into a default basemap.
- Preserves existing per-basemap tokens.
- Rejects malformed configurations with the affected `view_id`.
- Removes the obsolete top-level token and style keys.
- Updated test seed data.
- Added concise migration comments explaining its purpose and failure behavior.

This safely moves existing deployments to the new configuration structure.

- Manual verification in the browser.
- The E2E test saving two basemaps with distinct access tokens passed.
- The migration ran successfully when starting the stack.

## What I'm not doing here

- Supporting the deprecated top-level `MAPBOX_ACCESS_TOKEN` and `MAPBOX_STYLE` at runtime after migration.
- Refactoring unrelated configuration or map code.

## LLM use disclosure

Cursor Agent (GPT Sol 5.6) was used.


components/config/ConfigMap.vue
```diff
@@ -379,6 +356,29 @@ const fullWidthKeys = [
...
+                      :for-id="`${tableName}-basemap-access-token-${index}`"
+                      required
+                      class="mb-1"
+                    >
+                      {{ $t("mapboxAccessToken") }}
+                    </ConfigFieldLabel>
+                    <input
+                      :id="`${tableName}-basemap-access-token-${index}`"
+                      class="w-full px-4 py-2 bg-violet-100 border border-violet-200 rounded-lg focus:ring-2 focus:ring-violet-500 focus:border-violet-500 transition-colors"
+                      placeholder="pk.ey…"
```
rudokemper, line 370: Missing the pattern

```ts
            pattern="^pk\.ey.*"
```

components/config/ConfigCard.vue
```diff
@@ -230,9 +230,28 @@ const areColumnsLoading = computed(
...
+): boolean => {
+  if (!encodedBasemaps) return false;
+  try {
+    const basemaps = JSON.parse(encodedBasemaps) as BasemapConfig[];
+    if (!Array.isArray(basemaps) || basemaps.length === 0) return false;
+    return basemaps.every((basemap) => hasConfigValue(basemap.access_token));
+  } catch {
+    return false;
+  }
+};
```
rudokemper, line 250: If the validation is "at least one basemap exists and every token is set", then I'd rename this `hasValidBasemapConfig`.

rudokemper (approved):
Great. Nice to see the fallback handling (entirely vestigial now) go away.

Please do make the two changes requested before merging.

components/config/ConfigCard.vue
conservationtimothy: Done

components/config/ConfigMap.vue
conservationtimothy: Done

PR #648, opened 2026-09-08
[Config] Add an interactive map to help set Mapbox parameters (zoom, coords, pitch, bearing)

### LINKED ISSUE #620: [Config] Add an interactive map to help set Mapbox parameters (zoom, coords, pitch, bearing)
## Feature Request

The interactive map and the map settings parameters should stay synchronized so that either can be used to configure the map.

When the user interacts directly with the map, the corresponding settings should update automatically. For example:

* Panning the map updates **Map center latitude/longitude**
* Zooming updates **Zoom level**
* Tilting updates **Pitch degree**
* Rotating updates **Bearing degree**

The inverse should also work: changing any of those settings should immediately update the map. For example, entering a new latitude/longitude recenters the map, and changing the zoom slider changes the visible zoom level.

**Expected behavior:** the map should effectively act as a visual input for these parameters, while the form fields remain a precise/manual input for the same state. Changes made through either interface are reflected in the other, without requiring a page refresh or an explicit “Apply” button.

Terrastories does this too, c.f. https://github.com/Terrastories/terrastories/blob/master/rails/app/views/dashboard/themes/edit.html.erb:

Image

The map can use the first map style in the Background map list:

Image

The map can appear under the Background Map list and above the map params.

If no valid background map combo (style and access token) is set, then it should not render on the page.

Could we reuse (or adapt) `shared/Minimap.vue` component for this?

conservationtimothy:
## Goal

  Let users configure map center, zoom, pitch, and bearing through either an interactive map or the existing settings fields. Closes #620 

  ## Screenshots
CleanShot 2026-09-08 at 14 57 47

CleanShot 2026-09-08 at 19 54 49@2x

  ## What I changed and why

- Added an interactive map preview below Background Maps and above the map parameters in the config form
- Map movements update the form fields, and field edits update the map immediately
- Preview uses the first background style and access token, stays hidden until the background loads, and hides on load failure without falling back to another background
- Added a dedicated `ConfigMapPreview` component rather than extending `shared/Minimap.vue`, keeping interactive camera behavior in the settings form separate from the static minimap

  I tested the configuration flow in both directions. Simulated map movements update all four controls, and manual field edits update the map camera. 

  ## What I'm not doing here

Nothing extrenous

  ## LLM use disclosure

  Codex helped


conservationtimothy:
@rudokemper , to answer your question: for a brand-new map or alert config with no coordinates, the preview starts at `latitude 0`, `longitude 0`, `zoom 0`, `pitch 0°`, and `bearing 0°` once the first background style and access token are valid.
 
If settings already exist or were copied from another config, the preview uses those values instead.

rudokemper:
Okay great. 

I think that's acceptable for this first iteration.

I agree with the ethos of @IamJeffG's question that it could be good to do something with framing the user around where their data is located. But that can be a follow up.

rudokemper (commented):
Thanks @conservationtimothy 

Two things I would like for you to implement before I do code review:

(1) Add a section header below the background map options, and above the map, with the  title "Mapbox Settings" (in the style of "Mapbox Background Map(s)")

Then also a label (in the style of "Provide up to three background map options from Mapbox...") -- "Set the [camera position](https://docs.mapbox.com/android/maps/guides/camera-and-animation/camera/) of the map."

(2) Add a pin to the middle of the map to show what the exact centroid is.

conservationtimothy:
> Thanks @conservationtimothy
> 
> Two things I would like for you to implement before I do code review:
> 
> (1) Add a section header below the background map options, and above the map, with the title "Mapbox Settings" (in the style of "Mapbox Background Map(s)")
> 
> Then also a label (in the style of "Provide up to three background map options from Mapbox...") -- "Set the [camera position](https://docs.mapbox.com/android/maps/guides/camera-and-animation/camera/) of the map."
> 
> (2) Add a pin to the middle of the map to show what the exact centroid is.

Done 
CleanShot 2026-09-08 at 19 54 49@2x

rudokemper:
Oh oops. I meant to also add (sorry, multi-tasking like crazy today)

> Set the [camera position](https://docs.mapbox.com/android/maps/guides/camera-and-animation/camera/) of the map. You can either interact with the map to set these values, or fill in the fields manually.

conservationtimothy:
All good , [done](https://github.com/ConservationMetrics/gc-explorer/pull/648/commits/dc7c5ce516ea508436743979bc4db47634a7b8df)

rudokemper (approved):

rudokemper:
@conservationtimothy I am going to go ahead merge this so I can deploy this to prod instances, as we are overdue for pushing config updates to partner instances.

PR #651, opened 2026-09-09
Export ignores active date filter and exports all data points

### LINKED ISSUE #645: Export ignores active date filter and exports all data points
**Environment / Context:**
- **Feature:** KML Export / Map View  
- **Expected vs. Actual:** Export includes filtered-out points instead of matching the active viewport date range.

**Steps to Reproduce:**

1. Open the map/data view containing data points.
2. Apply a date range filter to narrow down the visible points on the map.
3. Click the **Export to KML** button.
4. Open the exported KML file.
    
**Expected Behavior:** The exported KML file should only contain the points currently visible in the active date range/viewport.

**Actual Behavior:** The exported KML file includes all dataset points, ignoring the applied date range filter.

**Suggested Solution / Improvement:**

- **Immediate Fix:** Ensure the export function respects the currently applied date range filters.
- **Alternative / UX Enhancement:** Add an export configuration prompt allowing the user to select their export scope (e.g., "Export visible points only" vs. "Export all points").

conservationtimothy:
## Goal

  Make dataset exports respect the selected date range. Closes #645 

  ## Screenshots
  
  Not a screenshot but I downloaded filtered by date (7 points) and exported KML and this was the output:

| Alert ID | Territory | Area (ha) | Grid | Basin ID | Detection Date | Confidence | Coordinates (approx. centroid lon, lat) |
|---|---|---|---|---|---|---|---|
| 2026016100161660431 | redacted | 0.66 | 6 | 6166043 | Jan 2026 | 1 | redacted |
| 20260111100161660491 | redacted | 1.48 | 11 | 6166049 | Jan 2026 | 1 | redacted |
| 20260112100161660411 | redacted | 1.22 | 12 | 6166041 | Jan 2026 | 1 | redacted |
| 20260119100161660482 | redacted | 3.58 | 19 | 6166048 | Jan 2026 | 1 | redacted |
| 20260119100161660481 | redacted | 0.71 | 19 | 6166048 | Jan 2026 | 1 | redacted |
| 20260111100161660492 | redacted | 1.58 | 11 | 6166049 | Jan 2026 | 1 | redacted |
| 2026016100161660432 | redacted | 0.93 | 6 | 6166043 | Jan 2026 | 1 | redacted |

## What I changed and why

- Alerts sidebar now includes selected months in the location download request, and the export endpoint filters rows by detection year and month before generating KML, CSV, or GeoJSON
- Fixed the shared map date filter to retain the full start timestamp, preventing UTC conversion followed by date truncation from including records before the selected start date

Manual verification really

## What I'm not doing here

Nothing extrenous 

 ## LLM use disclosure

Codex (Astra initially then Sol) helped

### DIFF
--- a/server/api/[table]/export.get.ts
+++ b/server/api/[table]/export.get.ts
@@ -7,6 +7,7 @@ import {
   filterGeoData,
   filterToSelectedValues,
   filterByDateRange,
+  filterAlertsByDateRange,
 } from "@/server/dataProcessing/dataFilters";
 import { hasValidCoordinates } from "@/utils/geoUtils";
 import { validatePermissions } from "@/utils/accessControls";
@@ -187,7 +188,13 @@ export default defineEventHandler(async (event: H3Event) => {
     const timestampColumn = tableConfig.TIMESTAMP_COLUMN;
     const minDate = (query.minDate as string)?.trim();
     const maxDate = (query.maxDate as string)?.trim();
-    if (timestampColumn && (minDate || maxDate)) {
+    if (viewType === "alerts" && (minDate || maxDate)) {
+      dataToExport = filterAlertsByDateRange(
+        dataToExport,
+        minDate || undefined,
+        maxDate || undefined,
+      );
+    } else if (timestampColumn && (minDate || maxDate)) {
       dataToExport = filterByDateRange(
         dataToExport,
         timestampColumn,
--- a/components/alerts/AlertsIntroPanel.vue
+++ b/components/alerts/AlertsIntroPanel.vue
@@ -160,6 +160,8 @@ const showDownloads = computed(
           <div class="flex justify-center [&>div]:!mt-0">
             <DownloadMapData
               :data-for-download="props.dataForAlertsIntroPanel"
+              :export-min-date="props.statsExportMinDate"
+              :export-max-date="props.statsExportMaxDate"
               variant="outline"
             />
           </div>
--- a/composables/useDateAndCategoryFilter.ts
+++ b/composables/useDateAndCategoryFilter.ts
@@ -79,9 +79,7 @@ export const useTimestampFilter = () => {
   const dateMax = ref("");
 
   const setDateRange = (payload: { start: Date | null; end: Date | null }) => {
-    dateMin.value = payload.start
-      ? payload.start.toISOString().slice(0, 10)
-      : "";
+    dateMin.value = payload.start ? payload.start.toISOString() : "";
     dateMax.value = payload.end ? payload.end.toISOString() : "";
   };
 
--- a/composables/useTableExportDownload.ts
+++ b/composables/useTableExportDownload.ts
@@ -47,7 +47,9 @@ export const buildTableExportQueryParams = (options: {
   }
   const isStatisticsExport = options.exportPath === "statistics-export";
   if (
-    (options.exportTimestampColumn || isStatisticsExport) &&
+    (options.exportTimestampColumn ||
+      isStatisticsExport ||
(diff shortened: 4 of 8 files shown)


conservationtimothy:
> "@nicopace indicates above that the problem is on Map View but I would check the export logic on Alerts Dashboard also." https://github.com/ConservationMetrics/gc-explorer/issues/645#issuecomment-5577943564

This only applies to Alerts. Maps does not have date filters; I checked the datasets on both staging and demo. It does have `Filter by: <column>`, though. I tested this with `/map/inaturalist-lake-accotink` by applying a filter and downloading the KML, and confirmed that the export respects the applied filter.

rudokemper:
There is also Filter by Timestamp. I assume that is not affected but would be good to test.

conservationtimothy:
> There is also Filter by Timestamp. I assume that is not affected but would be good to test.

Just did, added `_submission_time` as the timestamp column for `bcmform_responses`, filtered it and it works as expected.

rudokemper (commented):
This looks like it meets the requirements. Could @Atmosfearful do the code review?

server/api/[table]/export.get.ts
```diff
@@ -187,7 +188,13 @@ export default defineEventHandler(async (event: H3Event) => {
     const timestampColumn = tableConfig.TIMESTAMP_COLUMN;
     const minDate = (query.minDate as string)?.trim();
     const maxDate = (query.maxDate as string)?.trim();
-    if (timestampColumn && (minDate || maxDate)) {
+    if (viewType === "alerts" && (minDate || maxDate)) {
```
Comment (Atmosfearful) [inline server/api/[table]/export.get.ts:191]: I see, so we can make assumptions about the underlying data shape, based on the `viewType`...
Is this just because the behavior changes, as in we want to filter `year_detec` rather than `timestampColumn`?

Or are the underlying `DataEntry` objects actually shaped differently when the viewType is "alerts"?

server/api/[table]/export.get.ts
```diff
@@ -187,7 +188,13 @@ export default defineEventHandler(async (event: H3Event) => {
...
+      dataToExport = filterAlertsByDateRange(
+        dataToExport,
+        minDate || undefined,
+        maxDate || undefined,
+      );
+    } else if (timestampColumn && (minDate || maxDate)) {
       dataToExport = filterByDateRange(
         dataToExport,
         timestampColumn,
```
Comment (Atmosfearful) [inline server/api/[table]/export.get.ts:200]: I suppose this is the only concern I have, that we aren't able to leverage `timestampColumn` argument properly for "alerts" as we do for the other data types. If it's just a matter of timestamp normalization maybe it's worth trying.

I guess the challenge is that the alert date gets split into two columns... there isn't already a third column that joins `year_detec` and `month_detec` into `datestamp_detec` or something like that?

Atmosfearful (approved):
Nice, simple fix and understandable. Tentative approval with questions:

server/api/[table]/export.get.ts
Comment (conservationtimothy) [reply server/api/[table]/export.get.ts:200]: We cannot use `TIMESTAMP_COLUMN` for alerts (warehouse table) as it is only exposed and documented for map and gallery views, so an alerts view normally has no configured `TIMESTAMP_COLUMN`.

server/api/[table]/export.get.ts
Comment (conservationtimothy) [reply server/api/[table]/export.get.ts:191]: No, The export route always fetches raw warehouse rows.   However, a table used by an alerts view must satisfy a different data contract. The alerts endpoint explicitly requires `year_detec` and `month_detec`, and the
  `AlertsDashboard` date slider filters by the YYYYMM value derived from those columns.

PR #653, opened 2026-09-09
Stale column references block unrelated view configuration saves

### LINKED ISSUE #652: Stale column references block unrelated view configuration saves

Discovered while QA'ing [Export ignores active date filter and exports all data points
](https://github.com/ConservationMetrics/gc-explorer/pull/651) on `/config/bcmform_responses?view_type=map`.

Saving `TIMESTAMP_COLUMN = _submission_time` failed because the saved `view_config` JSON still pointed at `COLOR_COLUMN: "color"` and `ICON_COLUMN: "icon"`. Those warehouse columns do not exist on `bcmform_responses`. They do exist on some other datasets, such as CoMapeo observation tables.

The validator is doing what it was written to do: reject pointers to columns that are not on the table. The JSON data is from copy-from-another-view. Thus the save button stays disabled even when the field the user just changed is valid.

#### Steps to Reproduce

1. Open a map view whose `view_config` still has `COLOR_COLUMN` or `ICON_COLUMN` pointing at columns the warehouse table does not have.
2. Change another valid setting, such as `TIMESTAMP_COLUMN`.
3. Attempt to save.
4. Observe that Save stays disabled, or the API rejects the update because of the stale pointers.

## Expected Behavior

Already-saved pointers to columns that are no longer on the table should not block unrelated valid changes.

On save, those stale pointers should be omitted from the submitted configuration. Newly introduced invalid column selections should still be rejected. Views that legitimately use `color` / `icon` columns (for example CoMapeo) must keep those settings.

conservationtimothy:
## Goal
Allow users to save valid configuration changes when an existing `view_config` contains stale references to warehouse columns that no longer exist, while preserving `COLOR_COLUMN` and `ICON_COLUMN` for views where those columns are still valid. Closes #652.

## Screenshots

Not included 

## What I changed and why

On save, we now distinguish stale invalid pointers (already in the saved config) from newly introduced invalid selections. Omit only the stale ones from the submitted payload. Keep blocking new invalid selections.
This lets users repair old configurations while preserving strict validation for new changes.

- Manually tested columns I know this exists
- Tested creating a new config 

## What I'm not doing here

- Relaxing server-side validation for invalid API payloads. That would be wrong. 
- Anything extrenous

## LLM use disclosure

Cursor agent helped

### DIFF
--- a/tests/unit/components/ConfigCard.mapInit.test.ts
+++ b/tests/unit/components/ConfigCard.mapInit.test.ts
@@ -221,6 +221,26 @@ describe("ConfigCard map initialization", () => {
     expect(submission.config).not.toHaveProperty("ICON_COLUMN");
   });
 
+  it("keeps saved columns when no dataset columns are loaded", async () => {
+    const wrapper = mountConfigCard({
+      ...savedMapConfig,
+      COLOR_COLUMN: "color",
+      MEDIA_COLUMN: "photos",
+    });
+    const cardVm = wrapper.vm as unknown as { localConfig: ViewConfig };
+
+    cardVm.localConfig.MAPBOX_ZOOM = 12;
+    await nextTick();
+
+    await wrapper.get("form").trigger("submit");
+
+    const submission = wrapper.emitted("submitConfig")?.[0]?.[0] as {
+      config: ViewConfig;
+    };
+    expect(submission.config.COLOR_COLUMN).toBe("color");
+    expect(submission.config.MEDIA_COLUMN).toBe("photos");
+  });
+
   it("still blocks newly introduced unavailable columns", async () => {
     const wrapper = mountConfigCard(savedMapConfig, [
       {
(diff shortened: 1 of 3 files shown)


conservationtimothy:
Easiest way I can explain this is before we implemented copy config being strict for view type i.e. you can only copy an alerts view type if the `viewType` we are creating or editing is alerts, you could copy whatever and some data sources then contained columns e.g. `COLOR_COLUMN` that wasn't actually in the warehouse DB so I discovered our logic which now only filters available columns effectively meant the API was not being saved

tests/unit/components/ConfigCard.mapInit.test.ts
```diff
@@ -182,4 +186,76 @@ describe("ConfigCard map initialization", () => {
...
+    );
+    const cardVm = wrapper.vm as unknown as { localConfig: ViewConfig };
+
+    cardVm.localConfig.TIMESTAMP_COLUMN = "_submission_time";
+    await nextTick();
+
+    const submitButton = wrapper.get<HTMLButtonElement>(
+      '[data-testid="config-submit-button"]',
+    );
+    expect(submitButton.element.disabled).toBe(false);
```
Atmosfearful, line 212: nice, happy to see we are in the habit of adding coverage even for small fixes like this

Atmosfearful (approved):

tests/unit/components/ConfigCard.mapInit.test.ts
conservationtimothy: Thanks !!

PR #654, opened 2026-09-10
Alerts Dashboard - Fix map rendering for MultiPolygon

### LINKED ISSUE #649: [Alerts] MultiPolygon alerts not drawing on map on mount (but do when `resetToInitialState` called!)

IGP Programs discovered a bug with MultiPolygon alerts where they do **not** appear on the Alerts Dashboard view on first load - neither with or without query params (`?alertId=...`).

But, curiously, if the user presses "Reset Dashboard", which calls `resetToInitialState()`, they _do_ then appear.

Image

My guess is that we are not handling MultiPolygon painting correctly when the `AlertsDashboard.vue` component is mounted.

NB: The same might be true for `MapView.vue` component.

#### Steps to Reproduce

Check out https://github.com/ConservationMetrics/gc-programs/issues/228#issuecomment-5589474066

## Expected Behavior

MultiPolygons paint on the map the same as Polygons, no matter what state.

Atmosfearful:
## Goal
Fixes MultiPolygon alerts not appearing on the Alerts Dashboard when the page first loads or when opened with an alertId link.

Fixes #649 

## Screenshots
Below, the two red squares are a multipolygon (newly added to the seed data) rendering after page refresh, where before they did not render until the reset button was pressed.

Forgot to take screenshots of the "before" state.
image

## What I changed and why
- Add a MultiPolygon fake alert seed record and unit/E2E regression coverage.
- Update the filters to accept MultiPolygon or coalesce to `polygon`

red-green TDD and localhost browser testing

## What I'm not doing here
I avoided a larger refactor that would have supported multipolygon natively and created shared utils between the `reset` and the initial load logic.
Goal is just to fix the issue.

## LLM use disclosure
GPT 5.6 Terra via OpenCode TUI

# Future thinking

Created a refactoring ticket for shared util -> #655


conservationtimothy (approved):
Good fix.

rudokemper:
Just two quick cosmetic and logistical things:

1. Please put a more descriptive title on the PR before merging, as that will be the commit message after squash and merge.
2. Document the future thinking somewhere, either as a TODO comment in the code or in an issue. I find that these things are often quickly forgotten if they are buried on a Github PR / issue / comment.

PR #656, opened 2026-09-10
Alerts Dashboard - Refactor Map Layer Rendering to shared utils

### LINKED ISSUE #655: Map layer rendering refactor
Background: In #654 we implemented a targeted fix to address the bug in #649 
The bug only applied to the initial render logic, not the 'map reset' logic, even though both are doing similar things.

To avoid future drift like this in the future, I believe we can create a shared util and use it everywhere that currently duplicates the same logic;

- Initial source creation: [`components/AlertsDashboard.vue:620-1066`](https://github.com/ConservationMetrics/gc-explorer/blob/main/components/AlertsDashboard.vue#L620-L1066)
- Date-filter source updates: [`composables/useAlertsDateFilter.ts:135-210`](https://github.com/ConservationMetrics/gc-explorer/blob/main/composables/useAlertsDateFilter.ts#L135-L210)
- Reset source updates: [`components/AlertsDashboard.vue:1581-1651`](https://github.com/ConservationMetrics/gc-explorer/blob/main/components/AlertsDashboard.vue#L1581-L1651)
- Incident and feature-selection source/layer lists: [`composables/useIncidents.ts`](https://github.com/ConservationMetrics/gc-explorer/blob/main/composables/useIncidents.ts) and [`composables/useFeatureSelection.ts`](https://github.com/ConservationMetrics/gc-explorer/blob/main/composables/useFeatureSelection.ts)

To inform this ticket please look at the changes that were made in #654

Atmosfearful:
## Goal
Previously alert mapbox rendering logic was shared across multiple parts of the codebase, which exposed us to drift in behavior between them - for example the bug that was fixed in #654 

After this PR, the core of this logic is centralized to shared utils defined in alertMapLayers.ts

Closes #655 

There is probably still some opportunity to consolidate further but then I would be stepping into full re-write territory, which I want to avoid.

## Screenshots

The following screenshots were taken After the PR changes and compared to behavior seen on current `main` branch.

✅️ selecting a feature
after_polygon-selection
✅️ date filtering of features
after_polygons_datefilter
✅️ rendering multipolygons and polygons
after_polygons
✅️ reset button still works
after_reset2

## What I changed and why
Created a central matrix / list of the mapbox layers, and utils around those, then reused them in the 5 places where this logic was repeated;

- Initial rendering
- Date filtering
- Reset
- Selection
- Incident highlighting

This PR reduces the total LOC in AlertsDashboard.vue and composables files

I manually QA'd rendering, reset, and selection behavior for the available seed data sets.

## What I'm not doing here
Changing behavior for end-user

## LLM use disclosure
GPT-5.6 Terra for planning/editing and GPT-6.0 Astra for review

### DIFF
--- a/tests/unit/utils/alertMapLayers.test.ts
+++ b/tests/unit/utils/alertMapLayers.test.ts
@@ -1,0 +1,160 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  alertMapLayers,
+  getAlertGeometryRenderKind,
+  getAlertSourceFeatures,
+} from "@/utils/alertMapLayers";
+import type { Feature } from "geojson";
+
+const point: Feature = {
+  type: "Feature",
+  geometry: { type: "Point", coordinates: [1, 2] },
+  properties: { alertID: "point" },
+};
+
+const polygon: Feature = {
+  type: "Feature",
+  geometry: {
+    type: "Polygon",
+    coordinates: [
+      [
+        [0, 0],
+        [1, 0],
+        [1, 1],
+        [0, 0],
+      ],
+    ],
+  },
+  properties: { alertID: "polygon", geographicCentroid: "1,2" },
+};
+
+const multiPolygon: Feature = {
+  type: "Feature",
+  geometry: {
+    type: "MultiPolygon",
+    coordinates: [
+      [
+        [
+          [0, 0],
+          [1, 0],
+          [1, 1],
+          [0, 0],
+        ],
+      ],
+    ],
+  },
+  properties: { alertID: "multipolygon", geographicCentroid: "3,4" },
+};
+
+const lineString: Feature = {
+  type: "Feature",
+  geometry: {
+    type: "LineString",
+    coordinates: [
+      [0, 0],
+      [1, 1],
+    ],
+  },
+  properties: { alertID: "line", geographicCentroid: "5,6" },
+};
+
+describe("alertMapLayers", () => {
+  it("defines every stable alert source and layer ID once", () => {
+    expect(alertMapLayers).toEqual([
+      expect.objectContaining({
+        period: "mostRecent",
+        kind: "point",
+        sourceId: "most-recent-alerts-point",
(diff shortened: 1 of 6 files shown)


utils/alertMapLayers.ts
```diff
@@ -0,0 +1,147 @@
...
+    layerId: "previous-alerts-linestring",
+  },
+  {
+    period: "previous",
+    alertsKey: "previousAlerts",
+    kind: "centroids",
+    sourceId: "previous-alerts-centroids",
+    layerId: "previous-alerts-centroids",
+  },
+];
```
conservationtimothy, line 73: The new list adds most-recent (red) first, then previous (orange). Mapbox paints later layers on top, so if a previous alert overlaps a recent one, orange can cover red. The old code added previous first so red stayed on top.

You’ll see it where recent and previous polygons, lines, or centroids overlap. Check the style layer IDs on `/alerts/fake_alerts` and you can compare with [staging](https://explorer.cmistaging.guardianconnector.net/alerts/fake_alerts)

tests/unit/utils/alertMapLayers.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        kind: "linestring",
+        sourceId: "previous-alerts-linestring",
+        layerId: "previous-alerts-linestring",
+      }),
+      expect.objectContaining({
+        period: "previous",
+        kind: "centroids",
+        sourceId: "previous-alerts-centroids",
+        layerId: "previous-alerts-centroids",
+      }),
```
conservationtimothy, line 112: Nice, can the new test also cover:
- initial load putting a MultiPolygon in the polygon source 
- the same after a date filter 
- the same after reset 
- selection still hitting the right layer

conservationtimothy (approved):
Just a few requests before merge but otherwise LGTM!

rudokemper:
@Atmosfearful you may want to wrap this one up, as it could become stale with forthcoming code contributions that will touch `AlertsDashboard`?

PR #658, opened 2026-09-10
mapbox config save after preview recovery

conservationtimothy:
## Goal

Let a user save a map or alerts view after they correct Mapbox style or access token values and the settings preview shows the map again. Rudo and I discovered during a call. 

## What I changed and why

Save and preview were using different validation rules. Preview would render the map when the first background map had a valid style and public token, while save additionally rejected inline style objects, required a projection, and treated existing column references as invalid when no dataset columns were available to validate against. This meant a user could paste a valid token, successfully preview the map, and still be unable to save the configuration.

This change aligns the two paths by using shared validation helpers that accept either a Mapbox style URI or inline style object and a public `pk.ey` token.  Existing invalid column references are preserved and only removed on submit when a column list is available for validation.

Red - Green tests, manual verification

## What I'm not doing here

Nothing extrenous

## LLM use disclosure

Grok 4.6

### DIFF
--- a/components/config/ConfigCard.vue
+++ b/components/config/ConfigCard.vue
@@ -10,9 +10,11 @@ import {
 } from "@/types";
 import { VIEW_INFO_CONFIG_KEYS } from "@/composables/useCopyConfig";
 import { CONFIG_LIMITS } from "@/utils";
+import { isMapboxPublicToken, isUsableMapboxStyle } from "@/utils/mapGLHelpers";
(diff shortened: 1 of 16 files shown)


conservationtimothy:
I think @rudokemper is uniquely qualified to do review on this once since it is both small and we found it together 🙂

rudokemper (approved):

PR #687, opened 2026-09-17
UX solution for users interacting with both column and timestamp filters

### LINKED ISSUE #685: UX solution for users interacting with both column and timestamp filters
## Feature Request

We decided to only show one filter at a time; "filter data by column" or "filter by date".

But it should be possible to apply **both** simultaneously.

To support this, we should figure out a visual way to paint the buttons, cueing the user that a filter has been applied.

And/or, the buttons simply show/hide the popups, stacking them on top of each other if they are both toggled.

Ideas?

"I want only points about monkey sightings between the dates of Feb 1 2026 and March 1 2026"

## Implementation Plan

conservationtimothy:
## Goal

Show when a column filter, a date filter, or both filters are applied. Closes #685 

## Screenshots

CleanShot 2026-09-17 at 10 15 50@2x
CleanShot 2026-09-17 at 10 07 13@2x
CleanShot 2026-09-17 at 10 07 06@2x
CleanShot 2026-09-17 at 10 06 56@2x

## What I changed and why

- `MapFilterControls` now tracks `columnFilterActive` and `dateFilterActive` separately from popup open state
- A column filter is considered active when the selection has one or more values; a date filter is active when `TimestampFilter` emits `active: true`, meaning the month range is not the full span
- Each button uses `aria-expanded` for the open popup and `aria-pressed` for the applied filter state
- Active filter buttons show a white ring and an amber badge; both can show simultaneously so an AND of column and date filters is visible even when no popup is open
- Hidden date-slider tooltips use `display: none` to prevent them from drawing over the column panel

Manual testing

## What I'm not doing here

This PR does not change how date and column filters combine on the map as that AND already exists. 

## LLM use disclosure

Grok 4.6


rudokemper (approved):
Great solution. Thanks.

conservationtimothy:
Pleasure as always !

PR #689, opened 2026-09-17
[Map / Alerts] Provide access to a link with query params for map zoom and centroid

### LINKED ISSUE #643: [Map / Alerts] Provide access to a link with query params for map zoom and centroid
While working with a partner, I happened to share a url for a set of features we were looking at with him, and when he opened the link, the link was showing something different.

There is definitely a conflict between incident id and this, but would be good to think it through so this issue doesn't replicate.

conservationtimothy:
## Goal

Give Map and Alerts a shareable viewport. Users can send a link that opens a specific map center and zoom, so a view can show nearby context instead of only one selected feature. Closes #643 
> **Note:** This PR adds 843 lines. 412 of those lines are tests.

## Screenshots
CleanShot 2026-09-17 at 18 56 29@2x
CleanShot 2026-09-17 at 18 55 05@2x
CleanShot 2026-09-17 at 19 07 13@2x

## What I changed and why

- Map and Alerts now write `lat`, `lng`, and `zoom` to the URL when the map finishes a move, and restore that camera on load when those params are present
- Feature ids (`alertId`, `incidentId`, `secondaryDocId`, `mapeoDocId`) take priority over camera params — when a feature id is present, the map flies to that feature rather than restoring the saved camera
- Added a violet Copy control in the top-right map chrome that copies a location URL with the current camera and without any feature ids, without affecting the open feature or the live address bar
- Copy link to alert and copy link to incident still produce feature-only URLs and omit camera params

 In the running app I zoomed a Map view and saw the query update, then opened the same URL at zoom 10 and got the restored camera. On Alerts I opened a URL with both an alert id and a distant camera. The map flew to the alert. Copy link to alert produced a feature-only URL. Copy link to this location produced a camera-only URL. The live URL still held the alert id. The control uses the same violet as Reset map. Its chrome stacking stays at 1010, under the legend.

## What I'm not doing here

Copying a location does not deselect the open alert or incident. This change does not add bounding-box params.

## LLM use disclosure

Cursor Grok 4.6 helped


tests/e2e/16-map-camera-query.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  });
+
+  await expect(page).toHaveURL(/[?&]lat=-3\.12/);
+  await expect(page).toHaveURL(/[?&]lng=-60\.02/);
+  await expect(page).toHaveURL(/[?&]zoom=11\.5/);
+
+  const alertsPath = new URL(page.url()).pathname;
+  await page.goto(`${alertsPath}?lat=-3.12000&lng=-60.02000&zoom=11.50`);
+  await page.locator("#map[data-map-ready='true']").waitFor();
+
```
conservationtimothy, line 25: Straightforward but tests i hope would make @Atmosfearful proud 🙂

rudokemper:
@conservationtimothy for anyone not able to spin up this code in runtime, Can you share a screenshot of what the copy control looks like?

conservationtimothy:
CleanShot 2026-09-17 at 22 23 26@2x
CleanShot 2026-09-17 at 22 24 13@2x

nicopace (approved):
from a user perspective, this looks awesome!
thanks for implementing it!
Did not check the code though... will let others be the judge of that :)

components/shared/CopyMapLocationButton.vue
```diff
@@ -0,0 +1,77 @@
...
+        :class="{ 'text-green-300': showCopied }"
+        aria-hidden="true"
+      />
+    </button>
+    <span
+      class="copy-map-location-tooltip"
+      :class="{ 'is-visible': showCopied }"
+      role="tooltip"
+    >
+      {{ showCopied ? $t("copied") : $t("copyMapLocation") }}
```
Atmosfearful, line 34: do these appear on tap or focus as well? ( for mobile users )

Atmosfearful (approved):
Nice, very useful!

I should say, I have been bitten by this in the past, where users just wanted to share the link to the default view, but when they copy it from the browser bar it includes the url params to their current view.

The Default View is useful, especially for people who are bad at maps. If you click and are presented with a super zoomed in map, or one zoomed in on a single point, it's disorienting. The fly-to behavior from mapbox can help with this but not always.

So an alternate middle-ground solution is to:
- consume query params on page load, and strip them from the url immediately
- force users to use the "copy link to this location" button when they *intend* to share a link to a specific coordinate

And tangentially -> Another related feature would be, "copy link to this feature" which could be accessible from the feature details panel which opens when you click a map element.

rudokemper:
> * consume query params on page load, and strip them from the url immediately
> * force users to use the "copy link to this location" button when they _intend_ to share a link to a specific coordinate

I understand @Atmosfearful's point and I agree with this middle ground.

> And tangentially -> Another related feature would be, "copy link to this feature" which could be accessible from the feature details panel which opens when you click a map element.

This we have.

:thinking: 

...

_*actually checks*_

Wait, no we don't! Only for features on `AlertsDashboard` component :open_mouth: 

We definitely should have this on MapView, IMO!

conservationtimothy:
https://github.com/ConservationMetrics/gc-explorer/issues/691 created an issue !

components/shared/CopyMapLocationButton.vue
conservationtimothy: Hover is mouse-only. Keyboard uses focus-visible. Touch now shows the tooltip on focus (tap), then Copied! after a successful copy.

PR #692, opened 2026-09-21
Docker Image Bloat

### LINKED ISSUE #673: Docker image bloat
1GB is quite large for a webapp like this.
> Image

It wasn't always this way; looks like it ballooned on September 2 with build 1066:
[`20260831-1059`](https://hub.docker.com/layers/communityfirst/guardianconnector-explorer/20260831-1059/images/sha256-51ae4cf29a8c18ab2830f324ee3a086971c4e6cd19cf34028487eadb266cc4e2): Compressed size 164.81MB
[`20260902-1066`](https://hub.docker.com/layers/communityfirst/guardianconnector-explorer/20260902-1066/images/sha256-fdc5eb8c9d25aca005e96eb693a97272719198797e058466916da9374c27d14a): Compressed size 261.56MB

conservationtimothy:
## Goal

Reduce the size of the production Docker image. The image grew from about 600 MB to about 1 GB after the pnpm 11 upgrade. Closes #673 .

## Screenshots

CleanShot 2026-09-21 at 12 58 52@2x

App starts up !

## What I changed and why

The production image was installing pnpm and running `pnpm install --prod`, but the running application does not use those production dependencies. Nitro builds a standalone Node server into `.output`, and the container starts it directly with `node .output/server/index.mjs`.

I removed the unnecessary production-stage pnpm install and now copy only the files required at runtime: the Nitro `.output` directory, the start script, and the SQL migration files used by the migration plugin.

This also prevents pnpm’s registry metadata cache under `/root/.cache/pnpm` from being included in the final image, reducing unnecessary image size. The build stage is unchanged.

I compared published images `20260831-1059` (before pnpm 11), `20260902-1066` (pnpm 11), and `20260918-1166` (latest). `docker history` showed the `pnpm install --prod` layer growing from 358 MB to 756 MB, while the application output remained about 14 MB. The pnpm 11 image also contained a new ~385 MB cache under `/root/.cache/pnpm`, along with unused `/app/node_modules` and the global pnpm installation.

I then built the changed Dockerfile locally. The resulting image was 261 MB uncompressed and 83 MB compressed. The container still exited when database environment variables were missing. With test database environment variables set, the server reached `Listening on http://[::]:8080. 

This confirmed that the application can load and start from `.output` without `/app/node_modules`.

## What I'm not doing here

I do not change the build stage.  I do not change the start script or the migration file path.

## LLM use disclosure

None


Atmosfearful (approved):
Nice cleanup, agree we don't need pnpm for prod images

IamJeffG:
what's the new image size?

Issue #657, opened 2026-09-10
Flaky e2e test runs due to auth setup


Atmosfearful:
## Current Buggy or Suspicious Behavior
As shown in these logs here, a test run failed in CI. Re-running the same suite led to a pass on the second try.

https://github.com/ConservationMetrics/gc-explorer/actions/runs/34506289302/job/102970313144

Evidence from logs;
- It seems the tests were running while the user was not logged in. This is observed by downloading the error-context.md files from the playwright output from that run above
- There is code in `tests/e2e/auth.setup.ts` designed to wait for auth to complete before proceeding with the other tests, but this apparently had a false positive or naively assumes success -- either way it should not have proceeded.

## Expected Behavior
Authentication setup should wait longer or fail clearly, rather than pass invalid state to dependent tests.

## Proposed fix 
From GPT-6 Astra
> I would address this specific race first, rather than adding retries.
> 1. Make navigation checks origin-aware. A return to the application must match Explorer’s configured origin, not merely a path containing /login. Handle the Auth0 consent page separately.
> 2. Wait for actual application authentication. After returning, poll /api/_auth/session with the browser context’s cookies until it reports the expected account and role. Use a bounded timeout instead of fixed sleeps.
> 3. Save only verified state. Log success and write the state file only after that check passes. Verify the saved state in a fresh request context before releasing dependent tests.
> 4. Fail clearly when authentication does not complete. Report the final origin/path and whether an application user was present. Do not swallow the timeout. Avoid logging URL query strings, tokens, or cookie values.

PR #693, opened 2026-09-21
Regression on map rendering

### LINKED ISSUE #690: Regression on map rendering
A map used to work before, now it doesn't.
Details in https://github.com/ConservationMetrics/gc-programs/issues/238

The data is fetched from database properly.

conservationtimothy:
## Goal

Fix the polygon map rendering regression introduced by #654. Closes #690 

## What I changed and why

I removed `MultiPolygon` from the Mapbox `$type` filter. Mapbox uses `Polygon` for both Polygon and MultiPolygon features. The unsupported value caused Mapbox to reject the complete polygon layer.

I added fake Polygon and MultiPolygon warehouse rows and a database-backed browser test. The test creates an isolated map view, uses the normal map API and page routes, and removes the view after the test.

Manual

## What I'm not doing here

Nothing extrenous 

## LLM use disclosure

None


rudokemper (approved):

Atmosfearful:
@conservationtimothy Seems the newly added e2e test is flaky, it failed again on merge to main;

62a5f34b8de5531e96a8f4c320606a93a78e4643

I also don't like all the tsignore comments, and the fact that we change runtime prod code (assigning that window variable just to support the e2e tests)

But I still have a few bits I don't fully understand (why I didn't encounter this in my own QA), so I'll study that a bit more and open a cleanup PR if needed

conservationtimothy:
> @conservationtimothy Seems the newly added e2e test is flaky, it failed again on merge to main;
> 
> [62a5f34](https://github.com/ConservationMetrics/gc-explorer/commit/62a5f34b8de5531e96a8f4c320606a93a78e4643)
> 
> I also don't like all the tsignore comments, and the fact that we change runtime prod code (assigning that window variable just to support the e2e tests)
> 
> But I still have a few bits I don't fully understand (why I didn't encounter this in my own QA), so I'll study that a bit more and open a cleanup PR if needed

Yeah, correlation isn’t causation -- I’m not convinced the test itself is flaky just yet. I’ve noticed a pattern where, when two PRs are merged to `main` in quick succession, the second run fails.

I opened a throwaway PR to run the E2E suite again in isolation and everything is green: https://github.com/ConservationMetrics/gc-explorer/pull/694

Definitely worth digging into though, especially given the discrepancy with your own QA.

Atmosfearful:
yeah true I saw `mapbox-gl` in the error output and made a false assumption. It's not your test that is failing; The root cause of the ci failure is likely the one described in this ticket that is already open:

https://github.com/ConservationMetrics/gc-explorer/issues/657

Also I have a better understanding now of how the regression slipped through my QA, as always it's obvious in hindsight-- all my test coverage and manual QA was via Alerts, not MapView, which follows a different rendering pipeline.

Lack of real e2e mapbox rendering is definitely the root cause so I am glad you added that in this PR.

Just a few more questions on my end that im looking into at the moment-- at very least I think I will put up a PR to use a type overload rather than ts-expect-error

